Issue #29, opened 2018-07-17
Errors when running example


Daiver:
Hello! I just found that example from https://albumentations.readthedocs.io/en/latest/examples.html sometime gives me `TypeError` and OpenCV related errors

Here is minimal example
```
import albumentations
from albumentations import (
    HorizontalFlip, IAAPerspective, ShiftScaleRotate, CLAHE, RandomRotate90,
    Transpose, ShiftScaleRotate, Blur, OpticalDistortion, GridDistortion, HueSaturationValue,
    IAAAdditiveGaussianNoise, GaussNoise, MotionBlur, MedianBlur, IAAPiecewiseAffine,
    IAASharpen, IAAEmboss, RandomContrast, RandomBrightness, Flip, OneOf, Compose
)
import numpy as np

def strong_aug(p=.5):
    return Compose([
        RandomRotate90(),
        Flip(),
        Transpose(),
        OneOf([
            IAAAdditiveGaussianNoise(),
            GaussNoise(),
        ], p=0.2),
        OneOf([
            MotionBlur(p=.2),
            MedianBlur(blur_limit=3, p=.1),
            Blur(blur_limit=3, p=.1),
        ], p=0.2),
        ShiftScaleRotate(shift_limit=0.0625, scale_limit=0.2, rotate_limit=45, p=.2),
        OneOf([
            OpticalDistortion(p=0.3),
            GridDistortion(p=.1),
            IAAPiecewiseAffine(p=0.3),
        ], p=0.2),
        OneOf([
            CLAHE(clip_limit=2),
            IAASharpen(),
            IAAEmboss(),
            RandomContrast(),
            RandomBrightness(),
        ], p=0.3),
        HueSaturationValue(p=0.3),
    ], p=p)

print(albumentations.__version__)

for i in range(1000):
    print(i)
    image = np.ones((300, 300))
    mask = np.ones((300, 300))
    whatever_data = "my name"
    augmentation = strong_aug(p=0.9)
    data = {"image": image, "mask": mask, "whatever_data": whatever_data, "additional": "hello"}
    augmented = augmentation(**data)
    image, mask, whatever_data, additional = augmented["image"], augmented["mask"], augmented["whatever_data"], augmented["additional"]

```

First full error message
```
Traceback (most recent call last):
  File "a.py", line 49, in <module>
    augmented = augmentation(**data)
  File "/home/daiver/.local/lib/python3.6/site-packages/albumentations/core/composition.py", line 17, in __call__
[...]

PR #30, opened 2018-07-17
Use uint8 inputs in the example

creafz:
Use uint8 inputs in the example, because for now some augmentations seem to have problem with float inputs - [Issue #29]


ternaus:
:+1:

PR #38, opened 2018-07-24
supporting float32 input for jpeg augmentation

arsenyinfo:


setup.py
```diff
@@ -34,7 +34,7 @@ def get_long_description():
     license='MIT',
     url='https://github.com/albu/albumentations',
     packages=find_packages(exclude=['tests']),
-    install_requires=['numpy>=1.11.1', 'scipy', 'opencv-python', 'imgaug'],
+    install_requires=['numpy>=1.11.1', 'scipy', 'opencv-python', 'imgaug>=0.2.5'],
```
creafz: Could you please clarify why this change is made? Are there any interaction problems with albumentations and imgaug < 0.2.5?

arsenyinfo: Existing tests fail with imgaug < 0.2.5 due to `AttributeError: module 'imgaug.augmenters' has no attribute 'PerspectiveTransform'` (it was added in 0.2.5 only).

creafz:
Could you please also update the docstring for `JpegCompression`  and add `float32` as a supported image type? https://github.com/albu/albumentations/blob/master/albumentations/augmentations/transforms.py#L433

arsenyinfo:
@creafz done! I'm not sure if I should add a warning about worse performance on float32 into the docstring as the `Image types` may be used for some tests/CI purposes, and free speech may break it - please advice.

creafz:
For now, we don't use the `Image types` section for any tests.

I think the best way to add a warning it to use a `Note` section header:

```
	...
        Image types:
            uint8, float32

        Note:
            Warning message.
```

It will be displayed under the `Image types` section in the docs:

PR #39, opened 2018-07-25
 Do not modify the input image in Cutout and RGBShift

creafz:
Also I have updated default parameters for Cutout.


ternaus:
:+1:

PR #53, opened 2018-08-27
Feature/iaa apply to bboxes

BloodAxe:
This PR adds transformation of bboxes when using IAA-based augmentations.


albumentations/imgaug/transforms.py
```diff
@@ -22,13 +25,45 @@ def apply(self, img, **params):
 
 
 class DualIAATransform(DualTransform, BasicIAATransform):
-    pass
+    def __init__(self, p):
+        super(DualIAATransform, self).__init__(p)
+
+    def apply_to_bboxes(self, bboxes, **params):
+        rows, cols = params['rows'], params['cols']
```
creafz: I think we could use `apply_to_bboxes(self, bboxes, rows, cols, **params)` here, since we know that those variables (`rows` and `cols`) should be passed to `apply_to_bboxes`.

albumentations/imgaug/transforms.py
```diff
@@ -22,13 +25,45 @@ def apply(self, img, **params):
...
+class IAACropAndPad(DualIAATransform):
+    def __init__(self, px=None, percent=None, pad_mode='constant', pad_cval=0, keep_size=True, p=1):
+        super(IAACropAndPad, self).__init__(p)
+        self.processor = iaa.CropAndPad(px, percent, pad_mode, pad_cval, keep_size)
+
+
+class IAAFliplr(DualIAATransform):
+    def __init__(self, p=0.5):
+        super(IAAFliplr, self).__init__(p)
+        self.processor = iaa.Fliplr(1)
```
creafz: Shouldn't it be `iaa.Fliplr(p)` because we want to pass the probability of applying `iaa.Fliplr` and this probability could differ from 1?

albumentations/imgaug/transforms.py
```diff
@@ -22,13 +25,45 @@ def apply(self, img, **params):
...
+
+
+class IAAFlipud(DualIAATransform):
+    def __init__(self, p=0.5):
+        super(IAAFlipud, self).__init__(p)
+        self.processor = iaa.Flipud(1)
```
creafz: And also here, shouldn't it be `iaa.Flipud(p)`?

albumentations/imgaug/transforms.py
```diff
@@ -116,9 +151,26 @@ class IAAPiecewiseAffine(DualIAATransform):
         image, mask
     """
 
-    def __init__(self, scale=(0.03, 0.05), nb_rows=4, nb_cols=4, p=.5):
+    def __init__(self, scale=0, nb_rows=4, nb_cols=4, order=1, cval=0, mode="constant", p=.5):
```
creafz: I think it would be better to use single quotes for `mode='constant'` because we mainly use single quotes in the other parts of the project.

albumentations/imgaug/transforms.py
```diff
@@ -116,9 +151,26 @@ class IAAPiecewiseAffine(DualIAATransform):
...
+
+    Args:
+        p (float): probability of applying the transform. Default: 0.5.
+
+    Targets:
+        image, mask
+    """
+
+    def __init__(self, scale=1.0, translate_percent=None, translate_px=None, rotate=0.0, shear=0.0, order=1, cval=0,
+                 mode="reflect", p=0.5):
```
creafz: And single quotes here for `mode='reflect'`

albumentations/imgaug/transforms.py
BloodAxe: Totally makes sense.

albumentations/imgaug/transforms.py
BloodAxe: Probably you're right, that we should pass parameter `p` to `iaa.Fliplr`, and initialize `super(IAAFliplr, self).__init__(1)` to unconditional forward calls to imgaug transformations.

PR #56, opened 2018-09-04
Added bbox transform for RandomRotate90 and Transpose

BloodAxe:


ternaus:
:+1:

albumentations/augmentations/functional.py
```diff
@@ -552,6 +552,46 @@ def bbox_random_crop(bbox, crop_height, crop_width, h_start, w_start, rows, cols
...
 
+def bbox_rot90(bbox, factor, rows, cols):
+    """Rotates a bounding box by 90 degrees CCW (see np.rot90)
+
+    :param bbox:
+    :param n: Number of rotations [0;3]
+    :param rows:
+    :param cols:
+    :return:
+    """
```
creafz: We are using Google Style docstrings (here is an example of this style - http://www.sphinx-doc.org/en/stable/ext/example_google.html)

So the dosctring should look something like this:
```
"""Rotates a bounding box by 90 degrees CCW (see np.rot90)

Args:
    n (int): Number of rotations [0;3]
"""
```
This is not exactly what Google Style recommends (because this docstring doesn't describe all parameters), but I think that it provides better readability.

albumentations/augmentations/functional.py
```diff
@@ -552,6 +552,46 @@ def bbox_random_crop(bbox, crop_height, crop_width, h_start, w_start, rows, cols
...
+
+def bbox_transpose(bbox, axis, rows, cols):
+    """Transposes a bouning box along given axis.
+
+    :param bbox:
+    :param axis: 0 - main axis, 1 - secondary axis
+    :param rows:
+    :param cols:
+    :return:
+    """
```
creafz: And also the docstring here.

albumentations/augmentations/transforms.py
```diff
@@ -242,6 +245,9 @@ def get_params(self):
         # Random int in range [0, 4)
         return {'factor': np.random.randint(0, 4)}
 
+    def apply_to_bbox(self, bbox, factor=0, **params):
+        return F.bbox_rot90(bbox, factor, **params)
+
```
creafz: Could you please also update docstrings for `RandomRotate90` and `Transpose` and add `bboxes` to `Targets:` (after `image, mask`)?

PR #58, opened 2018-09-05
Use python random generator instead of np.random

creafz:
Reference to the PyTorch issue with np.random: https://github.com/pytorch/pytorch/issues/5059


ternaus:
:+1:

PR #69, opened 2018-09-17
Enhance bbox pipeline

albu:
add compose for bounding boxes for simpler usage.
all boxes not cropped until the end of pipeline
add filter by area and fraction


albumentations/augmentations/bbox_utils.py
```diff
@@ -147,4 +159,60 @@ def convert_bboxes_from_albumentations(shape, bboxes, target_format):
...
+        rows (int): Image rows.
+        rows (int): Image cols.
+    """
+    resulting_boxes = []
+    for bbox in bboxes:
+        if min_visibility:
+            transformed_box_area = calculate_bbox_area(bbox, rows, cols)
+            bbox[:4] = np.clip(bbox[:4], 0, 1.)
+            clipped_box_area = calculate_bbox_area(bbox, rows, cols)
+            if clipped_box_area / transformed_box_area < min_visibility:
```
creafz: Not sure, but maybe we need to check that `transformed_box_area` is not 0 to avoid ZeroDivisionError.

PR #82, opened 2018-09-24
add smallest_max_size and remove redundant arg/function name

erikgaas:


ternaus:
Can you add some tests?

PR #84, opened 2018-09-25
auto codestyle check

Scitator:
Hi! 
Let's make albumentations pep8 again!

How to run:
```
pip install yapf
bash yapf.sh --all
```

For now, there are a lot of todos :)
[out.txt](https://github.com/albu/albumentations/files/2415085/out.txt)

Hope this PR will help.

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,0 +1,4 @@
+[style]
+based_on_style=pep8
+allow_split_before_dict_value=False
+join_multiple_lines=False
--- a/yapf.sh
+++ b/yapf.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+
+# Cause the script to exit if a single command fails
+set -eo pipefail
+
+# this stops git rev-parse from failing if we run this from the .git directory
+builtin cd "$(dirname "${BASH_SOURCE:-$0}")"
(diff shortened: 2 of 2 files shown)


ternaus:
Do you have an example of how this extra style check will be beneficial with respect to the existing checks?

creafz:
Hello. Thank you for the pull request, but we already have a PEP8 style checker, that tests each commit ([we are using Travis-CI for it](https://github.com/albu/albumentations/blob/master/.travis.yml#L26)), also you can run the linter locally by executing `flake8` command.

For now `flake8` doesn't report any PEP8 violations, if you have found any, could you please post them, so we can check the linter?

Note that we deliberately choose to set the maximum line length to 119 symbols as a reasonable compromise, and not 79 as recommended in PEP8.

creafz:
As for the output in the `out.txt` file, it seems that `yapf` was run with the default settings for line length (79 symbols) and also it looks like it reformats some valid and PEP8 compatible code to another format, that was chosen by `yapf` authors.

Scitator:
Yeap, it can reformat code for appropriate codestyle defined by `.style.yapf`.
I will check its output with 120 right margin rule.

PR #101, opened 2018-10-17
Replace cv2.flip(hflip+vflip) to numpy flip (performance reason)

DBusAI:
numpy img[::-1,:,...] is faster then cv2.flip(arr, 0) 
and same for img[:,::-1,...] with cv2.flip(arr, 1)


ternaus:
Just curious, do you have numbers on the performance comparison?

PR #121, opened 2018-11-05
Added support for multiple masks

ZFTurbo:
Fix needed for mask-rcnn implementation


albu:
I'm going to improve custom targets in the nearest future, but this PR is quite ok for now

PR #131, opened 2018-11-23
add targets_as_params

albu:
Sometimes you want params being dependent on targets.
Example usecase:
Crop images randomly around single box in image. You need to pass this box to params to generate random positions for crop.


albu:
WARNING: IT BREAKS API

PR #133, opened 2018-11-26
add additional targets to Compose

albu:


Benz-Tracxpoint:
Do I need to pass a list of dictionaries to apply the augmentation to an entire folder of images? Or is there some generator-like way to do it?

Thanks in advance,
Felix

albu:
Hi, no, basically you need to apply transform to images one by one. 
If you use pytorch - you can just use custom Dataset and pass albumentations as transforms. DataLoader will do job generating batches for you.

This PR is mainly for tasks like image-to-image or other multi-image or multi-mask tasks, where you have to apply the same transform for bunch of images.

Venka97:
Can someone provide an example of how to use this with multi-image single mask task?  If I have a dictionary `sample = {'image1': _img1, 'image2': _img2, 'label': _target}` can I pass this to Compose directly?

albu:
right now there is a limitation that at least one image should be named `image`.
you can pass `{'image': _img1, 'image2': _img2, 'mask': _target}` and you also have to pass `{'image2': 'image'}` to `additional_targets`

PR #149, opened 2018-12-20
Added conda build configs and upload script

### LINKED ISSUE #91: Availability in Anaconda
Hello all,

thanks for working on such a nice project, I have tried imgaug, augmentor and keras, but wasn't satisfied since I am doing a lot of patch segmentation.

I would have wanted to know if it were possible to add albumentations in the repository of Anaconda, so that we could install it and keep track of the versions easily?

Thanks in advance,
Nicolas

vfdev-5:
Fixes #91

## TL;DR
To build package and upload to anaconda, simply run (on linux machine with conda installed)
```
export ANACONDA_TOKEN=<Albu's Token>
sh conda.recipe/build_upload.sh
```

## Details

1) How to generate ANACONDA_TOKEN: https://docs.anaconda.com/anaconda-cloud/user-guide/tasks/work-with-accounts#creating-access-tokens

2) When everything is done, you should have the same as here:
   
https://anaconda.org/vfdev-5/albumentations

3) Built packages: linux, windows, osx vs py27, py35, py36, py37

4) Imgaug has no conda builds thus we need to install it with `pip` 

I tested on conda environment on python 3.6 linux only


conda.recipe/conda_build_config.yaml
```diff
@@ -0,0 +1,6 @@
+python:
+  - 2.7
+  - 3.5
+  - 3.6
+  - 3.7
+
```
creafz: Looks like an extraneous empty line, could you please remove it?

vfdev-5: Okay, done.

PR #159, opened 2019-01-13
RandomSizedBBoxSafeCrop

daisukelab:
Hi,
This is updated pull request from #153.
It had unintentionally added extra function which caused error, now it was fixed.
RandomSizedBBoxSafeCrop is useful for keeping bounding boxes.
Thanks.


daisukelab:
Hi, thanks for your feedback and merge.
I see it would be better to have `bbox_safe` option.
I will try to implement, and request for merge again once it is ready.

PR #165, opened 2019-01-18
WebP compression as an another transformation

tomkee:
WebP is a modern image format that provides superior lossless and lossy compression for images on the web. Using WebP, webmasters and web developers can create smaller, richer images that make the web faster.
WebP compression is made in similar manner as JPEG compression

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -381,6 +381,29 @@ def jpeg_compression(img, quality):
 
 
 @preserve_shape
+def webp_compression(img, quality):
+    input_dtype = img.dtype
+    needs_float = False
+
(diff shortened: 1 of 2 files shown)


albu:
I think it's better to add compression type as argument so new class and method names would be like `compress_image`. Also make sure you have deprecation warning for jpeg compression, you can see it here https://github.com/albu/albumentations/blob/master/albumentations/augmentations/transforms.py#L925

tomkee:
Ok. Should compression type be string or enum? Currently `imencode` supports a few formats:
https://docs.opencv.org/3.0-beta/modules/imgcodecs/doc/reading_and_writing_images.html#imwrite

albu:
enum works fine

PR #190, opened 2019-02-14
Added 1 channel support to CLAHE

ternaus:
Following https://github.com/albu/albumentations/issues/180

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -1,8 +1,8 @@
 from __future__ import division
 
 import math
-from functools import wraps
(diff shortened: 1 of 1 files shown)


ternaus:
> Not sure about this because all other transforms do not support grayscale so it would work only in specific case of one transform.

I think all spatial transformations work well with grayscale. And probably some of the color ones.

we just added one more to the list.

I think for grayscale it is better to have CLAHE + spatial, rather than only spatial.

subijaheemesh:
python code for clahe on  gray scale imges?
pls share

Issue #194, opened 2019-02-16
Lack of ability to keep output size fit when using IAAAffine


lamhoangtung:
Hi, when using IAAAffine, there is no way to keep the image after affine transformation to completely contained in the output. A lot of people use this frequently so it is very annoying

Affine Transform in `imgaug` supported this use case, we just need to pass `fit_output=True` when init it, but we can't do that via albumentations.

So I made a small Pull Request to resolve this problem. Can you guys consider to merge it ?

Thanks ❤️

ternaus:
Hello,

Could you please share what is the use case, when you need both the original and transformed image?

lamhoangtung:
Hi @ternaus, 

The problem is after affine transformation parts of the image may be outside of the image plane or the image might make up only a small part of the image plane. Keeping output image fit can be useful e.g. for rotations by 45 degrees to avoid that the image corners are outside of the image plane.

To do that just need to use `fit_output=True` argument when we call `imgaug`, my PR addressed that ;)

albu:
it's `False` by default, let's keep default but passing this argument is good option.

lamhoangtung:
Yeah, my PR make it passable 😂

Vào Th 3, 19 thg 2, 2019 lúc 16:46 Alexander Buslaev <
notifications@github.com> đã viết:

ternaus:
We removed imgaug dependencies. but it has  `fit_ouput` parameter as well

PR #195, opened 2019-02-16
Add `fit_output` parameter for IAAAffine

lamhoangtung:
Hi, when using IAAAffine, there is no way to keep the image after affine transformation to completely contained in the output. A lot of people use this frequently so it is very annoying

Affine Transform in `imgaug` supported this use case, we just need to pass `fit_output=True` when init it, but we can't do that via albumentations.

So I made a small Pull Request to resolve this problem (#194 issue). Can you guys consider to merge it ?

Thanks ❤️

### DIFF
--- a/albumentations/imgaug/transforms.py
+++ b/albumentations/imgaug/transforms.py
@@ -235,7 +235,7 @@ class IAAAffine(DualIAATransform):
     """
 
     def __init__(self, scale=1.0, translate_percent=None, translate_px=None, rotate=0.0, shear=0.0, order=1, cval=0,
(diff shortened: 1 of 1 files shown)


albu:
looks like tests does not pass because we have too old imgaug @creafz

lamhoangtung:
I don’t think newer version of imgaug can broke anything 💪

Vào Th 3, 19 thg 2, 2019 lúc 16:58 Alexander Buslaev <
notifications@github.com> đã viết:

Dipet:
too old

PR #200, opened 2019-02-22
Fix bug in keypoint scale in random sized crop

BloodAxe:


ternaus:
Could you please add a test for this case?

PR #208, opened 2019-03-02
Fix in torchvision2albumentations notebook

ternaus:
Added img_to_tensor operation to make albumentations DataLoader directly used in PyTorch.

### DIFF
--- a/notebooks/migrating_from_torchvision_to_albumentations.ipynb
+++ b/notebooks/migrating_from_torchvision_to_albumentations.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


notebooks/migrating_from_torchvision_to_albumentations.ipynb
```diff
@@ -102,7 +103,7 @@
     "        if self.transform:\n",
     "            augmented = self.transform(image=image)\n",
     "            image = augmented['image']\n",
-    "        return image, label\n",
+    "        return img_to_tensor(image), label\n",
```
albu:
```suggestion
    "        return image, label\n",
```

PR #211, opened 2019-03-02
Faster hflip with OpenCV

vfdev-5:
Hi, 
A simple replacement with OpenCV seems to give a ~3x acceration:
- Numpy: '1.16.1'
- OpenCV: 4.0.0

```
x.shape = (281, 500, 3)
```
1) current implementation
```python
%%timeit -r 7 -n 10
for _ in range(1000):
    hflip(x)
> 681 ms ± 3.62 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
2) with opencv
```python
%%timeit -r 7 -n 10
for _ in range(1000):
    hflip_fast(x)
> 197 ms ± 77.8 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
and compare with `vflip`
```python
%%timeit -r 7 -n 10
for _ in range(1000):
    vflip(x)
> 30.3 ms ± 108 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Both `hflip*` implementations give the same result and `C_CONTIGUOUS` arrays:
```python
(hflip(x) == hflip_fast(x)).all()
> True
r1 = hflip(x)
r1.flags['C_CONTIGUOUS']
> True
r2 = hflip_fast(x)
r2.flags['C_CONTIGUOUS']
> True
```

What do you think about such modification ?

**EDIT:**
on `x.shape=(281, 500, 1)` this implementation is slower than the original one:
```python
%%timeit -r 7 -n 10
for _ in range(1000):
    hflip(x)
> 48.5 ms ± 2.64 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit -r 7 -n 10
for _ in range(1000):
    hflip_fast(x)
> 62.6 ms ± 177 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit -r 7 -n 10
for _ in range(1000):
    vflip(x)
> 12.6 ms ± 112 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
For other number of channels, e.g. 2, 3, 5, ... opencv seems to be faster but scales in time, however numpy is not.

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -66,6 +66,10 @@ def hflip(img):
     return np.ascontiguousarray(img[:, ::-1, ...])
 
 
+def hflip_cv2(img):
+    return cv2.flip(img, 1)
+
+
 @preserve_shape
 def random_flip(img, code):
     return cv2.flip(img, code)
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -149,7 +149,9 @@ class HorizontalFlip(DualTransform):
     """
 
     def apply(self, img, **params):
-        if img.ndim == 3 and img.shape[2] > 1:
+        if img.ndim == 3 and img.shape[2] > 1 and img.dtype == np.uint8:
+            # Opencv is faster than numpy only in case of
+            # non-gray scale 8bits images
             return F.hflip_cv2(img)
         else:
             return F.hflip(img)
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -149,7 +149,10 @@ class HorizontalFlip(DualTransform):
     """
 
     def apply(self, img, **params):
-        return F.hflip(img)
+        if img.ndim == 3 and img.shape[2] > 1:
+            return F.hflip_cv2(img)
+        else:
+            return F.hflip(img)
 
     def apply_to_bbox(self, bbox, **params):
         return F.bbox_hflip(bbox, **params)


creafz:
@vfdev-5 Could you also test a case with a np.float32 input? On my machine, OpenCV is slower than NumPy:
```
x = np.random.rand(281, 500, 3).astype(np.float32)
```

```
%%timeit -r 7 -n 10
for _ in range(1000):
    hflip(x)
> 509 ms ± 1.64 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
```
%%timeit -r 7 -n 10
for _ in range(1000):
    hflip_cv2(x)
> 1.62 s ± 329 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

vfdev-5:
@creafz yes, I confirm that opencv is slower on f32 similar to your results (not x3 but ~1x). 
Do you have use-cases of images in float32 after reading from disk ? But you're right it make sense to check the performances on int16, int32 input data.

creafz:
We try to make all transformations compatible with a `np.float32` input, so they can work not only with a `np.uint8` input but with other data types as well. [Here is an example notebook](https://github.com/albu/albumentations/blob/master/notebooks/example_16_bit_tiff.ipynb) that shows how to work with 16-bit tiff images by converting them to float32 using `ToFloat` transformation (and since we can use `ToFloat` to convert any datatype to float32 we don't need to support all the datatypes in our transformations, we just need to support `np.uint8` and `np.float32` inputs).

ternaus:
@vfdev-5 If you will modify PR so that the implementation of the transform `numpy` vs `cv2` was chosen based on the type of the data it would be really cool.

HorizontalFlip is the most widely used transform and the fact that we are lacking in the performance is not good. Hence, your PR is highly appreciated.

vfdev-5:
@ternaus I did not have time to write the message to ask whether it is something like that you would like to have and you already merge it :)

Anyway it would be cool to see the benchmark on HorizontalFlip with this add

Issue #215, opened 2019-03-03
Error in returning invisible keypoints


cdicle:
## 🐛 Bug

One can return invisible keypoints by setting `remove_invisible` flag to `False` as follows
```
augmentor = albu.Compose([...], keypoint_params={'format': 'xy', 'remove_invisible': False})
```
However, due to this [check](https://github.com/albu/albumentations/blob/master/albumentations/core/composition.py#L216), those invisible keypoints result in an error. Note that, `check_validity` flag which supposed to be conditioned on `remove_invisible` flag is always `True`.

Will submit a fix following up this bug report.
Best.

BloodAxe:
Merged. Thanks for the PR!

PR #216, opened 2019-03-03
Condition keypoint validity check on input param

cdicle:
This PR solves the issue #215 by conditioning keypoint validity check on `remove_invisible` flag.


albu:
cool, it would be perfect if you can add test for this error
thank you for cooperation 👍

PR #222, opened 2019-03-03
add kernel installation description

n01z3:
Добавил 3 строчки о том, как поставить в кернел кэггла


albu:
already fixed in https://github.com/albu/albumentations/pull/221 which was merged previously. thanks for cooperation

PR #228, opened 2019-03-06
opencv-python-headless should be enough, and easier to deploy

ocourtin:
Hi,

It looks like nothing, but could really make packager life easier.
As opencv-python imply several extra libs to be installed on system (such as libSM).

Thanks in advance,

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ setup(
     license='MIT',
     url='https://github.com/albu/albumentations',
     packages=find_packages(exclude=['tests']),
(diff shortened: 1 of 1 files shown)


albu:
hi, thanks for the pr.
would it install opencv-python-headless in case if opencv-python is already installed?

ocourtin:
@albu
> would it install opencv-python-headless in case if opencv-python is already installed?

Yes indeed.

But the point is that you cannot only trust pip on this one, 
as opencv-python can be installed but not able to works as some lib dependencies missed...

PR #231, opened 2019-03-09
Added GaussianBlur transformation with tests

vfdev-5:
Hi, in this PR I added a gaussian blur similarly to ordinary blur and the median one. Can be useful for some other folks as me. What do you think ?


ternaus:
@creafz Do we have an instruction on how to update README when a new transformation is added?

PR #235, opened 2019-03-10
Adding transforms from https://github.com/albu/albumentations/issues/191

ternaus:
Added RandomSnow

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -670,6 +670,68 @@ def add_sun_flare(img, flare_center_x, flare_center_y, angle, num_circles, src_r
 
 
 @preserve_shape
+def add_shadow(img, shadow_roi, num_shadows, shadow_dimension):
+    """Adds shadows to the image.
+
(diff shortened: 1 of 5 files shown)


albumentations/augmentations/functional.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import division
 
 import math
+import random
```
albu: you cannot use random here, it leads to errors in case of additional image-type targets

BloodAxe:
Can you please update notebooks with showcase of new augmentations? We will need fancy pictures for a new release :)

BloodAxe:
As I understand these augmentations works only for 3-channel image (assuming images are RGB format). If a user will use single-channel or multispectral image it will crash in `cv2.cvtColor` with a not-so-user-friendly exception message (Let's put aside a question why someone may use non-rgb image with AddSnow augmentation. It basically will happen). I'd love to see some type checking of input image for correct number of channels and user-friendly message how to use it properly.

I really like how it's made in Angular framework. When you do something wrong, they print what was wrong and what is typical workaround.

PR #236, opened 2019-03-11
Added test to check multitarget images consistency

ternaus:
We need to verify that when we use multitarget and apply a transform to the equal two images, transformed images are also equal.


albu:
it was added in one of previous PRs that you approved https://github.com/albu/albumentations/pull/214/files#diff-c14a2f858bf526cb9ca600fb52c50342

Issue #263, opened 2019-05-20
Improve Cutout augmentation


kirillbobyrev:
Random Erasing ([Paper](https://arxiv.org/abs/1708.04896), [Code](https://github.com/zhunzhong07/Random-Erasing)) is an interesting augmentation which would be nice-to-have. Since Cutout already has similar functionality, I am proposing to extend it in several ways to allow more flexibility and improve user experience. There are several ways Cutout could be improved:

- [x] Allow cutting out a random number of regions of random size (sample `num_holes` and `[height, width]` from uniform random distribution).
- [x] Support more targets such as masks, bounding boxes and keypoints.
- [ ] Allow user to specify the patch modification: e.g. just zero out (like `Cutout` currently does), fill each channel with it's mean or take user-provided mean values as an argument (this is how Random Erasure is currently [done](https://github.com/zhunzhong07/Random-Erasing/blob/master/transforms.py#L46-L51)). I'm not sure whether it fits the API of Cutout or not, but it would be even better to support applying other augmentations to the region: e.g. use color jitter/blur/any other distortion.

ternaus:
`Allow cutting out a random number of regions of random size (sample num_holes and [height, width] from uniform random distribution).` is implemented => good enough. 

Let's merge it and work on the other modifications like extra targets and patch modifications later. (You may keep this feature request so that we would not forget.)

ogencoglu:
Can we fill the regions with random noise (pixel-level) instead of a value? From the documentation it seems like it just accepts a scalar value to fill in.

ternaus:
@ogencoglu you can have the desired transform using [Lambda](https://github.com/hongzhenwang/RRPN-revise)

ryanchesler:
Very interested in extending cutout to be applied to masks. Seems like it would be trivial to implement. Is there any reason for this not already being an option? I can spend some time implementing it if need be and send a pull request

ternaus:
[CoarseDropout](https://explore.albumentations.ai/transform/CoarseDropout) works on all targets, inluding masks, bounding boxes and keypoints

If fill_value = random, generate random noise to fill the hole

PR #264, opened 2019-05-20
Improve Cutout augmentation

kirillbobyrev:
* Allow cutting out random number of regions.
* Allow cutting out regions of random size.
* Slightly simplify the code and remove the code which always creates
  patches of even height and width.
* Add sanity checks for non-negative parameters.
* Fix documentation which mentions squares being zeroed out from the
  image while the applied patches can be rectangular.
* Change tests so that serialization is provided with new parameters.

Implements the first part of #263.

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -901,12 +901,18 @@ class Normalize(ImageOnlyTransform):
 
 
 class Cutout(ImageOnlyTransform):
-    """CoarseDropout of the square regions in the image.
+    """CoarseDropout of the rectangular regions in the image.
 
     Args:
-        num_holes (int): number of regions to zero out
-        max_h_size (int): maximum height of the hole
-        max_w_size (int): maximum width of the hole
+        max_holes (int): Maximum number of regions to zero out.
+        max_height (int): Maximum height of the hole.
+        min_width (int): Maximum width of the hole.
+        min_holes (int): Minimum number of regions to zero out. If `None`,
+            `min_holes` is be set to `max_holes`. Default: `None`.
+        min_height (int): Minimum height of the hole. Default: None. If `None`,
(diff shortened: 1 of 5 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -921,11 +927,20 @@ class Cutout(ImageOnlyTransform):
...
-        self.max_h_size = max_h_size
-        self.max_w_size = max_w_size
+        self.max_holes = max_holes
+        self.max_height = max_height
+        self.max_width = max_width
+        self.min_holes = min_holes if min_holes is not None else max_holes
+        self.min_height = min_height if min_height is not None else max_height
+        self.min_width = min_width if min_width is not None else max_width
+
+        assert 0 < self.min_holes and self.min_holes <= self.max_holes
```
ternaus: I would believe that 

`assert 0 < self.min_holes <= self.max_holes`

is a better syntax.

kirillbobyrev: Fair point. I guess the pain from using C++ where it would evaluate into `((0 < self.min_holes) <= self.max_holes)` is somehow hardwired in my brain so I decided to use this instead.

albu:
we have to preserve old api to not break pipelines with updates, let's make two versions and deprecation warning for previous (for 0.4.0)

kirillbobyrev:
@albu I'm not sure what's the best way to do that, but I've tried to address your comment in my last commit by introducing `CoarseDropout` and deprecating `Cutout` in favor of it. Is that what you suggested?

albu:
Yes, thanks. I think it's perfect option. Just make few style fixes as reported by travis and we can merge it

kirillbobyrev:
@albu Great, thanks! I've fixed CI issues and the patch should be ready for the merge now.

PR #267, opened 2019-05-27
added fill value to Cutout constructor

bfialkoff:
I had the need to use the cutout augmentation, but I wanted to be able to set the color. As it was, the cutout was always black. I thought it would be a useful feature to be able to set the fill_value.


kirillbobyrev:
Hi! I have recently proposed `Cutout` improvements in #263 and this looks like a subcase for the third improvement (patch modification). I've implemented `CoarseDropout` which deprecates `Cutout` in #264 and I think this patch should at least be applied to `CoarseDropout`. If you could find a way to generalize it and build a high-level interface for patch modifications that could be even better, though I don't have much experience with the project so I don't know what would be the best way to do that.

ternaus:
Does not pass the tests.

bfialkoff:
Yes, I know I made some changes that caused the fail and just haven't
gotten around to correcting it yet.

John-Yao:
hello, I had tried to set the value of Cutout's fill_value . It does work well. All the cut hole  is fill with black.

bfialkoff:
That's strange. Please sharead your code and I'll try and diagnose the
problem.

John-Yao:
import sys
import cv2
import albumentations as A
from mmcv.runner import obj_from_dict
print(A.__version__)
class Augtor(object):
    
    def __init__(self,
                 albumentations = None):
        self.abmtt_transforms = []
        
        if albumentations is not None:
            for _transform in albumentations:                
                self.abmtt_transforms.append(obj_from_dict(_transform, parent=A))
                
    def __call__(self, img):
        img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
        for transform in self.abmtt_transforms: #rgb
            img = transform(image=img)['image']
        img = cv2.cvtColor(img, cv2.COLOR_RGB2BGR)
        return img
sampler = Augtor(albumentations=[
               dict(
                    max_h_size=100,
                    max_w_size=100,
                    p=1.0,
                    num_holes = 2,
                    fill_value=[255,0,0],#rgb
                    type='Cutout'
                )
])
I = cv2.imread(img_fname)
sample_img = sampler(I[:1000,:1000])
plt.imshow(cv2.cvtColor(sample_img,cv2.COLOR_BGR2RGB).astype(np.int))
plt.show()

Thanks for your reply! the 'img_fname' have to be set by yourself.
The following image is my result.

image

albumentations/core/transforms_interface.py
```diff
@@ -95,7 +95,11 @@ def targets(self):
     def update_params(self, params, **kwargs):
         if hasattr(self, 'interpolation'):
             params['interpolation'] = self.interpolation
-        params.update({'cols': kwargs['image'].shape[1], 'rows': kwargs['image'].shape[0]})
+        if hasattr(self, 'fill_value'):
+            params['interpolation'] = self.fill_value
```
albu: it's strange to set interpolation as fill value

bfialkoff:
Not strange, just plain incorrect! That's what I get for copy-pasting. I
fixed it and updated the PR.

albumentations/core/transforms_interface.py
bfialkoff: Not strange, just plain incorrect! That's what I get for copy-pasting. I
fixed it and updated the PR.

PR #269, opened 2019-06-12
Added brightness / contrast integer version for an extra speedup

ternaus:
Adopted from https://github.com/MIPT-Oulu/solt

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -34,7 +34,6 @@ def clipped(func):
 
 def preserve_shape(func):
     """Preserve shape of the image."""
-
     @wraps(func)
     def wrapped_function(img, *args, **kwargs):
         shape = img.shape
@@ -47,7 +46,6 @@ def preserve_shape(func):
 
 def preserve_channel_dim(func):
     """Preserve dummy channel dim."""
-
     @wraps(func)
     def wrapped_function(img, *args, **kwargs):
         shape = img.shape
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -34,6 +34,7 @@ def clipped(func):
 
 def preserve_shape(func):
     """Preserve shape of the image."""
+
     @wraps(func)
     def wrapped_function(img, *args, **kwargs):
         shape = img.shape
@@ -46,6 +47,7 @@ def preserve_shape(func):
 
 def preserve_channel_dim(func):
     """Preserve dummy channel dim."""
+
     @wraps(func)
     def wrapped_function(img, *args, **kwargs):
         shape = img.shape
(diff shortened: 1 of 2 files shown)


BloodAxe:
According to docs https://docs.opencv.org/master/d2/de8/group__core__array.html#gab55b8d062b7f5587720ede032d34156f, LUT operates on uint8 types only. That makes sense, since lookup tables for uint16 would occupy CPU cache making the LUT operation inefficient. To me this PR has a value of speeding up transform for uint8 types. However I have few suggestions:

1) Can you please add unit tests to ensure that transform in uint16 and uint32 types works as intended.
2) Update table of performance comparison between Albumentations and the rest :)

albumentations/augmentations/functional.py
```diff
@@ -876,16 +876,42 @@ def gauss_noise(image, gauss):
...
+    dtype = np.dtype('uint8')
+
+    max_value = MAX_VALUES_BY_DTYPE[dtype]
+
+    lut = np.arange(0, max_value + 1).astype('float32')
+
+    if alpha != 1:
+        lut *= alpha
+    if beta != 0:
+        lut += beta * np.mean(img)
```
albu: why do you think that 
```
    if alpha != 1:
        lut *= alpha
    if beta != 0:
        lut += beta * np.mean(img)
```
and applying two actions after it
is faster then
```
    if alpha != 1:
        img *= alpha
    if beta != 0:
        img += beta * np.mean(img)
```

ternaus: Iw ould belive that power of the 1d array [0, 255] + look up in the table is faster than of the 2d array (N, N, 3)

before on 200 images, using our benchmark
```
Brightness                     56
Contrast                       60
BrightnessContrast             51
```

With the speedup:

```
Brightness                    168
Contrast                      363
BrightnessContrast            172

```

albumentations/augmentations/functional.py
```diff
@@ -876,16 +876,42 @@ def gauss_noise(image, gauss):
     return image + gauss
 
 
-@clipped
-def brightness_contrast_adjust(img, alpha=1, beta=0):
+def _brightness_contrast_adjust_non_unit(img, alpha=1, beta=0):
```
albu:
```suggestion
def _brightness_contrast_adjust_non_uint(img, alpha=1, beta=0):
```

albumentations/augmentations/functional.py
```diff
@@ -876,16 +876,42 @@ def gauss_noise(image, gauss):
...
     img = img.astype('float32')
+
     if alpha != 1:
         img *= alpha
     if beta != 0:
         img += beta * np.mean(img)
     return img
 
 
+@preserve_shape
```
albu: why do we need to preserve shape here?

tests/test_functional.py
```diff
@@ -845,3 +845,42 @@ def test_is_multispectral_image():
 
     gray_image = np.ones((5, 5, 1), dtype=np.uint8)
     assert not F.is_multispectral_image(gray_image)
+
+
+def test_is_brightness_contrast():
```
albu:
```suggestion
def test_brightness_contrast():
```

albumentations/augmentations/functional.py
ternaus: `img = cv2.LUT(img, lut)`

changes the shape from `(1, N, N)` to `(N, N)`

PR #270, opened 2019-06-12
gauss noie can be negative

ternaus:


albu:
What is usecase? Don't it break clipping?

ternaus:
I would say, it is just the definition of the Gaussian Noise. Should have zero mean.

We clip on both sides [0, 255] => Everything should work.

PR #278, opened 2019-06-25
Added zigsaw transform

ternaus:


albumentations/augmentations/functional.py
```diff
@@ -1177,3 +1179,59 @@ def py3round(number):
...
+        seq:
+        num_horizontal_patches:
+
+    Returns:
+
+    """
+    return np.concatenate([np.concatenate(p, axis=1) for p in partition(num_horizontal_patches, seq)], axis=0)
+
+
+def zigsaw(image, num_cuts, random_state=None, **kwargs):
```
BloodAxe: Maybe put `_combine` and `_split` inside `zigsaw`?

PR #283, opened 2019-07-01
Separate fill values for image and mask targets

BloodAxe:
Added `mask_value` parameters to specify fill value applied to masks.


setup.py
```diff
@@ -35,7 +35,7 @@ def get_long_description():
     license='MIT',
     url='https://github.com/albu/albumentations',
     packages=find_packages(exclude=['tests']),
-    install_requires=['numpy>=1.11.1', 'scipy', 'opencv-python-headless', 'imgaug>=0.2.5,<0.2.7', 'PyYAML'],
+    install_requires=['numpy>=1.11.1', 'scipy', 'opencv-python>4.1.0', 'imgaug>=0.2.5,<0.2.7', 'PyYAML'],
```
creafz: Are there any problems with OpenCV older than 4.1.0 or it is chosen to be a recent enough version?

BloodAxe: Bumped to most recent version

BloodAxe: I reverted back versionless `opencv-python-headless` to have more clear changeset.

PR #292, opened 2019-07-11
Add `get_transform_init_args_names` to `NoOp`

BloodAxe:
Add missing implementation for `get_transform_init_args_names` to `NoOp` to make it serializable


Dipet:
@BloodAxe what about this PR?

PR #310, opened 2019-08-12
Solarize transform from Pillow

Dipet:


albumentations/augmentations/transforms.py
```diff
@@ -1554,6 +1554,40 @@ def get_transform_init_args_names(self):
...
+
+        """
+    def __init__(self, threshold=128, always_apply=False, p=0.5):
+        super(Solarize, self).__init__(always_apply, p)
+
+        if isinstance(threshold, (int, float)):
+            threshold = max(0, threshold)
+            self.threshold = (threshold, threshold)
+        else:
+            self.threshold = to_tuple(threshold, low=0)
```
albu: can we use just `to_tuple(threshold, low=threshold)` here?

Dipet: Yes it is look better

albu:
please note that library is py27-compatible so it does not support typing.

other problems reported by CI:
```
./albumentations/augmentations/transforms.py:1558:1: D204 1 blank line required after class docstring
./albumentations/augmentations/transforms.py:1558:1: D208 Docstring is over-indented
```

Dipet:
Fixed.

ternaus:
@Dipet Could you please add a test that shows that this implementation gives the same result as one by the Pillow?

Dipet:
@ternaus Done

ternaus:
:+1:

PR #311, opened 2019-08-12
Random grid shuffle

toshiks:
New augmentation for shuffle grid's cells

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -1290,16 +1290,16 @@ def split_and_shuffle_shape_by_grid(shape, grid, random_state=42):
     shifted_index_height_matrix = height_matrix[1:, 1:]
     shifted_index_width_matrix = width_matrix[1:, 1:]
 
-    height_box_sizes = shifted_index_height_matrix - index_height_matrix
-    width_box_sizes = shifted_index_width_matrix - index_width_matrix
+    height_tile_sizes = shifted_index_height_matrix - index_height_matrix
+    width_tile_sizes = shifted_index_width_matrix - index_width_matrix
 
-    bbox_sizes = np.stack((height_box_sizes, width_box_sizes), axis=2)
+    tiles_sizes = np.stack((height_tile_sizes, width_tile_sizes), axis=2)
 
     index_matrix = np.indices((n, m))
     new_index_matrix = np.stack(index_matrix, axis=2)
 
-    for bbox_size in np.unique(bbox_sizes.reshape(-1, 2), axis=0):
-        eq_mat = np.all(bbox_sizes == bbox_size, axis=2)
+    for bbox_size in np.unique(tiles_sizes.reshape(-1, 2), axis=0):
+        eq_mat = np.all(tiles_sizes == bbox_size, axis=2)
         new_index_matrix[eq_mat] = np.random.permutation(new_index_matrix[eq_mat])
 
     new_index_matrix = np.split(new_index_matrix, 2, axis=2)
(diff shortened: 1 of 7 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -912,6 +912,53 @@ def get_transform_init_args_names(self):
...
+            "bboxes": F.split_and_shuffle_shape_by_grid((rows, cols), self.grid, random_state),
+        })
+
+        return params
+
+    def apply(self, img, bboxes=None, **params):
+        if bboxes is None:
+            bboxes = []
+
+        return F.swap_bboxes_on_image(img, bboxes)
```
ternaus: The name `bboxes` may be misleading.

Maybe `tiles` ?

albumentations/augmentations/transforms.py
```diff
@@ -912,6 +912,53 @@ def get_transform_init_args_names(self):
...
+
+    def update_params(self, params, **kwargs):
+        params = super(RandomGridShuffle, self).update_params(params, **kwargs)
+        rows = params['rows']
+        cols = params['cols']
+
+        random_state = random.randint(0, 10000)
+
+        params.update({
+            "bboxes": F.split_and_shuffle_shape_by_grid((rows, cols), self.grid, random_state),
```
ternaus: I would, probably replace

`random_state` by `np.random.RandomState(random_state)`

and avoid setting seed in `split_and_shuffle_state_by_grid`, similar to https://github.com/albu/albumentations/blob/master/albumentations/augmentations/transforms.py#L1854

But I do not have strong opinion.

@albu @BloodAxe @creafz what do you think?

tests/test_augmentations.py
```diff
@@ -141,12 +143,12 @@ def test_imgaug_image_only_augmentations(augmentation_cls, image, mask):
     assert np.array_equal(data['mask'], mask)
 
 
-@pytest.mark.parametrize('augmentation_cls', [IAAPiecewiseAffine, IAAPerspective])
-def test_imgaug_dual_augmentations(augmentation_cls, image, mask):
-    aug = augmentation_cls(p=1)
-    data = aug(image=image, mask=mask)
-    assert data['image'].dtype == np.uint8
-    assert data['mask'].dtype == np.uint8
+# @pytest.mark.parametrize('augmentation_cls', [IAAPiecewiseAffine, IAAPerspective])
```
ternaus: Why is this commented?

albumentations/augmentations/transforms.py
toshiks: I followed the next example: https://github.com/albu/albumentations/blob/ef68cd48cb9743f25846913965e2143a98f062a5/albumentations/augmentations/functional.py#L789

tests/test_augmentations.py
toshiks: My mistake, I'm sorry

albumentations/augmentations/transforms.py
toshiks: Yep. It sounds great! Fixed!

albumentations/augmentations/functional.py
```diff
@@ -1256,3 +1256,80 @@ def py3round(number):
...
+def split_and_shuffle_shape_by_grid(shape, grid, random_state=42):
+    """
+    Split image by grid.
+
+    Args:
+        shape (tuple): image shape.
+        grid (tuple): tuple with two values.
+        random_state (int): seed for numpy
+    """
+    np.random.seed(random_state)
```
albu: we don't modify random seed in functions. we instantiate all params in class. please see how it's done in other classes

albumentations/augmentations/transforms.py
```diff
@@ -912,6 +912,53 @@ def get_transform_init_args_names(self):
...
+
+    Image types:
+        uint8, float32
+    """
+
+    def __init__(self, grid=(3, 3), always_apply=False, p=1.0):
+        super(RandomGridShuffle, self).__init__(always_apply, p)
+        self.grid = grid
+
+    def update_params(self, params, **kwargs):
```
albu: not sure if we use `update_params`.. looks like you need `get_params` or `get_params_dependent_on_targets`

albumentations/augmentations/transforms.py
```diff
@@ -912,6 +912,53 @@ def get_transform_init_args_names(self):
...
+        params = super(RandomGridShuffle, self).update_params(params, **kwargs)
+        rows = params['rows']
+        cols = params['cols']
+
+        random_state = random.randint(0, 10000)
+
+        params.update({
+            "tiles": F.split_and_shuffle_shape_by_grid((rows, cols), self.grid, random_state),
```
albu: F (aka functional) as only for transforms, not for params generation.

albu:
please also provide examples of what transform does

toshiks:
There is example of transformation. Split the image into tiles by the grid and shuffle them

Example

albumentations/augmentations/transforms.py
toshiks: Fixed all requests.

albu:
How should it work for bounding boxes?

toshiks:
It should work only with masks and keypoints (not implemented).

ternaus:
@albu For bounding boxes, it is undefined unless we will find a meaningful way to define them.

albumentations/augmentations/transforms.py
```diff
@@ -912,6 +912,96 @@ def get_transform_init_args_names(self):
...
+        if tiles is None:
+            tiles = []
+
+        return F.swap_tiles_on_image(img, tiles)
+
+    def get_params_dependent_on_targets(self, params):
+        height, width = params['image'].shape[:2]
+        n, m = self.grid
+
+        if n == 0 or m == 0:
```
albu:
```suggestion
        if n <= 0 or m <= 0:
```

albumentations/augmentations/functional.py
```diff
@@ -1256,3 +1256,20 @@ def py3round(number):
...
     return input_obj
+
+
+def swap_tiles_on_image(image, tiles):
+    """
+    Swap tiles on image.
+
+    Args:
+        image (np.ndarray): Input image.
+        tiles (np.ndarray): array of tuples with bounding boxes
```
albu: indexing is not obvious here, please make a class or namedtuple or something like this because it's not clear what is tiles without digging into a code

toshiks: I supplemented the description

PR #312, opened 2019-08-13
Changed JPEGCompress to ImageCompress for supporting webp

### LINKED ISSUE #220: Add WebP transform
The work was started in https://github.com/albu/albumentations/pull/165

but the code looks very similar to JpegCompression, hence it will make sense to merge them into one transform.

toshiks:
Works fine with the lastets OpenCV version (previous version had bug: https://github.com/opencv/opencv/issues/14978)

It should resolve #220


albumentations/augmentations/transforms.py
```diff
@@ -1064,12 +1065,14 @@ def get_transform_init_args_names(self):
                 'min_height', 'min_width')
 
 
-class JpegCompression(ImageOnlyTransform):
```
albu: we cannot change api, see how we deprecate things in `RandomBrightness` and `RandomContrast`

toshiks: I'm so sorry. It's my mistake. I fixed that.

albumentations/augmentations/functional.py
```diff
@@ -447,21 +447,26 @@ def motion_blur(img, kernel):
 
 
 @preserve_shape
-def jpeg_compression(img, quality):
+def image_compression(img, quality, image_type):
+    quality_flag = cv2.IMWRITE_JPEG_QUALITY
```
ternaus: I would probably make it something like:

```
if image_type == '.jpeg`:
    quality_flag = cv2.IMWRITE_JPEG_QUALITY
elif image_type == '.webp
    quality_flag = cv2.IMWRITE_WEBP_QUALITY
else:
   NotimplementedError("Only '.jpeg' and '.webp' compression transforms are implemented. ")
```

toshiks: It looks better :)

albu:
will be merged after fixing ci errors

toshiks:
opencv-python-headless doen't contain lastets OpenCV version (previous version had bug: opencv/opencv#14978). So, I can't fix ci errors

ternaus:
@toshiks could you please check if it can be merged?

toshiks:
Master is merged. Now it looks good

PR #314, opened 2019-08-14
Added page that describes how to solve OpenCV + multiprocessing

ternaus:


albu:
is it the same as https://github.com/albu/albumentations#comments ?

ternaus:
Yep.

albu:
so why do we need to tell the same thing twice?

ternaus:
The experiment showed that not everyone can get to the bottom of the long README.

I believe, that if users are missing it there, we should duplicate it somewhere else and documentation looks like a good place for this.

albu:
let's remove it from readme then? :)

ternaus:
Professional deformation from software engineering is so strong in you :)

Duplication is not always evil. If we knew that there is an amazing place to put it once so that people would see it => sure, should be mentioned once.

But because we do not know such a great place, we can put it in two ok places.

albu:
but if we want to change it to something - we will change it in only one place and will get two different versions of the same thing. life is like programming :)

ternaus:
> but if we want to change it to something - we will change it in only one place and will get two different versions of the same thing. life is like programming :)

Do you propose parsing readme and adding the info to doc or parsing doc to add info to readme?

When we will come to a better structure for docs we can make it more `high style`.

Our readme is long, and I am not sure that docs are in the most useful form. But that is what we have.

PR #315, opened 2019-08-14
Bugfix in GaussNoise

ternaus:


Dipet:
var_limit must be >= 0, because function use square from this value

albumentations/augmentations/transforms.py
```diff
@@ -1791,30 +1791,39 @@ class GaussNoise(ImageOnlyTransform):
 
     Args:
         var_limit ((float, float) or float): variance range for noise. If var_limit is a single float, the range
-            will be (-var_limit, var_limit). Default: (10., 50.).
+            will be (-var_limit, var_limit). Default: (10.0, 50.0).
```
albu: here you make var_limit[0] negative and raise ValueError right after it

albumentations/augmentations/transforms.py
```diff
@@ -1791,30 +1791,39 @@ class GaussNoise(ImageOnlyTransform):
...
         super(GaussNoise, self).__init__(always_apply, p)
+        if isinstance(var_limit, tuple):
+            if var_limit[0] < 0:
+                raise ValueError("Lower var_limit should be non negative.")
+            if var_limit[1] < 0:
+                raise ValueError("Upper var_limit should be non negative.")
+        elif isinstance(var_limit, float):
+            raise ValueError(" var_limit should be non negative.")
+
         self.var_limit = to_tuple(var_limit)
```
albu: why do you check conditions before making a tuple?

albumentations/augmentations/transforms.py
```diff
@@ -1791,30 +1791,39 @@ class GaussNoise(ImageOnlyTransform):
...
+        self.mean = mean
 
     def apply(self, img, gauss=None, **params):
         return F.gauss_noise(img, gauss=gauss)
 
     def get_params_dependent_on_targets(self, params):
         image = params['image']
         var = random.uniform(self.var_limit[0], self.var_limit[1])
-        mean = var
```
albu: looks like it changes current behavior so it breaks existing user pipelines. we should not do something like this without major version update

albumentations/augmentations/transforms.py
```diff
@@ -1802,17 +1802,20 @@ class GaussNoise(ImageOnlyTransform):
...
-    def __init__(self, var_limit=(10.0, 50.0), mean=0, always_apply=False, p=0.5):
+    def __init__(self, var_limit=(10.0, 50.0), mean=None, always_apply=False, p=0.5):
         super(GaussNoise, self).__init__(always_apply, p)
         if isinstance(var_limit, tuple):
             if var_limit[0] < 0:
                 raise ValueError("Lower var_limit should be non negative.")
             if var_limit[1] < 0:
                 raise ValueError("Upper var_limit should be non negative.")
+            self.var_limit = var_limit
         elif isinstance(var_limit, float):
```
albu: `int`?

Issue #313, opened 2019-08-14
GaussNoise not centered at 0


TontonTremblay:
Hello, 

We were looking at the GaussNoise function and noticed that the noise is not centered at zero. 

https://github.com/albu/albumentations/blob/master/albumentations/augmentations/transforms.py#L1814

Is that on purpose? We have a PR ready for that one if you are interested.

ternaus:
Fixed in https://github.com/albu/albumentations/pull/315

PR #316, opened 2019-08-15
Changed GaussNoise mean to zero, instead of setting mean to var

timeous:
Hello, this is in reference to Issue #313. @TontonTremblay and I noticed that the setting of `GaussNoise` `mean` is set to `var`. Therefore, `GaussNoise` will always brighten the image, because the resulting noise vector `gauss` has non-zero mean.

I believe the intended behavior is to apply Gaussian white noise (i.e., zero-mean noise) to the image, which it does now after this fix. This also means that, with this fix, `GaussNoise` will no longer always brighten the image.


ternaus:
Working on this in https://github.com/albu/albumentations/pull/315

ternaus:
Fixed in https://github.com/albu/albumentations/pull/315

PR #318, opened 2019-08-15
deprecate totensor

albu:
current ToTensor version leads to a lot of questions. we have to deprecate it because it's bad

### DIFF
--- a/albumentations/pytorch/transforms.py
+++ b/albumentations/pytorch/transforms.py
@@ -3,11 +3,12 @@ from __future__ import absolute_import
 import numpy as np
 import torch
 from torchvision.transforms import functional as F
+import warnings
 
 from ..core.transforms_interface import BasicTransform
 
 
-__all__ = ['ToTensor']
+__all__ = ['ToTensor', 'ImgToTensor']
 
 
 def img_to_tensor(im, normalize=None):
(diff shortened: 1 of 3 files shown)


albumentations/pytorch/transforms.py
```diff
@@ -53,6 +71,7 @@ def __init__(self, num_classes=1, sigmoid=True, normalize=None):
         self.num_classes = num_classes
         self.sigmoid = sigmoid
         self.normalize = normalize
+        warnings.warn("ToTensor is deprecated and will be removed in albumentations 1.0.0", DeprecationWarning)
```
ternaus: Maybe:

in 0.4 ?

albu: ok

vfdev-5:
Would you be interested by the following `ToTensor` implementation : 
```python
class ToTensor(BasicTransform):

    def __init__(self):
        super(ToTensor, self).__init__(always_apply=True)

    @property
    def targets(self):
        return {
            'image': self.apply,
            'mask': self.apply_to_mask
        }

    def apply(self, img, **params):
        return torch.from_numpy(img.transpose(2, 0, 1))

    def apply_to_mask(self, mask, **params):
        return torch.from_numpy(mask)
```

albu:
looks good

Issue #295, opened 2019-07-16
Keypoints pre- and post- processing support


GazizovMarat:
Hi, Team. When I try to pass N sets of keypoints with 'xy' format, the first set will be pre- and post- processed into internal format, but other will not. It raises an exception.

ternaus:
Could you please share an example code to reproduce an issue?

GazizovMarat:
https://colab.research.google.com/drive/1S5Lm3dYHli1CXq5MEIWzc3mWJhDFvqNa

bonlime:
@ternaus @GazirovMarat
I already made a PR fixing this issue. 
https://github.com/albu/albumentations/pull/287
Just merge it and everything should be fine.

PR #319, opened 2019-08-16
Refactor Compose

albu:
Compose now can take `KeypointParams` or `BboxParams` as arguments.
Too hard logic splitted to classes, now it's much easier to understand.
Fix bug #295
Move tansforms processing logic into a class


albumentations/augmentations/keypoints_utils.py
```diff
@@ -1,5 +1,98 @@
 from __future__ import division
 import math
+import warnings
+
+from albumentations.core.utils import Params, DataProcessor
+
+__all__ = ['check_keypoints', 'convert_keypoints_from_albumentations', 'convert_keypoints_to_albumentations',
+           'filter_keypoints', 'KeypointParams']
+
+keypoint_formats = {'xy', 'yx', 'xya', 'xys', 'xyas', 'xysa'}
```
ternaus: Could you please add a comment that describes what x, y, s, a are?

Or/and add a page to the documentation that has this info.

albu: @BloodAxe we need you

BloodAxe: x - X coordinate,
y - Y coordinate
s - Keypoint scale
a - Keypoint orientation in radians or degrees (depending on KeypointParams.angle_in_degrees)

albumentations/augmentations/bbox_utils.py
```diff
@@ -1,10 +1,88 @@
...
 import numpy as np
 
 __all__ = ['normalize_bbox', 'denormalize_bbox', 'normalize_bboxes', 'denormalize_bboxes', 'calculate_bbox_area',
            'filter_bboxes_by_visibility', 'convert_bbox_to_albumentations', 'convert_bbox_from_albumentations',
-           'convert_bboxes_to_albumentations', 'convert_bboxes_from_albumentations']
+           'convert_bboxes_to_albumentations', 'convert_bboxes_from_albumentations', 'BboxParams']
+
+
+class BboxParams(Params):
+    def __init__(self, format=None, label_fields=None, min_area=0.0, min_visibility=0.0):
```
BloodAxe: I'd suggest to not use `format=None` to disable boxes transform. I'd rather say if `bbox_params is None` then it's disabled.

albumentations/augmentations/keypoints_utils.py
```diff
@@ -1,5 +1,98 @@
...
+
+
+class KeypointParams(Params):
+    def __init__(self, format=None, label_fields=None, remove_invisible=True, angle_in_degrees=True):
```
BloodAxe: Same here with `format=None`.

albumentations/augmentations/keypoints_utils.py
```diff
@@ -1,5 +1,98 @@
...
+        return "keypoints"
+
+    def ensure_data_valid(self, data):
+        if self.params.label_fields:
+            if not all(l in data.keys() for l in self.params.label_fields):
+                raise Exception("Your 'label_fields' are not valid - them must have same names as params in "
+                                "'keypoint_params' dict")
+
+    def ensure_transforms_valid(self, transforms):
+        pass
```
BloodAxe: pass?

albumentations/core/utils.py
```diff
@@ -10,3 +11,74 @@ def format_args(args_dict):
             v = "'{}'".format(v)
         formatted_args.append('{}={}'.format(k, v))
     return ', '.join(formatted_args)
+
+
+class Params:
+    __metaclass__ = ABCMeta
```
creafz: If I am not mistaken this syntax works only for Python 2. For Python 3 you have to pass a metaclass as a `metaclass` argument to the class definition. To make this class compatible both with Python 2 and Python 3 you can use `with_metaclass` (example - https://github.com/albu/albumentations/blob/master/albumentations/core/transforms_interface.py#L45)

creafz:
Maybe we should use `ValueError` instead of a generic `Exception` class for exceptions that we raise when input parameters are incorrect?

albumentations/augmentations/bbox_utils.py
```diff
@@ -1,10 +1,55 @@
...
+
+        for data_name in self.data_fields:
+            data[data_name] = filter_bboxes(data[data_name], rows, cols,
+                                            min_area=self.params.min_area, min_visibility=self.params.min_visibility)
+
+            if self.params.format == 'albumentations':
+                check_bboxes(data[data_name])
+            else:
+                data[data_name] = convert_bboxes_from_albumentations(data[data_name], self.params.format, rows, cols,
+                                                                     check_validity=True)
```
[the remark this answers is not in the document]
albu: refactored this all as base methods

albumentations/core/utils.py
albu: it works, but added with_metaclass to be 100% sure

albu:
fixed all comments. @BloodAxe @creafz

PR #320, opened 2019-08-20
Histogram equalization transform

Dipet:
Transform supports 2 histogram equalization modes:
1. OpenCV;
2. Pillow.

This methods gets a little bit different results.

Supports Grayscale and RGB images. If image is RGB we can process channels separately or convert image to YCrCb mode and equalize only Y (intensity channel) values.

Also, we can get mask for equalization.  If given, only the pixels selected by the mask are included in the analysis. But I can not figure out how add this functional to transform interface.

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -308,7 +308,7 @@ def shift_hsv(img, hue_shift, sat_shift, val_shift):
     return img
 
 
-def solarize(img: np.ndarray, threshold=128):
+def solarize(img, threshold=128):
     """Invert all pixel values above a threshold.
 
     Args:
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -322,9 +322,15 @@ def solarize(img: np.ndarray, threshold=128):
     dtype = img.dtype
     max_val = MAX_VALUES_BY_DTYPE[dtype]
 
-    if isinstance(dtype, (np.uint8, np.uint16, np.uint32)):
+    if dtype == np.dtype('uint8'):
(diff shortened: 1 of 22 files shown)


albumentations/augmentations/functional.py
```diff
@@ -338,6 +338,111 @@ def solarize(img, threshold=128):
...
+        by_channels (bool): If True, use equalization by channels separately,
+            else convert image to YCbCr representation and use equalization by `Y` channel.
+
+    Returns:
+        Equalized image.
+
+    """
+    modes = ['cv', 'pil']
+
+    if mode not in modes:
```
ternaus: Could you please add a check for dtype of the input image?

BloodAxe:
@Dipet I have two questions related this PR:
1) Do we really need two (PIL and CV) histogram equalization implementations? To me - one is enough. AFAIK, the only place where we use PIL is IAA-based augmentations, but rest of library is PIL-free, and I'd like to keep that.
2) It seems, the code for CV implementation can be simplified by replacing histogram building and LUT by `cv2.equalizeHist` https://docs.opencv.org/master/d6/dc7/group__imgproc__hist.html#ga7e54091f0c937d49bf84152a16f76d6e. Can we use it instead of computing histogram manualy? Unless there is real need to do manual calculations, we should try our best to use library functions.

Dipet:
@BloodAxe Sorry for the long silence.
1. Pillow and OpenCV get different equalization results, for somebody it may by important, and i think 2 implementations is not a big overhead, because they are simple. Current realization does not use Pillow for transform, I use only numpy and OpenCV functions. Pillow used only in test functions to show thats Pillow and my implementation gets equal results.
2. I use histogram calculations only for equalization by mask case, because OpenCV does not support equalization by mask.

PR #329, opened 2019-09-01
brightness_contrast_adjust optimization

Dipet:
Before
```
                   albumentations imgaug torchvision keras augmentor  solt
Brightness                    128    529          77    25        80  1055
Contrast                      183    592          58     -        61   930
BrightnessContrast            140    340          34     -        33   527
```

After
```
                   albumentations imgaug torchvision keras augmentor  solt
Brightness                    987    594          80    24        81   984
Contrast                     1227    608          59     -        61  1191
BrightnessContrast           1266    356          35     -        35   373
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ Pixel-level transforms will change just an input image and will leave any additi
 - [ChannelShuffle](https://albumentations.readthedocs.io/en/latest/api/augmentations.html#albumentations.augmentations.transforms.ChannelShuffle)
 - [CoarseDropout](https://albumentations.readthedocs.io/en/latest/api/augmentations.html#albumentations.augmentations.transforms.CoarseDropout)
 - [Cutout](https://albumentations.readthedocs.io/en/latest/api/augmentations.html#albumentations.augmentations.transforms.Cutout)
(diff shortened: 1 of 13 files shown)


ternaus:
This speedup looks tempting. But we get it because of the change in the algorithm.

Could you please share what was the motivation behind it? Is it the way it is implemented in other libraries?

Dipet:
1. I remove clipping from uint8 version (~2 times faster )
2.  Remove mean calculation from image and use dtype maximum.
imgaug [Add](https://github.com/aleju/imgaug/blob/38f305aaa5df833228d95d95de36e77d1a4ac88c/imgaug/augmenters/arithmetic.py#L51) and solt [ImageRandomBrightness](https://github.com/MIPT-Oulu/solt/blob/a6891d4dfa6bfd0d01a4ca1d6e7c03d15468d9c3/solt/transforms/_transforms.py#L1030) transforms use values for add from your interval but it will be in range [-dtype.max, dtype.max].
For example, we set interval [-25, 25] in this implementation it will be equivalent to [-25/256, 25/256]

ternaus:
Benchmark and default behavior are different now. 

Let's delete the changes in the benchmark file. In version 0.4, they will start working automatically.

Dipet:
@ternaus Benchmark changes removed

PR #331, opened 2019-09-01
gamma_transform optimization

Dipet:
Before
```
                   albumentations imgaug torchvision keras augmentor solt
Gamma                         527      -         172     -         -  602
```

After
```
                   albumentations imgaug torchvision keras augmentor solt
Gamma                         942      -         179     -         -  599
```

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -910,7 +910,7 @@ def channel_dropout(img, channels_to_drop, fill_value=0):
 def gamma_transform(img, gamma):
     if img.dtype == np.uint8:
         invGamma = 1.0 / gamma
-        table = ((np.arange(0, 256) / 255) ** invGamma) * 255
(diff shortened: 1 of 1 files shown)


albumentations/augmentations/functional.py
```diff
@@ -910,8 +910,8 @@ def channel_dropout(img, channels_to_drop, fill_value=0):
 def gamma_transform(img, gamma):
     if img.dtype == np.uint8:
         invGamma = 1.0 / gamma
-        table = np.array([((i / 255.0) ** invGamma) * 255 for i in np.arange(0, 256)]).astype("uint8")
-        img = cv2.LUT(img, table)
+        table = ((np.arange(0, 256) / 255) ** invGamma) * 255
```
ternaus: It looks like `((np.arange(0, 256 / 255 , 1.0 / 255) ** invGamma) * 255)` is slightly faster.

Dipet: done

PR #332, opened 2019-09-01
shift_hsv optimization

Dipet:
Before
```
                   albumentations imgaug torchvision keras augmentor solt
ShiftHSV                      578    480         123     -         -  219
```

After
```
                   albumentations imgaug torchvision keras augmentor solt
ShiftHSV                     1116    478         121     -         -  200
```

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -291,28 +291,33 @@ def clamping_crop(img, x_min, y_min, x_max, y_max):
     return img[int(y_min):int(y_max), int(x_min):int(x_max)]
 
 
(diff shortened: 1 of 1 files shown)


albumentations/augmentations/functional.py
```diff
@@ -294,9 +294,26 @@ def clamping_crop(img, x_min, y_min, x_max, y_max):
 def shift_hsv(img, hue_shift, sat_shift, val_shift):
```
ternaus: I would probably split `shift_hsv` to `shift_hsv_uint ` and `shift_hsv_non_uint`  similar to how it is done in the `adjust_brightness_contrast`

PR #333, opened 2019-09-01
Posterize transform from Pillow

Dipet:
Bench:
```
               albumentations imgaug torchvision keras augmentor  solt pillow
Posterize               10794      -           -     -         -     -   4100
```

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -338,6 +338,29 @@ def solarize(img, threshold=128):
     return result_img
 
 
+@preserve_shape
+def posterize(img, bits):
+    assert img.dtype == np.uint8, 'Image must have uint8 channel type'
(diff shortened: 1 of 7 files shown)


albumentations/augmentations/functional.py
```diff
@@ -338,6 +338,29 @@ def solarize(img, threshold=128):
     return result_img
 
 
+@preserve_shape
+def posterize(img, bits):
+    assert img.dtype == np.uint8, 'Image must have uint8 channel type'
```
BloodAxe: Could you please add comment here to clarify what possible values `bits` can be?

Dipet: done

ternaus:
@Dipet could you please resolve conflicts, please?

PR #335, opened 2019-09-02
Added RandomSizedCropTorchVision

### LINKED ISSUE #259: is there a transform work like RandomResizedCrop in torchvison
I find that RandomResizedCrop work well on my dataset, But i can't find in the albumentation.
I try to replace  by using RandomSizedCrop and RandomCrop,  the result is not good.

vfdev-5:
WIP PR to fix #259

cc @ternaus

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -667,7 +667,7 @@ class RandomCropNearBBox(DualTransform):
 
 
 class _BaseRandomSizedCrop(DualTransform):
-    # Base class for RandomSizedCropOld and RandomSizedCropTorchVision
+    # Base class for RandomSizedCrop and RandomResizedCrop
 
     def __init__(self, height, width, interpolation=cv2.INTER_LINEAR, always_apply=False, p=1.0):
         super(_BaseRandomSizedCrop, self).__init__(always_apply, p)
@@ -716,8 +716,7 @@ class RandomSizedCrop(_BaseRandomSizedCrop):
                                               interpolation=interpolation,
                                               always_apply=always_apply, p=p)
         self.min_max_height = min_max_height
-        self.w2h_ratio = w2h_ratio
-        warnings.warn("This class has been deprecated. Please use RandomSizedCropTorchVision", DeprecationWarning)
+        self.w2h_ratio = w2h_ratio        
 
     def get_params(self):
         crop_height = random.randint(self.min_max_height[0], self.min_max_height[1])
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -666,7 +666,7 @@ class RandomCropNearBBox(DualTransform):
         return ('max_part_shift',)
 
 
-class RandomSizedCrop(DualTransform):
+class RandomSizedCropOld(DualTransform):
     """Crop a random part of the input and rescale it to some size.
 
     Args:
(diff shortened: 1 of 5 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -717,7 +718,112 @@ def apply_to_keypoint(self, keypoint, crop_height=0, crop_width=0, h_start=0, w_
...
+
+    def get_params(self):
+        return {}
+
+    @property
+    def targets_as_params(self):
+        return ['image']
+
+
+def RandomSizedCrop(*args, mode='old', **kwargs):
```
albu: looks like we can pass height and width for both transforms so no need to use args/kwargs for them. for other arguments it seems like we can add here all params (with same defaults) from new version and just use args, kwargs for old.

vfdev-5: I see what what you mean, sounds good. What about `mode='old'` and "add here all params (with same defaults) from new version", this would be a bit strange ?

albu: yeah it's a bit strange but doesn't break anything (but changing to mode='torchvision' breaks). at the same time it tries to push users to new api (gently).

albu: oh, I see what happens, agree with you - it's strange and weird :(

albu: so new suggestion. if user passes `min_max_height` - run old version. otherwise run new. how it sounds?

vfdev-5:
I refactored transforms to avoid some deprecation warnings with `RandomSizedCropTorchVision` and updated input args of `RandomSizedCrop`.

@albu Are you ok with that `RandomSizedCrop` is a **method** and not a class ?

albumentations/augmentations/transforms.py
vfdev-5: I'm not sure how to determine whether user passes `min_max_height`. For example, in your tests:
https://github.com/albu/albumentations/blob/448761df9a008384cf914343f25e3cfb7c4d7551/tests/test_bbox.py#L189

In this case we need to check what type of input is passed. So decision based on type of input seems not very safe ?

vfdev-5: > yeah it's a bit strange but doesn't break anything (but changing to mode='torchvision' breaks). at the same time it tries to push users to new api (gently).

Yes, you are right, we need to keep `mode='old'` to avoid BC breaks.

albu:
> I refactored transforms to avoid some deprecation warnings with `RandomSizedCropTorchVision` and updated input args of `RandomSizedCrop`.
> 
> @albu Are you ok with that `RandomSizedCrop` is a **method** and not a class ?

it could break documentation and probably something else.

I wonder if just adding new crop with different name could be better and safer solution..

vfdev-5:
> I wonder if just adding new crop with different name could be better and safer solution..

I agree. I let you decide between maintainers what names to choose.

albu:
maybe `RandomResizedCrop`?

ternaus:
It looks like Torchvision deprecated `RandomSizedCrop` in favor of `RandomResizedCrop`

I would love to see:
1. Right now (0.3) to have `RandomSizedCrop` that we had earlier.
2. `RandomResizedCrop` with the API and functionality of what torchvision has https://pytorch.org/docs/stable/torchvision/transforms.html#torchvision.transforms.RandomResizedCrop
3. Deprecation warning in `RandomSizedCrop`

=> in 0.4 we will match the behavior of Torchvision, keeping users informed of the future breaking changes.

So, to me, looks like adding second transform `RandomResizedCrop` is the best solution.

vfdev-5:
@ternaus review, please :)

albumentations/augmentations/transforms.py
```diff
@@ -688,16 +712,12 @@ class RandomSizedCrop(DualTransform):
...
+        super(RandomSizedCrop, self).__init__(height=height, width=width,
+                                              interpolation=interpolation,
+                                              always_apply=always_apply, p=p)
         self.min_max_height = min_max_height
         self.w2h_ratio = w2h_ratio
-
-    def apply(self, img, crop_height=0, crop_width=0, h_start=0, w_start=0, interpolation=cv2.INTER_LINEAR, **params):
-        crop = F.random_crop(img, crop_height, crop_width, h_start, w_start)
-        return F.resize(crop, self.height, self.width, interpolation)
+        warnings.warn("This class has been deprecated. Please use RandomSizedCropTorchVision", DeprecationWarning)
```
ternaus: Could you please fix this warning?

vfdev-5: Done

PR #336, opened 2019-09-02
Updated pytorch transforms and tests

vfdev-5:
Following https://github.com/albu/albumentations/pull/318#issuecomment-527161032 and https://github.com/albu/albumentations/pull/318#issuecomment-527173032

@albu @ternaus what do you think ?

### DIFF
--- a/albumentations/pytorch/transforms.py
+++ b/albumentations/pytorch/transforms.py
@@ -1,8 +1,6 @@
 from __future__ import absolute_import
 
-import numpy as np
 import torch
-from torchvision.transforms import functional as F
 
 from ..core.transforms_interface import BasicTransform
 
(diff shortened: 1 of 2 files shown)


albu:
looks fine but we still have to preserve old behavior for all these who uses ToTensor

vfdev-5:
@albu how do you prefer to keep the old behavior : 
- Old `ToTensor` -> `ToTensor_deprecated` with warning about removal of the old behaviour
- New `ToTensor`

or 

- Old `ToTensor` -> `ToTensor` with warning about removal of the old behaviour
- New `ToTensor` -> `ToTensorV2` and rename it `ToTensor` once in v0.4.0

ternaus:
I would prefer the second option.

albu:
second option will force developers to replace ToTensor twice. first in this release - by ToTensorV2 and in 0.4.0 ToTensorV2 -> ToTensor. it's weird

vfdev-5:
@albu 

> second option will force developers to replace ToTensor twice. first in this release - by ToTensorV2 and in 0.4.0 ToTensorV2 -> ToTensor. it's weird

I agree, but like that we are keeping BC. Assuming that users mostly use current version `ToTensor`, they are just become aware of deprecation and replacement. Others (like me) do not use `ToTensor` but their own version of it and they will just switch to `A.ToTensorV2` and since 0.4.0 to `A.ToTensor()`.

Otherwise, what do you think anout 1st option ?

albu:
first option does not work at all. so from two evils second is better. I just wonder if there are other better options..

PR #339, opened 2019-09-05
README: Delete deprecated function RandomContrast

apatsekin:
deprecated function in readme


Dipet:
Will be fixed in #596

Issue #343, opened 2019-09-08
TypeError: 'int' object is not subscriptable for RandomGamma show


Rishi5583:
## 🐛 Bug
albumentations.RandomGamma is showing  by cv2 : 'int' object is not subscriptable. same problem in google colab also 
albumentations\augmentations\transforms.py", line 2122, in get_params
    'gamma': np.random.randint(int(self.gamma_limit[0]), int(self.gamma_limit[1])) / 100.0
TypeError: 'int' object is not subscriptable.
## To Reproduce
def augment_and_show(aug, image):
    image = aug(image=image)['image']
    cv2.imshow('im',image)

>>> aug=RandomGamma(gamma_limit=70, p=1)
>>> augment_and_show(aug, image)

1.simple following the example:
In [7]:

def augment_and_show(aug, image):
    image = aug(image=image)['image']
    plt.figure(figsize=(10, 10))
    plt.imshow(image)

In [8]:

aug = HorizontalFlip(p=1)
augment_and_show(aug, image)
1.
1.

## Expected behavior

tried in transforms.py 
by changing : gamma': np.random.randint(int(self.gamma_limit[0]), int(self.gamma_limit[1])) / 100.0
It works only one time , after that again same error is showing, any clue ?

## Environment

 - Albumentations version (e.g., 0.1.8):
 - Python version (e.g., 3.6.8):
 - win 10 
 - How you installed albumentations (` `pip`, source):
 - Any other relevant information:

Dipet:
fixed in #344

PR #344, opened 2019-09-08
Fix GammaTransform int as tuple, division by zero

Dipet:
#343 
What a lower bound for gamma?
I think, values in range `[-inf, 0)` will give strange (no intuitive) results.
Maybe need call `to_tuple(gamma_limit, low=0)`?


ternaus:
I believe we need to add min/max values of all input parameters and check if input value is in allowed range. https://github.com/albu/albumentations/issues/337

PR #350, opened 2019-09-14
Deterministic/Replay mode for augmentations

albu:
Now we can create augmentations pipeline and save all params of applied augmentations for later replay or debug. See example in jupyter notebook for more info.


Dipet:
I think #372 will fix tests.

ternaus:
It would be nice to add a link to the notebook to the readme.

albu:
done

PR #358, opened 2019-09-18
Update composition.py

BelBES:
Fix Compose.__call__: image can be passed as a positional argument through 'force_apply' variable

### DIFF
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -144,6 +144,7 @@ class Compose(BaseCompose):
         self.add_targets(additional_targets)
 
     def __call__(self, force_apply=False, **data):
+        assert isinstance(force_apply, (bool, int)), 'force_apply must have bool or int type'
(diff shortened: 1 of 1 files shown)


vfdev-5:
It would be better to write 
```
if not isinstance(force_apply, (bool, int)):
    raise TypeError('force_apply must have bool or int type')
```
instead of `assert` ?

BelBES:
> It would not be better to write
> 
> ```
> if not isinstance(force_apply, (bool, int)):
>     raise TypeError('force_apply must have bool or int type')
> ```
> 
> instead of `assert` ?

So, I used 'assert' because it already used in albumentation for check type of input arguments  (for example: https://github.com/albu/albumentations/blob/master/albumentations/augmentations/functional.py#L403).

But I see that 'raise' also used (for example: https://github.com/albu/albumentations/blob/master/albumentations/augmentations/functional.py#L271).

I think it should be determined in contribution guide.

PR #359, opened 2019-09-18
Fix RandomSizedBBoxSafeCrop.

Jae-Hyuck:
To my understanding,  even though 'bboxes' is given,
'get_params_dependent_on_targets' method cannot return proper params,
because 'targets_as_params' does not include 'bboxes'.

So, I add 'bboxes' to the 'targets_as_params'.
Now, 'bboxes' is necessary to run RandomSizedBBoxSafeCrop. (An empty list is allowed.)

Please check my thinking is correct.
I also changed some test code.
Thank you.


albumentations/augmentations/transforms.py
```diff
@@ -846,7 +846,7 @@ def apply(self, img, crop_height=0, crop_width=0, h_start=0, w_start=0, interpol
 
     def get_params_dependent_on_targets(self, params):
         img_h, img_w = params['image'].shape[:2]
-        if 'bboxes' not in params:  # less likely, this class is for use with bboxes.
+        if len(params['bboxes']) == 0:  # less likely, this class is for use with bboxes.
```
albu: if `bboxes` not in params this code will throw 'KeyError' so we can check both conditions

tests/conftest.py
```diff
@@ -41,7 +41,7 @@ def mask():
 
 @pytest.fixture
 def bboxes():
-    return [[15, 12, 75, 30, 1], [55, 25, 90, 90, 2]]
+    return [[0.15, 0.12, 0.75, 0.30, 1], [0.55, 0.25, 0.90, 0.90, 2]]
```
albu: why did you changed this?

tests/test_transforms.py
```diff
@@ -99,7 +99,6 @@ def test_elastic_transform_interpolation(monkeypatch, interpolation):
     [A.ShiftScaleRotate, {'rotate_limit': 45}],
     [A.RandomScale, {'scale_limit': 0.5}],
     [A.RandomSizedCrop, {'min_max_height': (80, 90), 'height': 100, 'width': 100}],
-    [A.RandomSizedBBoxSafeCrop, {'height': 100, 'width': 100}],
```
Dipet: Why did you remove this tests?

albumentations/augmentations/transforms.py
Jae-Hyuck: https://github.com/albu/albumentations/blob/d85bab59eb8ccb0a2fec86750f94173e18e86395/albumentations/core/transforms_interface.py#L56-L58

if `bboxes` is not given, line 57 will throw 'KeyError', before enter the `get_params_dependent_on_targets` method.

tests/conftest.py
Jae-Hyuck: Interger bboxes throw errors while checking **test_serialization.py**.
RandomSizedBBoxSafeCrop only allows float bboxes now.
In my understanding, float bboxes is standard for this library, so I changed it.
Am I wrong?

tests/test_transforms.py
Jae-Hyuck: After this commit, `bboxes` parameter is essential to run `RandomSizedBBoxSafeCrop`.  So, I tried to add any `bboxes` parameter to the test. But there are some augmentations not supporting `bboxes` in the test, so I just removed `RandomSizedBBoxSafeCrop`. If this is not a good way, please give me an advice.

Dipet: I see, this tests work only with images or images and masks. Add RandomSizedBBoxSafeCrop to this tests was a mistake.

tests/conftest.py
Dipet: @albu He is right, tests functions does not call Compose transform, so bboxes  in test are not normalized. 

I think, need rename this function to `normalized_bboxes` and return back original function. Then in `test_serialization` rename all `bboxes` test functions params to `normalized_bboxes`, except `test_transform_pipeline_serialization_with_bboxes`

albumentations/augmentations/transforms.py
Dipet: @albu This transform works only with bboxes, so if we get `KeyError` it will be right behavior

tests/conftest.py
albu: or `albumentations_boxes`.

albumentations/augmentations/transforms.py
albu: Then we should assert that this transform does not work without bboxes.

Jae-Hyuck: @albu Done.

tests/conftest.py
Jae-Hyuck: @Dipet Fixed.

mirko-m:
Thanks for fixing this. It seems to me that the code in `get_params_dependent_on_targets` assumes that the bboxes are in the `albumentations` format. Is that intentional? I got an unexpected result, because I passed bboxes in a different format, but there is nothing to warn the user against doing that.

Maybe this could be fixed with a call to `convert_bboxes_to_albumentations` ?

PR #367, opened 2019-09-20
Add isort to Travis CI

creafz:
- Sort imports with isort (apply isort only to albumentations and not to the helper code like benchmark/, notebooks/, etc.)
- Add isort check to Travis-CI
- Fix a cyclic import between albumentations/__init__.py and  albumentations/core/serialization.py. (Importing the __version__ variable from __init__.py  fails after applying isort to this file).

I don't like the default sorting from isort, so I have changed a few config params here - https://github.com/albu/albumentations/compare/add_isort_to_travis_ci?expand=1#diff-380c6a8ebbbce17d55d50ef17d3cf906R6-R10. A full list of isort settings is located here - https://github.com/timothycrosley/isort/wiki/isort-Settings.


ternaus:
Why would you prefer to skip helper code like benchmark, etc?

tests/test_core.py
```diff
@@ -4,11 +4,31 @@
...
+    ImageOnlyTransform,
+    to_tuple,
+)
+
+from .compat import (
+    MagicMock,
+    Mock,
+    call,
+    mock,
+)
```
albu: I'm not really sure that I like this style more then plain lines. it's 22 lines of code instead of 3

tests/test_augmentations.py
```diff
@@ -5,17 +5,61 @@
...
+    RandomSunFlare,
+    RGBShift,
+    Rotate,
+    ShiftScaleRotate,
+    Solarize,
+    ToFloat,
+    ToGray,
+    Transpose,
+    VerticalFlip,
+)
```
albu: 50 lines!!

ternaus:
Should we close it as `isort`  and `black` do not work together yet.

creafz:
> Why would you prefer to skip helper code like benchmark, etc?

You are right. I think it is better to apply isort to those files as well if we decide to use it someday.

> Should we close it as isort and black do not work together yet.

Yep, closing this pull request for now.

PR #369, opened 2019-09-20
Check black

ternaus:


ternaus:
@BloodAxe It is not just black, it is more general config, defined in https://www.python.org/dev/peps/pep-0518/

PR #379, opened 2019-09-25
SIIM-ACR Pneumothorax Segmentation

alimbekovKZ:
33 place solo model in SIIM-ACR Pneumothorax Segmentation


ternaus:
Thank you for your PR!

It looks like you need to add one empty line between names and `code, solution` to make formatting consistent.

Issue #381, opened 2019-09-25
Propose to save used parameters and was transform applied or not as a parameter


IlyaOvodov:
Sometimes it is nesessary to know was a particular transform applied or not and with what randomly chosen parameters.
For instance, flips can be an acceptable transform for classification or detection task provided that labels would be changed in a proper way (say, "left turn" road sign should be changed to "right turn" if hflip was applied).
I dont thing it is worth handling all such cases by the Albumentations library itself and sometimes it is better to leave it for user side.

IlyaOvodov:
Properties `was_applied` and `params` were added to all transforms to keep an information about was this transform applied and with what parameters.
Usage examle:
```python
aug_res = self.albumentations(image=img, bboxes=rects)
aug_bboxes = aug_res['bboxes']
for t in self.albumentations.transforms:
    if isinstance(t, T.VerticalFlip) and t.was_applied:
        aug_bboxes = [vflip_rect_label(b) for b in aug_bboxes]
    if isinstance(t, T.HorizontalFlip) and t.was_applied:
        aug_bboxes = [hflip_rect_label(b) for b in aug_bboxes]
```

IlyaOvodov:
See discussion under PR #382 
Probably required functionality will be available after merge of PR #350

albu:
I've added an example how new replay mode works. https://github.com/albu/albumentations/blob/deterministic_poc/notebooks/replay.ipynb 

As for 
> For instance, flips can be an acceptable transform for classification or detection task provided that labels would be changed in a proper way (say, "left turn" road sign should be changed to "right turn" if hflip was applied).
You can use `get_params_dependent_on_targets` and write your own flip which flips only applicable signs.

Something like this:
```
class SignHorFlip(HorizontalFlip):
    def __init__(self, ignored_labels=[], always_apply=False, p=1.):
        super(SignHorFlip, self).__init__(always_apply=always_apply, p=p)
        self.ignored_labels = ignored_labels

    def get_params_dependent_on_targets(self, params):
        label = params['label']
        if label in self.ignored_labels:
            return {'apply': False}
        return {'apply': True}

    def targets_as_params(self):
        return ['label']

    def apply(self, img, apply=True):
        if not apply:
            return img
        if img.ndim == 3 and img.shape[2] > 1 and img.dtype == np.uint8:
            return F.hflip_cv2(img)
        else:
            return F.hflip(img)
```

Hope that helps

albu:
```
[...]

PR #382, opened 2019-09-25
#381 implementation. Properties `was_applied` and `params`.

IlyaOvodov:
properties `was_applied` and `params` were added to all transforms to keep an information about was this transform applied and with what parameters.

### DIFF
--- a/albumentations/core/transforms_interface.py
+++ b/albumentations/core/transforms_interface.py
@@ -48,26 +48,30 @@ class BasicTransform(object):
         self.p = p
         self.always_apply = always_apply
         self._additional_targets = {}
+        self.params = {}
+        self.was_applied = False
 
     def __call__(self, force_apply=False, **kwargs):
+        self.was_applied = False
         if (random.random() < self.p) or self.always_apply or force_apply:
-            params = self.get_params()
-            params = self.update_params(params, **kwargs)
+            self.params = self.get_params()
+            self.params = self.update_params(self.params, **kwargs)
             if self.targets_as_params:
                 assert all(key in kwargs for key in self.targets_as_params), "{} requires {}".format(
                     self.__class__.__name__, self.targets_as_params
                 )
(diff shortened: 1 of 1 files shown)


IlyaOvodov:
I've looked at #350. I don't see in what way it is the same. I understand it provides a possibility to replay transforms with the same parameters. It is not what I am going to get. See description in #381. 

Multiprocess execution really can be a problem as far as in this case the same transform can be run with different parameters in different threads, so applied parameters are not a property of a Compose object but a property of a specific (parallel) call to it. Also, my approach can be incorrect if augmentation contains OneOf or smth. similar. Better approach is to store fired transforms and their parameters and return it in an augmentation results dict. This approach is partly implemented in #350, but it is invoked only in a replay mode. Why do you need a special ReplayCompose class there? Why not to store parameters of invoked transforms always and not to return them with other results?

So I agree that this PR is not a good idea and should be declined, I will return to this problem when #350 will be merged.

albu:
Don't worry, this PR is a really good idea, we thought about it for a long time.

What's done in #350 :
- https://github.com/albu/albumentations/pull/350/files#diff-2f0d114a776157b2c551dc907cec3253R67 writes transformation params to results dict.
- https://github.com/albu/albumentations/pull/350/files#diff-c3271807771eaa1e34cfc16dd8660db5R342 fills `applied` field. 

So you receive all applied augmentations' params and flag (is_applied) for every image.
We have to use another class `ReplayCompose` because if you pass dict to `ToTensor` and `DataLoader` in pytorch - all the dimensions of all args have to be the same. I'm not sure if dimensions of resulting dicts will be the same and how it will interact with torch.

You can just replace normal `Compose` by `ReplayCompose`. The only limitation right now - it cannot be serialized.

I'll add example of usage to #350 to make it more clear.

If you have ideas on how to improve #350 - you are really welcomed.

Cheers

Issue #142, opened 2018-12-07
About RandomSizedCrop implementation


lyakaap:
When using RandomSizedCrop(), is it better to use padding if a suggested crop region is bigger than the original image size, rather than raise ValueError?
For example, the augmentation method used for achieving cityscapes good result uses padding.
https://github.com/PkuRainBow/OCNet/blob/8184d9d55ec474c7e22278256f1db8b3d7ebce88/dataset/cityscapes.py#L121

So I suggest changing this part;
https://github.com/albu/albumentations/blob/8626026910d6b0236bccd09af276ca9432af5d22/albumentations/augmentations/functional.py#L207-L221

to 
```
def random_crop(img, crop_height, crop_width, h_start, w_start, pad_value=(0.0,)):
    height, width = img.shape[:2]
    if height < crop_height or width < crop_width:
        pad_h = max(crop_height - height, 0)
        pad_w = max(crop_width - width, 0)
        img = cv2.copyMakeBorder(img, 0, pad_h, 0,
                                 pad_w, cv2.BORDER_CONSTANT,
                                 value=pad_value)  # if gt label, pad_value should be ignore index.
    x1, y1, x2, y2 = get_random_crop_coords(height, width, crop_height, crop_width, h_start, w_start)
    img = img[y1:y2, x1:x2]
    return img
```

lyakaap:
Thanks for quick replying!
I'm not sure if this issue should be closed until the change will be merged, so I remain it opened.

lyakaap:
Sorry, the above code I suggested was buggy. I fixed.

vfdev-5:
Any updates on this ?

Dipet:
Unclear and unexpected padding may be a source of bugs.
If you need this type of transform you can create [Lambda](https://albumentations.readthedocs.io/en/latest/api/augmentations.html#albumentations.augmentations.transforms.Lambda) transform with this changes.

PR #393, opened 2019-10-02
Pad image for large random sized crop.

Dipet:
#142


ternaus:
I do not support these changes. Unclear and unexpected padding may be a source of bugs.

For Crops, we have an Exception. We may update it and recommend PadIfNeeded to be used before the crop function to make it more explicit.

For RandomSizedCrop we should probably make crop smaller if it fits into the proposed range of the transformation and throw an Exception otherwise.

albu:
we need to discuss it more broader

Dipet:
To be honest I agree with @ternaus. I think, If we need this type of transform I see 2 options how we can do this:
1. Create new transform like RandomSizedCropWithPad;
2. Add flag `pad_if_need` or `accept_large_crop` to transforms that uses `random_crop`

albu:
yes, both options are good

ternaus:
I would prefer second option.

We already have two `RandomSized` transforms. Would prefer to avoid generating third :)

albu:
"RandomCrop",
    "CenterCrop",
    "RandomSizedCrop",
    "RandomResizedCrop",
    "RandomCropNearBBox",
    "RandomSizedBBoxSafeCrop",

you're too optimistic about 2 :)

Dipet:
Too old, not relevant

Issue #184, opened 2019-02-10
Add keypoint support to Crop, Resize, LongestMaxSize, SmallestMaxSize


ternaus:
To:
 - [x] Crop
 - [x] LongestMaxSize
 - [x] SmallestMaxSize
 - [x] Resize

PR #395, opened 2019-10-02
Keypoints support

Dipet:
Added keypoints support to next transforms:
- Crop
- CropNonEmptyMaskIfExists
- LongestMaxSize
- RandomCropNearBBox
- Resize
- SmallestMaxSize
- Transpose

#184

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -32,6 +32,16 @@ def clipped(func):
     return wrapped_function
 
 
+def angle_to_2pi(angle):
+    if 0 <= angle <= 2 * np.pi:
+        return angle
+
+    if angle < 0:
+        angle += abs(angle // (2 * np.pi) + 1) * 2 * np.pi
+
+    return angle % (2 * np.pi)
+
+
 def preserve_shape(func):
     """
     Preserve shape of the image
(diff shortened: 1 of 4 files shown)


albumentations/augmentations/functional.py
```diff
@@ -32,6 +32,16 @@ def wrapped_function(img, *args, **kwargs):
     return wrapped_function
 
 
+def angle_to_2pi(angle):
```
ternaus: We should have a docstring here. 

Or name to be slightly longer like `angle_to_2pi_range`

albumentations/augmentations/functional.py
```diff
@@ -32,6 +32,16 @@ def wrapped_function(img, *args, **kwargs):
...
+    if 0 <= angle <= 2 * np.pi:
+        return angle
+
+    if angle < 0:
+        angle += abs(angle // (2 * np.pi) + 1) * 2 * np.pi
```
BloodAxe: return angle_to_2pi_range(angle + 2 * np.pi) ?

albumentations/augmentations/functional.py
```diff
@@ -32,6 +32,16 @@ def wrapped_function(img, *args, **kwargs):
...
+
+    return angle % (2 * np.pi)
```
BloodAxe: return angle_to_2pi_range(angle - 2 * np.pi) ?

albumentations/augmentations/functional.py
Dipet: What if angle less then `-2 * np.pi`?

Dipet: I think better use one expression, than recurrent functions. If angle will be very large it will be much faster.

PR #402, opened 2019-10-05
Float random number generator instead of int

apatsekin:
The function params are defined as integers. However, at this step you are converting them to float and then add to int32 image using cv2.add which I guess makes no sense.


apatsekin:
The same applies to RGBShift

albu:
these limits are not strictly int limits. we need to clarify it in docs, thanks.

btw uniform returns float but we cast it to the needed dtype (int or float)

albu:
created issue about docs

PR #404, opened 2019-10-06
Add citations.rst

creafz:


ternaus:
Looks great! 

But I would also add link from Readme to this doc from the section https://github.com/albu/albumentations#citing

creafz:
@ternaus done.

PR #405, opened 2019-10-06
Fix bug in grid_distortion: index out of range

apatsekin:
Current implementation error outs under certain input params:
https://github.com/albu/albumentations/issues/134

Here is an example to reproduce:
```
import albumentations, numpy as np
augmentation = albumentations.Compose([albumentations.GridDistortion(num_steps=21, distort_limit=0.3, interpolation=cv2.INTER_LINEAR, border_mode=cv2.BORDER_CONSTANT, p=1)], p=1)
image = np.random.random((100, 100,3))
image = augmentation(image=image)['image']
```

Proposed change uses `np.split_array` instead to make an approx. equal grid for an input image.

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -1008,35 +1008,23 @@ def grid_distortion(
     """
     height, width = img.shape[:2]
 
-    x_step = width // num_steps
-    xx = np.zeros(width, np.float32)
+    x_split = np.array_split(range(width), num_steps)
     prev = 0
-    for idx, x in enumerate(range(0, width, x_step)):
-        start = x
-        end = x + x_step
-        if end > width:
-            end = width
-            cur = width
-        else:
-            cur = prev + x_step * xsteps[idx]
-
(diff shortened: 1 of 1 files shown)


albumentations/augmentations/functional.py
```diff
@@ -1008,35 +1008,23 @@ def grid_distortion(
...
-        else:
-            cur = prev + x_step * xsteps[idx]
-
-        xx[start:end] = np.linspace(prev, cur, end - start)
+    for i, chunk in enumerate(x_split):
+        cur = prev + len(chunk) * xsteps[i]
+        split = np.linspace(prev, cur, len(chunk))
+        x_split[i] = split
         prev = cur
+    xx = np.concatenate(x_split)
```
Dipet: I think we can create one function for this case, something like `split_axis`. After it this code will look like this:
```
x_split = split_axis(width, num_steps, xsteps)
y_split = split_axis(height, num_steps, ysteps)
```

PR #409, opened 2019-10-08
Update benchmark for 0.4.0

creafz:


Dipet:
Need update Brightness - `albumentations.brightness_contrast_adjust(img, beta=0.5, beta_by_max=True)` - it is new behavior

ternaus:
`birghtness_by_max=True` is the default behavior. I believe it is picked automatically.

Dipet:
benchmark uses functional api, so need set this flag

ternaus:
Also, I do not see transforms:

```
        Posterize(),
        Solarize(),
        Equalize(),
```

PR #412, opened 2019-10-09
preserve_channels flag for ToGray transform

Dipet:
Do not preserve channel is faster and sometimes need 1 channel image


albumentations/augmentations/transforms.py
```diff
@@ -2725,6 +2725,7 @@ class ToGray(ImageOnlyTransform):
     than 127, invert the resulting grayscale image.
 
     Args:
+        preserve_channels (bool): If True transform return 3 channel image, else 1 channel. Default: True.
```
ternaus: I would make `preserve_channels=False` as a default behavior.

Dipet: I did reset commit. I changed my mind again. 😄 
But I do not sure that is right way, because we provide augmentations and I think they should not change images shapes without explicit directions.
But ToGray transform is mean that that transform give to us 1 channel image... I am little bit confused 😕

Dipet:
As I can see Flips works well with Grayscale images.
If we provide Grayscale image into Compose pipeline it is does not converts to 3 channel.
As I know all our transforms can work with grayscale images, except something special, like `ChannelShuffle`

Dipet:
But I do not sure that current PR version is right, because we provide augmentations and I think they should not change images shapes without explicit directions.
I think  we need leave `preserve_channels=True`, so that the default behavior will not change.

albu:
https://github.com/albu/albumentations/issues/290
https://github.com/albu/albumentations/issues/180
I'm not sure if we fixed it.

Need additional testing for all transforms on grayscale images that they work as expected because initially we didn't work on grayscale (and I thought noone made grayscale support)

Dipet:
Need more clearly implementations for support of gray images  for all transforms.

Issue #411, opened 2019-10-09
Normalized keypoints angles


Dipet:
At this time we does not normalize keypoints angles to `[0, 2 * PI]` range.

**Need**
- [ ] Normalize keypoints angle in `convert_keypoint_to_albumentations`
- [ ] Create wrap function to keep keypoints angle in `[0, 2 * PI]` range.
- [ ] Wrap functions, that chage keypoints angle.

**Related files**
* `albumentations/augmentations/keypoints_utils.py`
* `albumentations/augmentations/functional.py`

Sriharsha-hatwar:
Hi @Dipet , would definitely like to work on this, need to understand a bit about `convert_keypoint_from_albumentations` is there any usage example of this.

Dipet:
Hi @Sriharsha-hatwar `convert_keypoint_from_albumentations` is internal function.
 It is used in `Compose` method.
Stacktrace looks like this:
1. We are create transform and call it:
```
transform = albumentations.Compose(transforms)
result = transform(image=image, keypoints=keypoints)
```
2. [`__call__`](https://github.com/albu/albumentations/blob/ff1a17f52f363a3901f2b7a2d720b7745e1c3bc4/albumentations/core/composition.py#L163) method of Compose find and call `convert_keypoint_to_albumentations` in [`preprocess`](https://github.com/albu/albumentations/blob/ff1a17f52f363a3901f2b7a2d720b7745e1c3bc4/albumentations/core/composition.py#L174) method. 
3. After apply transform called [`postprocess`](p.postprocess(data)).
`postprocess` method find [`convert_keypoint_from_albumentations`](https://github.com/albu/albumentations/blob/ff1a17f52f363a3901f2b7a2d720b7745e1c3bc4/albumentations/core/utils.py#L65) and call it.

If you does not understand, write about this, and I will try to explain it better.

Sriharsha-hatwar:
sorry I could not do justice to this. was about to look into this today.

ternaus:
We normalize now

PR #426, opened 2019-10-16
Keypoint angle normalization

Dipet:
1. Normalize keypoint angle to `[0, 2 * PI)` when convert to albumentations format.
2. Wrap function for angle normalization.

Also `angle_to_2pi_range` moved to `keypoint_utils`.

Attention! Current version may return a different angle, compared to original code.
For example, if in previous version `angle = 4 * PI`, after transforms it will be `angle = 2 * PI`.
Current version in this case returns `angle = 0`


Dipet:
#411

albumentations/augmentations/functional.py
```diff
@@ -18,6 +19,15 @@
 }
 
 
+def angle_2pi_range(func):
+    @wraps(func)
+    def wrapped_function(keypoints, *args, **kwargs):
```
BloodAxe: `keypoint`, singular, not plural

albumentations/augmentations/keypoints_utils.py
```diff
@@ -15,6 +16,14 @@
 keypoint_formats = {"xy", "yx", "xya", "xys", "xyas", "xysa"}
 
 
+def angle_to_2pi_range(angle):
+    two_pi = 2 * math.pi
+    if 0 <= angle < two_pi:
```
BloodAxe: actually, with modulo operator we don't need the check at all and always return `angle % two_pi`

Issue #427, opened 2019-10-16
Mark deprecated transforms in README and docs


Dipet:
1. Update docstrings for deprecated transforms.
**Related files:** `albumentations/augmentations/transforms.py`
2. Mark deprecated functions in README using ` [Depricated]` suffix.
**Related files:** `tools/make_transforms_docs.py`

Soniyanayak51:
Hi, I would like to work on this. How do you want the docstrings to be updated?

Dipet:
See #596

ternaus:
Fixed

PR #431, opened 2019-10-17
Marked deprecated transforms in README and docstrings.

Soniyanayak51:
Fixes issue #427

Added [Deprecated] tags in README and modified the docstrings for deprecated transforms.


Soniyanayak51:
@Dipet please review.

Dipet:
Thank you for PR.
I think we also need add to docstring a new  tansform name, that was replaced deprecated. Like `Deprecated. Use ImageCompression`.
Also, because we create links to transforms in README automatically (see https://github.com/albu/albumentations/blob/e0b92cfba2b40483e6cbac63e8412c20d4cc1ab1/tools/make_transforms_docs.py#L51), we need create tags for deprecated transforms automatically to. You can do this checks that docstring for transform starts with `Deprecated`. You can do this using `inspect` module.

Dipet:
Will be fixed in #596

Issue #436, opened 2019-10-21
Unexpected behavior of `max_holes` in CoarseDropout


t-hanya:
## 🐛 Bug

CoarseDropout may produce `max_holes + 1` holes. 

## To Reproduce

<img width="932" alt="スクリーンショット 2019-10-21 21 00 58" src="https://user-images.githubusercontent.com/18238865/67203592-73406980-f446-11e9-865c-88337a34899c.png">

In CoarseDropout code, the number of holes is determined by `random.randint(self.min_holes, self.max_holes + 1)`, but `random.randint(a, b)` function returns random integer in range[a, b], including both end points.

```python
>>> import random
>>> for _ in range(10):
...     print(random.randint(0, 1))
... 
0
0
0
0
0
1
0
1
0
1
>>> 
```

## Expected behavior

For option name `max_holes`, we expect CoarseDropout function produces at most `max_holes` holes.

## Environment

 - Albumentations version (e.g., 0.1.8): 0.4.1
 - Python version (e.g., 3.7): 3.6.5
 - OS (e.g., Linux): macOS
 - How you installed albumentations (`conda`, `pip`, source): pip

ternaus:
Fixed

PR #437, opened 2019-10-21
Fix max_holes behavior in CoarseDropout

t-hanya:
Current implementation of CoarseDropout may produce `max_holes + 1` holes. I modified the logic to determine the number of holes.
issue: #436

Could you check the changes and consider to merge it?
Thank you :)

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -1550,7 +1550,7 @@ class CoarseDropout(ImageOnlyTransform):
         height, width = img.shape[:2]
 
         holes = []
-        for _n in range(random.randint(self.min_holes, self.max_holes + 1)):
+        for _n in range(random.randint(self.min_holes, self.max_holes)):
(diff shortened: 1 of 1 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -1550,7 +1550,7 @@ def get_params_dependent_on_targets(self, params):
         height, width = img.shape[:2]
 
         holes = []
-        for _n in range(random.randint(self.min_holes, self.max_holes + 1)):
+        for _n in range(random.randint(self.min_holes, self.max_holes)):
             hole_height = random.randint(self.min_height, self.max_height + 1)
             hole_width = random.randint(self.min_width, self.max_width + 1)
```
Dipet: I think we need fix this lines to.

t-hanya: Thanks for your review. I'm sorry I missed them. I will fix these lines.

t-hanya:
I fixed the `random.randint` range for `hole_width` and `hole_height`.

PR #440, opened 2019-10-22
pytorch module visibility when import albumentations

Dipet:


Dipet:
Pytorch must be installed for this changes, but it is not in library requirements.

PR #441, opened 2019-10-23
fix inaccucaries

ternaus:


albumentations/augmentations/functional.py
```diff
@@ -1526,8 +1526,6 @@ def bbox_rot90(bbox, factor, rows, cols):
         bbox = 1 - x_max, 1 - y_max, 1 - x_min, 1 - y_min
     elif factor == 3:
         bbox = 1 - y_max, x_min, 1 - y_min, x_max
-    else:
```
albu: ?

ternaus: This `else` does not do anything and in similar parts of the code we do not have it. -> less code + more consistency

PR #449, opened 2019-10-27
Better benchmarks

creafz:
- Force all benchmarked transformations to produce a contiguous NumPy array with the np.uint8 data type.
- Force NumPy to use a single thread in all benchmarks.
- Add Dockerfile to build an image with benchmarking scripts to improve reproducibility.
- Add an option to benchmark script for producing a markdown table that could be pasted directly into [README.md#benchmarking-results](https://github.com/albu/albumentations#benchmarking-results).
- Benchmark torchvision only with the Pillow-SIMD backend.


Dipet:
Maybe better to use pyenv, then docker? Because pyenv is lighter then docker.

creafz:
Docker is also a very lightweight abstraction layer (when it runs on a Linux machine). Performance penalties when using Docker come mainly from network and storage virtualisation. Since the benchmark doesn't use any networking features and keeps all the images in memory, the cost of containerisation should be close to zero.

On the other side, using Docker helps to keep a more reproducible environment. With virtualenv there are many more moving parts like Python version and system libraries versions, so it is harder to obtain consistent benchmarking results.

PR #450, opened 2019-10-27
MaskDropout

BloodAxe:
New augmentation `A.MaskDropout`, inspired by https://www.kaggle.com/c/severstal-steel-defect-detection/discussion/114254#latest-659361.

It takes random object from mask and zero-out (or with any other value from c'tor args) image and mask. It can drop at most N objects and instead naive zeroing can also do inpaining on image target.


Dipet:
Are we have `skimage` in library dependency?

BloodAxe:
> skimage

Probably not. I will try to use `cv2.connectedComponents` instead

BloodAxe:
@Dipet I have reverted `cv2.connectedComponents` -> `label` from `scikit-image`.
Reason:
1) We already have scikit-image (`import scimage`) in requirements
2) OpenCV implementation supports only uint8 masks

albumentations/augmentations/transforms.py
```diff
@@ -2970,3 +2972,86 @@ def targets_as_params(self):
...
+
+    Inspired by https://www.kaggle.com/c/severstal-steel-defect-detection/discussion/114254
+    """
+
+    def __init__(self, max_objects=1, image_fill_value=0, mask_fill_value=0, always_apply=False, p=0.5):
+        """
+        Args:
+            max_objects: Maximum number of labels that can be zeroed out. Can be tuple, in this case it's [min, max]
+            image_fill_value: Fill value to use when filling image.
+                Can be 'inpaint' to apply inpaining (works only  for 3-chahnel images)
```
khornlund: typo "channel"

PR #460, opened 2019-11-01
Fix bbox validations

Dipet:


albumentations/augmentations/bbox_utils.py
```diff
@@ -321,10 +321,10 @@ def convert_bboxes_from_albumentations(bboxes, target_format, rows, cols, check_
     return [convert_bbox_from_albumentations(bbox, target_format, rows, cols, check_validity) for bbox in bboxes]
 
 
-def check_bbox(bbox):
+def check_bbox(bbox, eps=1e-6):
     """Check if bbox boundaries are in range 0, 1 and minimums are lesser then maximums"""
     for name, value in zip(["x_min", "y_min", "x_max", "y_max"], bbox[:4]):
-        if not 0 <= value <= 1:
+        if not 0 - eps <= value <= 1 + eps:
```
BloodAxe: Is it really required? Call it sixth sense, but I'm afraid this may have unexpected consequences.

maruschin: With epsilon, there is no need equality check:
```
if not 0 - eps < value < 1 + eps:
```

maruschin: And checking floating point value for equality is not right. There will always be errors.

Dipet: @BloodAxe yes, in some cases we maybe can get undefined behavior. I changed fix for rounding error using np.round.

kidkych:
Any idea when this fix will be merged into master? Currently using the albumentations package in a project where I need to work with YOLO style bboxes.

kidkych:
Cheers! Thanks so much

PR #463, opened 2019-11-06
To Sepia transformation

IliaLarchenko:
Hi, I have just created ToSepia transformation (as part of the library exploration). It applies the Sepia filter to the image. I have not found a similar functional in the library, so maybe it will be useful to pull it.
I have also added a linear_transformation_rgb function that applies the linear transformation to the RGB channels of an image (actually, it is the same as 1x1 conv with 3 out_channels + clipping). I think it can be useful to other sepia-like transformations (if they will be added), that use linear combinations of different channels to produce the result.


ternaus:
Could you please add a similar page to the documentation? https://albumentations.readthedocs.io/en/latest/augs_overview/image_only/image_only.html#multiplicativenoise

albumentations/augmentations/functional.py
```diff
@@ -620,6 +620,16 @@ def shift_rgb(img, r_shift, g_shift, b_shift):
     return _shift_rgb_non_uint8(img, r_shift, g_shift, b_shift)
 
 
+@clipped
+def linear_transformation_rgb(img, transformation_matrix, shifts=None):
+    result_img = cv2.transform(img, transformation_matrix.astype(np.float32))
```
Dipet: You use there shifts, but in ToSepia class you do not use it.
I think, it is not needed, because we have RGBShift transform and this pipeline can do similar thinks:
```
A.Compose([
    A.Sepia(),
    A.RGBShift()
])
```

IliaLarchenko: You are right. `shifts`  is not needed for sepia and can be deleted for now, but I have added it because I wanted to have full linear transformation function in channel space (y = Ax + b, instead of separate y = Ax and y = x + b) that can be used somewhere else.

Your approach using Compose - is ok, I have thought about it, but it is not 100% precise because of `@clipped` - if I understand it correctly it will clip everything before RGB shift. So there can be some mistakes.

IliaLarchenko:
> Could you please add a similar page to the documentation? https://albumentations.readthedocs.io/en/latest/augs_overview/image_only/image_only.html#multiplicativenoise

Yes, I will.

But I have a couple of clarifying questions:
1. ToSepia is similar to ToGray - so there is no randomness or variability. It just a fixed transformation. The example page will have only two images - original and transformed. Is it ok? Or should I add a "level of sepianess" variable and add some randomness to it?
2. Do you have a code for the generation of such pages (at least partial), or it should be done manually?

ternaus:
1. Original + transformed is ok.
2. We do not have a code to generate such a page. But it does not look like a lot of manual work.

PR #466, opened 2019-11-09
sovle KeyError

junedgar:
if image's type is float64 may lead to keyError  bug.
```return clip(func(img, *args, **kwargs), dtype, maxval)
  File "....../site-packages/albumentations/augmentations/functional.py", line 1271, in _brightness_contrast_adjust_non_uint
    max_value = MAX_VALUES_BY_DTYPE[dtype]
KeyError: dtype('float64')
```

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -1262,8 +1262,8 @@ def gauss_noise(image, gauss):
 
 @clipped
 def _brightness_contrast_adjust_non_uint(img, alpha=1, beta=0, beta_by_max=False):
-    dtype = img.dtype
     img = img.astype("float32")
+    dtype = img.dtype
 
     if alpha != 1:
         img *= alpha


albumentations/augmentations/functional.py
```diff
@@ -1262,8 +1262,8 @@ def gauss_noise(image, gauss):
 
 @clipped
 def _brightness_contrast_adjust_non_uint(img, alpha=1, beta=0, beta_by_max=False):
-    dtype = img.dtype
     img = img.astype("float32")
+    dtype = img.dtype
```
junedgar: get the right img type

junedgar:
solve the dtype problem

albumentations/augmentations/functional.py
```diff
@@ -1262,8 +1262,8 @@ def gauss_noise(image, gauss):
 
 @clipped
 def _brightness_contrast_adjust_non_uint(img, alpha=1, beta=0, beta_by_max=False):
-    dtype = img.dtype
```
junedgar: get  the  img's type after the operation of img 's type transformed.

Dipet:
Library does not supports `float64` images, only `float32`. At this transform we need original image type for normalization. With your changes we will get incorrect results for `uint8` images, because we will get always `MAX_VALUES_BY_DTYPE[dtype] == 1`, but we need at this case `MAX_VALUES_BY_DTYPE[dtype] == 255`

PR #469, opened 2019-11-14
Augmentations examples

IliaLarchenko:
I have added to docs several examples of augmentations (the simplest ones actually).
Will probably add more examples of the more complex transformations later.


docs/augs_overview/image_only/image_only.rst
```diff
@@ -29,3 +29,171 @@ API link: :class:`~albumentations.augmentations.transforms.ToSepia`
...
+-------------------
+
+
+API link: :class:`~albumentations.augmentations.transforms.Blur`
+
+
+1. Original image
+2. :code:`Blur(blur_limit=7, p=1.0)`
+3. :code:`Blur(blur_limit=15, p=1.0)`
+4. :code:`Blur(blur_limit=50, p=1.0)`
```
albu: i don't think blur with kernel >~21 makes any sence

ternaus: It depends on the use case. I would say if the kernel is smaller than image => good enough.

albu: Do you know how blur works?

PR #471, opened 2019-11-15
Tool for examples generation

IliaLarchenko:
1. Added the original parrot image
2. Added my cheap and dirty notebook for the generation of examples (maybe something will bу useful)
3. Added a couple of new examples, just created them to test the notebook


Dipet:
Thanks for PR. We will look to your code and possible create a tool in albumentaions_example for that purposes.

PR #483, opened 2019-12-01
Add ShotNoise augmentation

Erlemar:
Adding ShotNoise augmentation from "Benchmarking Neural Network Robustness to Common Corruptions and Perturbations"

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -1964,3 +1964,9 @@ def fancy_pca(img, alpha=0.1):
     orig_img = orig_img.astype(np.uint8)
 
     return orig_img
+
+
+@clipped
+def shot_noise(image, shotnoise):
+
+    return shotnoise
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -41,6 +41,7 @@ __all__ = [
     "MedianBlur",
     "GaussianBlur",
     "GaussNoise",
+    "ShotNoise",
     "CLAHE",
     "ChannelShuffle",
     "InvertImg",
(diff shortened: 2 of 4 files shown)


ternaus:
You need to add this transform to tests.

ternaus:
Just add the name of the transform to all lists like: 

https://github.com/albumentations-team/albumentations/blob/master/tests/test_transforms.py#L244

albumentations/augmentations/transforms.py
```diff
@@ -3082,3 +3083,44 @@ def apply_to_mask(self, img, dropout_mask=None, **params):
...
+
+    Reference:
+    |  https://arxiv.org/abs/1903.12261
+    |  https://github.com/hendrycks/robustness/blob/master/ImageNet-C/create_c/make_imagenet_c.py
+    """
+
+    def __init__(self, noise_severity=12, always_apply=False, p=0.5):
+        super(ShotNoise, self).__init__(always_apply, p)
+
+        assert noise_severity >= 3
```
ternaus: Assertions in the code is not the best practice.

Could you please change to:

```
if noise_severity < 3:
    raise ValueError("Noise Severity should be more or equal to 3, but we got {}".format(noise_severity))
```

albumentations/augmentations/transforms.py
```diff
@@ -3082,3 +3083,44 @@ def apply_to_mask(self, img, dropout_mask=None, **params):
...
+
+        self.noise_severity = noise_severity
+
+    def apply(self, img, shotnoise=None, **params):
+        return F.shot_noise(img, shotnoise=shotnoise)
+
+    def get_params_dependent_on_targets(self, params):
+        image = params["image"]
+        image = np.array(image) / 255.0
```
ternaus: How will this transform work for non-uint8 images? Should it be another factor here that depends on the dtype?

albumentations/augmentations/transforms.py
Erlemar: Changed

albumentations/augmentations/transforms.py
Erlemar: Fixed it

albumentations/augmentations/functional.py
```diff
@@ -1964,3 +1964,9 @@ def fancy_pca(img, alpha=0.1):
     orig_img = orig_img.astype(np.uint8)
 
     return orig_img
+
+
+@clipped
+def shot_noise(image, shotnoise):
+
+    return shotnoise
```
Dipet: Are you sure that we are need there empty function? I think is better call `clipped` function in `get_params_dependent_on_targets`

Erlemar: I thought it is necessary to have a function in "functional" script. I can do as you are suggesting.

BloodAxe:
There is already this kind of noise (based on poisson process) called `A.ISONoise`. To me it does not make sense to have duplicate.

Erlemar:
It seems that this augmentation is duplicate

PR #490, opened 2019-12-08
Add GlassBlur augmentation

Erlemar:
Add GlassBlur augmentation

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -1968,17 +1968,24 @@ def fancy_pca(img, alpha=0.1):
 
 
 @jit(nopython=True)
-def _shuffle(x, i, h, w, md):
+def _shuffle(x, h, w, md, random_state):
+    np.random.seed(random_state)
     dx, dy = np.random.randint(-md, md, size=(2,))
     h_prime, w_prime = h + dy, w + dx
     x[h, w], x[h_prime, w_prime] = x[h_prime, w_prime], x[h, w]
+
     return x
 
 
 @clipped
-def glass_blur(img, sigma, max_delta, iterations, mode):
+def glass_blur(img, sigma, max_delta, iterations, random_state, mode):
     coef = MAX_VALUES_BY_DTYPE[img.dtype]
     x = np.uint8(cv2.GaussianBlur(np.array(img) / coef, sigmaX=sigma, ksize=(0, 0)) * coef)
+
+    if random_state is None:
+        state = np.random.RandomState(1234)
+    else:
+        state = np.random.RandomState(random_state)
 
     if mode == "fast":
 
@@ -1988,8 +1995,8 @@ def glass_blur(img, sigma, max_delta, iterations, mode):
         w = np.repeat(ws, hs.shape[0])
 
         for i in range(iterations):
-            dy = np.random.randint(-max_delta, max_delta, size=(h.shape[0],))
-            dx = np.random.randint(-max_delta, max_delta, size=(h.shape[0],))
+            dy = state.randint(-max_delta, max_delta, size=(h.shape[0],))
+            dx = state.randint(-max_delta, max_delta, size=(h.shape[0],))
             x[h, w], x[h + dy, w + dx] = x[h + dy, w + dx], x[h, w]
 
     elif mode == "exact":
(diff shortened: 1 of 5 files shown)


taxicoo:
The code can be rewritten with 1 cycle instead of using product with 3 cycles.
I've tested both approaches and 1 cycle approach is ~20x faster. See the code below (need to be tested)
`
hs = np.arange(img.shape[0]-max_delta, max_delta, -1)
ws = np.arange(img.shape[1]-max_delta, max_delta, -1)
h = np.tile(hs, ws.shape[0])
w = np.repeat(ws, hs.shape[0])

dy = np.random.randint(-max_delta, max_delta, size=(h.shape[0],))
dx = np.random.randint(-max_delta, max_delta, size=(h.shape[0],))

for i in range(iterations):
    x[h, w] = x[h+dy, w+dx]
`

albumentations/augmentations/functional.py
```diff
@@ -1964,3 +1965,27 @@ def fancy_pca(img, alpha=0.1):
...
+    coef = MAX_VALUES_BY_DTYPE[img.dtype]
+    x = np.uint8(cv2.GaussianBlur(np.array(img) / coef, sigmaX=sigma, ksize=(0, 0)) * coef)
+
+    # locally shuffle pixels
+    for i, h, w in product(
+        range(iterations),
+        range(img.shape[0] - max_delta, max_delta, -1),
+        range(img.shape[1] - max_delta, max_delta, -1),
+    ):
+        x = _shuffle(x, i, h, w, max_delta)
```
Dipet: If you use numba, please implemet `for` exprasion inside numba function, else i don`t see why there need numba.

ternaus:
@Erlemar Could you please add a test for `test_additional_targets_for_image_only` transform ?

https://github.com/albumentations-team/albumentations/blob/c83ee1dabdcf20ccb260700a2ba630c64637b900/tests/test_transforms.py#L283

?

ternaus:
@Erlemar Removing the test that does not pass is not really an option.

albumentations/augmentations/functional.py
Erlemar: In the latest version I implemented both versions, because the original and the faster numpy version give a bit different result.
So there is a "fast" implementation with the code by taxicoo and an "exact" implementation with the code in numba.
I tried to wrap numba over the loop, but the speed decreased. Having numba over the `_shuffle` function has a huge increase over not having it.

albumentations/augmentations/functional.py
```diff
@@ -1964,3 +1965,46 @@ def fancy_pca(img, alpha=0.1):
     orig_img = orig_img.astype(np.uint8)
 
     return orig_img
+
+
+@jit(nopython=True)
+def _shuffle(x, h, w, md, random_state):
+    np.random.seed(random_state)
+    dx, dy = np.random.randint(-md, md, size=(2,))
```
albu: this is against the spirit of library. all the random numbers must be instantiated on `get_params` stage

Erlemar: The problem is that the following code:
```
random_state = np.random.RandomState(random_state)
random_state.randint()
```
doesn't work in numba. As a result I set `random_state` in `get_params` and then pass the value to `_shuffle` without possibility to change it.

albu: please refer to other transforms. the idea is that all the randomness happens in class instead of function. function only applies fixed transformation from parameters generated in class. https://github.com/albumentations-team/albumentations/blob/master/albumentations/augmentations/transforms.py#L1474

Erlemar: I saw a similar approach in elastic transform: https://github.com/albumentations-team/albumentations/blob/master/albumentations/augmentations/functional.py#L1144 and iso_noise: https://github.com/albumentations-team/albumentations/blob/master/albumentations/augmentations/functional.py#L1342
And thought it is okay.

If it is not acceptable, I'll try to fix it.

BloodAxe: Agree with @albu, it's better to separate any random sampling from the transformation code, so later becomes stateless function without side-effects. @Erlemar, we try to keep this approach everywhere and two augmentation you mentioned will be changed in future to follow this convention. Anyway, thanks for the effort, I know sometimes making the PR to be frustrating ;)

Issue #493, opened 2019-12-19
bbox_params with multiple label_fields go wrong


suishicity:
## To Reproduce

When using Compose with multiple label_fields as follows
```Python
    aug = Compose([
         RGBShift(),
         Flip(), 
    ], bbox_params=BboxParams(format='pascal_voc', label_fields=['labels', 'difficult'], min_visibility=0.05))
```
the program goes wrong. The bug info is
```
  File "C:\Users\morgan\Miniconda3\lib\site-packages\albumentations\core\utils.py", line 110, in remove_label_fields_from_data
    field_values.append(bbox[self.data_length + idx])
IndexError: tuple index out of range
```
In `albumentations\core\utils.py` at line 111, I found that the `self.data_length` would be greater than 4. This might be wrong
```Python
    def add_label_fields_to_data(self, data):
        if self.params.label_fields is None:
            return data
        for data_name in self.data_fields:
            for field in self.params.label_fields:
                data_with_added_field = []
                for d, field_value in zip(data[data_name], data[field]):
                    self.data_length = len(list(d))
                    data_with_added_field.append(list(d) + [field_value])
                data[data_name] = data_with_added_field
        return data
```

## Environment

 - Albumentations version 0.4.3:
 - Python version 3.6:
 - OS Win10;
 - How you installed albumentations  `pip`:
 - Any other relevant information:

IlyaOvodov:
it is not an only problem with this code. See #503.

PR #494, opened 2019-12-20
Fix invalid length with multiple label fields

Dipet:
#493


IlyaOvodov:
Approach with `self.data_length` has a shortcoming that it fails if not only bbox but another data fields are used i.e len(self.data_fields)>1 and these data fields have different data length.
I've fixed the problem that this PR fixes, as well as the above one, as a side effect of fixing #503 in PR #504, and in a more ellegant manner, I suppose.

Dipet:
I think #503 is better decision

Issue #499, opened 2019-12-25
HueSaturationValue: Invalid number of channels in input image 


Dok11:
## To Reproduce

```py
for image_index in range(len(images_1)):
    augmentation = aug(p=0.9)
    data = {'image': images_1[image_index]}
    augmented = augmentation(**data)


def aug(p=0.5):
    return Compose([HueSaturationValue(p=0.3)], p=p)
```

Where `images_1[image_index]` is `ndarray (64, 64, 1)

## Environment

 - Albumentations version: 0.4.3
 - Python version: 3.7.4
 - OS: Win 10 x64
 - How you installed albumentations (`conda`, `pip`, source): pip

## Additional context

Very similar to https://github.com/albumentations-team/albumentations/issues/180

Dipet:
I am cannot reproduce this error.
Please add code for image generation and error traceback.
```
import numpy as np
import albumentations as A


def aug(p=0.5):
    return A.Compose([A.HueSaturationValue(p=0.3)], p=p)


print(A.__version__) # 0.4.3
np.random.seed(0)
images_1 = [np.random.randint(0, 256, [64, 64, 1], dtype=np.uint8)]


for image_index in range(len(images_1)):
    augmentation = aug(p=0.9)
    data = {'image': images_1[image_index]}
    augmented = augmentation(**data)
```

Dok11:
```
  File "C:\Users\User\Anaconda3\envs\DLDM\lib\site-packages\albumentations\core\composition.py", line 176, in __call__
    data = t(force_apply=force_apply, **data)
  File "C:\Users\User\Anaconda3\envs\DLDM\lib\site-packages\albumentations\core\transforms_interface.py", line 87, in __call__
    return self.apply_with_params(params, **kwargs)
  File "C:\Users\User\Anaconda3\envs\DLDM\lib\site-packages\albumentations\core\transforms_interface.py", line 100, in apply_with_params
    res[key] = target_function(arg, **dict(params, **target_dependencies))
  File "C:\Users\User\Anaconda3\envs\DLDM\lib\site-packages\albumentations\augmentations\transforms.py", line 2110, in apply
    return F.shift_hsv(image, hue_shift, sat_shift, val_shift)
  File "C:\Users\User\Anaconda3\envs\DLDM\lib\site-packages\albumentations\augmentations\functional.py", line 387, in shift_hsv
    return _shift_hsv_non_uint8(img, hue_shift, sat_shift, val_shift)
[...]

PR #500, opened 2019-12-25
Fix shift hsv gray

Dipet:
#499

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -381,10 +381,19 @@ def _shift_hsv_non_uint8(img, hue_shift, sat_shift, val_shift):
 
 
 def shift_hsv(img, hue_shift, sat_shift, val_shift):
-    if img.dtype == np.uint8:
-        return _shift_hsv_uint8(img, hue_shift, sat_shift, val_shift)
+    is_gray = is_grayscale_image(img)
+    if is_gray:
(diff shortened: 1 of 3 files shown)


albumentations/augmentations/functional.py
```diff
@@ -381,10 +381,19 @@ def _shift_hsv_non_uint8(img, hue_shift, sat_shift, val_shift):
 
 
 def shift_hsv(img, hue_shift, sat_shift, val_shift):
+    is_gray = is_grayscale_image(img)
+    if is_gray:
+        img = cv2.cvtColor(img, cv2.COLOR_GRAY2RGB)
```
albu: I think it's better to shift only value for grayscale and throw exception in we want to shift other things because hsv shift does not make sence to me

Dipet: > hsv shift does not make sence to me

I agree with you.

But question is what we want from all transforms.

In dataset must be all RGB images, so exception will allow to catch error.
But what if we know, that dataset has few grayscale images and we need shift_hsv transform.

I think is the best solutions is add warning that will be say, that something wrong, but transform will be executed.

albu: agree, so let's just shift V(value) and ignore other and warn that hue and saturation are not applicable to grayscale if they are not zeros.

PR #504, opened 2019-12-31
Fix #503 bbox_params = {'label_fields': []} results in error

### LINKED ISSUE #503: bbox_params = {'label_fields': []} results in error
## 🐛 Bug
Setting `'label_fields'` to `[]` for `bbox_params` when creating `Compose` objects results in empty resulting `bbox` tuples after augmentationas are applied. The error appeared after v.0.3.1, for v.0.3.1 if worker correcly.
The problem is caused that `DataProcessor.data_length` is not initialized at `DataProcessor.add_label_fields_to_data` in this case and remains 0, but is used at `DataProcessor.remove_label_fields_from_data` to cut `data`.
Also I noted that using `DataProcessor.data_length` implies that all `data_fields` have the same data length. An opposite case is probably seldom but will result in error.

## To Reproduce

```code=python
import albumentations as A
import albumentations.augmentations.transforms as T
import numpy as np

augs_list = [T.HorizontalFlip(p=1.)]
alb = A.Compose(augs_list, bbox_params={'format': 'pascal_voc',
        'label_fields': [] # if you comment out this line it will operate properly
})
img = np.zeros((256,512,3), dtype = np.uint8)
rects = [[131, 129, 173, 175, 15], [173, 115, 216, 177, 4],]
aug_res = alb(image=img, bboxes=rects)
print(aug_res['image'].shape, aug_res['bboxes'])
->
(256, 512, 3) [(), ()]
```

## Expected behavior

`(256, 512, 3) [(339.0, 129.0, 381.0, 175.0, 15), (296.0, 115.0, 339.0, 177.0, 4)]` in the above example.

## Environment
 - Albumentations version (master rev. c5807e )
 - Python version (e.g., 3.6)
 - OS (Windows)
 - How you installed albumentations (git clone)

## Additional information
I'll fix it.

IlyaOvodov:


IlyaOvodov:
Also it fixed #494 as a side effect.

PR #511, opened 2020-01-10
Fix grid distortion bug.

tmramalho:
Grid distortion can crash for certain image sizes because `range(0, width, x_step)` may produce a number of steps larger than the size of `xsteps`.

Simple repro, should crash with current master.
```
import numpy as np
import albumentations as ab
transf = ab.GridDistortion(p=1, num_steps=15)
x = np.random.rand(17, 17, 3)
transf(image=x)
```

Using linspace guarantees there are only `num_steps` steps.

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -1054,9 +1054,9 @@ def grid_distortion(
     x_step = width // num_steps
     xx = np.zeros(width, np.float32)
     prev = 0
-    for idx, x in enumerate(range(0, width, x_step)):
-        start = x
-        end = x + x_step
+    for idx, x in enumerate(np.linspace(0, width, num_steps)):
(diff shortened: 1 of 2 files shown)


tmramalho:
Where do you suggest I add the test? There doesn't seem to be a file for augmentation-specific issues. 
test_augmentations.py work?

Dipet:
Yes, you can add it into `test_transforms.py`

tmramalho:
Added tests.

Dipet:
Please reformat code using `black` and `flake8`. You also can use [pre-commit](https://github.com/albumentations-team/albumentations/blob/master/docs/contributing.rst) hook for this purposes.

tmramalho:
Ok formatted the code. Also squashed all commits.

Issue #515, opened 2020-01-22
A CoarseDropout fails with rgb


korabelnikov:
Documentation says nothing about 1-channel input
`    """CoarseDropout of the rectangular regions in the image.

    Args:
        max_holes (int): Maximum number of regions to zero out.
        max_height (int): Maximum height of the hole.
        min_width (int): Maximum width of the hole.
        min_holes (int): Minimum number of regions to zero out. If `None`,
            `min_holes` is be set to `max_holes`. Default: `None`.
        min_height (int): Minimum height of the hole. Default: None. If `None`,
            `min_height` is set to `max_height`. Default: `None`.
        min_width (int): Minimum width of the hole. If `None`, `min_height` is
            set to `max_width`. Default: `None`.
        fill_value (int, float, lisf of int, list of float): value for dropped pixels.

    Targets:
        image

    Image types:
        uint8, float32`

Dipet:
Library use `RGB` images in `HWC` format. I can not reproduce problem.
```
import albumentations as A
import numpy as np

t = A.CoarseDropout()
img_rgb = np.random.randint(0, 256, [100, 100, 3], dtype=np.uint8)
res = t(image=img_rgb)['image']
print(res.shape)
```

PR #516, opened 2020-01-22
Make cutout work with 3-channels images

korabelnikov:
This patch checked inside pytorch pipeline with input shape [3, :, :]

https://github.com/albumentations-team/albumentations/issues/515


Dipet:
Library uses images in standard format for OpenCV: `HWC`
So this changes is wrong.

Dipet:
You can work with RGB images like this:
```
import albumentations as A
import numpy as np


t = A.CoarseDropout()
img_rgb = np.random.randint(0, 256, [100, 100, 3], dtype=np.uint8)
res = t(image=img_rgb)['image']
print(res.shape)
```

PR #520, opened 2020-01-28
GridDropout aug

tatigabru:
added GridDropout augmentation, similar by https://arxiv.org/pdf/2001.04086.pdf

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -79,7 +79,7 @@ __all__ = [
     "MultiplicativeNoise",
     "FancyPCA",
     "MaskDropout",
-    "GridCutout",
+    "GridDropout",
 ]
 
 
@@ -3180,9 +3180,9 @@ class GlassBlur(Blur):
         return ["image"]
 
 
-class GridCutout(ImageOnlyTransform):
+class GridDropout(ImageOnlyTransform):
     """
-    GridCutout, cuts out rectangular regions of an image in a grid fashion.
+    GridDropout, drops out rectangular regions of an image in a grid fashion.
 
         Args:
             ratio (float): the ratio of the mask holes to the unit_size (same for horizental and vertical directions).
@@ -3221,8 +3221,7 @@ class GridCutout(ImageOnlyTransform):
         always_apply: bool = False,
         p: float = 0.5,
     ):
-
-        super(GridCutout, self).__init__(always_apply, p)
+        super(GridDropout, self).__init__(always_apply, p)
         self.ratio = ratio
         self.unit_size_min = unit_size_min
         self.unit_size_max = unit_size_max
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -3245,7 +3245,7 @@ class GridDropout(ImageOnlyTransform):
                 2 <= self.unit_size_min <= self.unit_size_max
             ), "Max unit size should be >= min size, both at least 2 pixels."
             assert self.unit_size_max <= min(height, width), "Grid size limits must be within the shortest image edge."
-            self.unit_width = np.random.randint(self.unit_size_min, self.unit_size_max + 1)
+            self.unit_width = random.randint(self.unit_size_min, self.unit_size_max + 1)
             self.unit_height = self.unit_width
         else:
             # set grid using holes numbers
@@ -3275,11 +3275,11 @@ class GridDropout(ImageOnlyTransform):
         if self.shift_y is None:
             self.shift_y = 0
         if self.random_offset:
-            self.shift_x = np.random.randint(0, self.unit_width)
-            self.shift_y = np.random.randint(0, self.unit_height)
+            self.shift_x = random.randint(0, self.unit_width)
+            self.shift_y = random.randint(0, self.unit_height)
         else:
-            self.shift_x = max(min(0, self.shift_x), self.unit_width)
-            self.shift_y = max(min(0, self.shift_y), self.unit_height)
+            self.shift_x = min(max(0, self.shift_x), self.unit_width)
+            self.shift_y = min(max(0, self.shift_y), self.unit_height)
 
         holes = []
         for i in range(width // self.unit_width + 1):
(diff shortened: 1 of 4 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -3177,3 +3178,132 @@ def get_transform_init_args_names(self):
...
+        unit_size_min: int = None,
+        unit_size_max: int = None,
+        holes_number_x: int = None,
+        holes_number_y: int = None,
+        shift_x: int = 0,
+        shift_y: int = 0,
+        random_offset: bool = False,
+        fill_value: int = 0,
+        always_apply: bool = False,
+        p: float = 0.5,
```
Dipet: Please add `shift_x` and `shift_y` into docstring.

tatigabru: done

Dipet: Place them on different lines, because parsers cant parse this doscstring correctly.

albumentations/augmentations/transforms.py
```diff
@@ -3177,3 +3178,133 @@ def get_transform_init_args_names(self):
...
+                    1 <= self.holes_number_x <= width // 2
+                ), "The hole_number_x must be between 1 and image width//2."
+                self.unit_width = width // self.holes_number_x
+            if self.holes_number_y is None:
+                self.unit_height = max(min(self.unit_width, height), 2)
+            else:
+                assert (
+                    1 <= self.holes_number_y <= height // 2
+                ), "The hole_number_y must be between 1 and image height//2."
+                self.unit_height = height // self.holes_number_y
```
Dipet: Are you sure, that we need save `self.unit_width` and `self.unit_height` in this object scope?
I think we can use them as `unit_width` and `unit_height` without self, because for each image this values may be different. Also look to another variables like `hole_width` and `hole_height`

albumentations/augmentations/transforms.py
```diff
@@ -3177,3 +3178,133 @@ def get_transform_init_args_names(self):
...
+            self.shift_x = 0
+        if self.shift_y is None:
+            self.shift_y = 0
+        if self.random_offset:
+            self.shift_x = random.randint(0, self.unit_width)
+            self.shift_y = random.randint(0, self.unit_height)
+        else:
+            self.shift_x = min(max(0, self.shift_x), self.unit_width)
+            self.shift_y = min(max(0, self.shift_y), self.unit_height)
+
```
Dipet: Here you rewrite values, thats was defined by user. Remove `self` qualifier for new `shift_x` and `shift_y`

albumentations/augmentations/transforms.py
```diff
@@ -3177,3 +3178,133 @@ def get_transform_init_args_names(self):
...
+    def targets_as_params(self):
+        return ["image"]
+
+    def get_transform_init_args_names(self):
+        return (
+            "ratio",
+            "unit_size_min",
+            "unit_size_max",
+            "holes_number_x",
+            "holes_number_x",
```
Dipet: typo

tatigabru: All corrected. Actually, I added holes_number_x and holes_number_y for usability, they used only limits of unit_size_min, unit_size_max in the paper, I can delete them or leave?

Dipet: I think is better to have more customization options, then less

Dipet:
Please look to #521 This PR implements the same things. I think you need to decide with author whats to do with this transformation.

albumentations/augmentations/transforms.py
```diff
@@ -3177,3 +3178,135 @@ def get_transform_init_args_names(self):
...
+                y1 = min(shift_y + unit_height * j, height)
+                x2 = min(x1 + hole_width, width)
+                y2 = min(y1 + hole_height, height)
+                holes.append((x1, y1, x2, y2))
+
+        return {"holes": holes}
+
+    @property
+    def targets_as_params(self):
+        return ["image"]
```
Dipet: I think you do not need this property, because `height` and `width` already in the params with names `rows` and `cols`

tatigabru: did not know that, thank you

albumentations/augmentations/transforms.py
```diff
@@ -3177,3 +3178,135 @@ def get_transform_init_args_names(self):
...
+        self.unit_size_max = unit_size_max
+        self.holes_number_x = holes_number_x
+        self.holes_number_y = holes_number_y
+        self.shift_x = shift_x
+        self.shift_y = shift_y
+        self.random_offset = random_offset
+        self.fill_value = fill_value
+        assert 0 < self.ratio < 1, "ratio must be between 0 and 1."
+
+    def apply(self, image, fill_value=0, holes=[], **params):
```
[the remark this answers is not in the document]
tatigabru: added

albumentations/augmentations/transforms.py
Dipet: Strange that params does not contains this values. Will need to check it out in other transforms, thanks for info.

albumentations/augmentations/transforms.py
```diff
@@ -3177,3 +3178,146 @@ def get_transform_init_args_names(self):
...
+        self.unit_size_min = unit_size_min
+        self.unit_size_max = unit_size_max
+        self.holes_number_x = holes_number_x
+        self.holes_number_y = holes_number_y
+        self.shift_x = shift_x
+        self.shift_y = shift_y
+        self.random_offset = random_offset
+        self.fill_value = fill_value
+        self.mask_fill_value = mask_fill_value
+        assert 0 < self.ratio < 1, "ratio must be between 0 and 1."
```
Dipet: I am not sure, but maybe there need conditions looks like: `0 < self.ratio <= 1`

albumentations/augmentations/transforms.py
```diff
@@ -3177,3 +3178,146 @@ def get_transform_init_args_names(self):
...
+                If 'True', entered shift_x, shift_y are ignored and set randomly. Default: `False`.
+            fill_value (int): value for the dropped pixels. Default = 0
+            mask_fill_value (int): value for the dropped pixels in mask.
+                If `None`, tranformation is not applied to the mask. Default: `None`.
+        Targets:
+            image, mask
+        Image types:
+            uint8, float32
+        References:
+            https://arxiv.org/pdf/2001.04086.pdf7
```
ternaus: https://arxiv.org/abs/2001.04086

PR #521, opened 2020-01-29
Add GridMask augmentation

haqishen:
An implementation of GridMask : https://arxiv.org/abs/2001.04086

demo： https://www.kaggle.com/haqishen/gridmask

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -50,6 +50,7 @@ __all__ = [
     "JpegCompression",
     "ImageCompression",
     "Cutout",
+    "GridMask",
     "CoarseDropout",
     "ToFloat",
     "FromFloat",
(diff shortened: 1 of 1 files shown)


Dipet:
Please add information about transform like described [here](https://github.com/albumentations-team/albumentations#contributing)
And reformat code using black, for this purposes you can use [pre-commit hook](https://github.com/albumentations-team/albumentations/blob/master/docs/contributing.rst#to-create-a-pull-request)

albumentations/augmentations/transforms.py
```diff
@@ -1497,6 +1498,98 @@ def get_transform_init_args_names(self):
...
+        self.rotate = rotate
+        self.mode = mode
+        self.masks = None
+        self.rand_h_max = []
+        self.rand_w_max = []
+
+    def init_masks(self, height, width):
+        if self.masks is None:
+            self.masks = []
+            for n, n_g in enumerate(range(self.num_grid[0], self.num_grid[1] + 1, 1)):
```
Dipet: When images have different shapes this type of caching values is wrong

albumentations/augmentations/transforms.py
```diff
@@ -1497,6 +1498,98 @@ def get_transform_init_args_names(self):
...
+    |  https://arxiv.org/abs/2001.04086
+    |  https://github.com/akuxcw/GridMask
+    """
+
+    def __init__(self, num_grid=3, fill_value=0, rotate=0, mode=0, always_apply=False, p=0.5):
+        super(GridMask, self).__init__(always_apply, p)
+        if isinstance(num_grid, int):
+            num_grid = (num_grid, num_grid)
+        if isinstance(rotate, int):
+            rotate = (-rotate, rotate)
```
Dipet: You can use here our `to_tuple` method

Dipet:
Please look to #520 This PR implements the same things. I think you need to decide with author whats to do with this transformation.

tatigabru:
Hi @haqishen, yes, in #520 PR implemented the same things, but I added more user-friendly parameters. And I did not add rotation to avoid overloading it, as there are other augmentations that cover it (you can rotate the image and then apply grid mask, with the same effect)
If rotation is needed within this augmentation, maybe when I finish my PR you could add your rotation to it, for example? I do not know, what to do for the case of the double-PR

Dipet:
I think is better wait when we will merge #520, then @haqishen can add his improvements like different transform mods and rotations.

haqishen:
I read the code of #520, I think it's better to add my code on #520 as well
Will do it later.
I'm closing this PR.

PR #522, opened 2020-01-31
Pytorch GPU augs

Dipet:
Benchmark results is here: https://github.com/albumentations-team/albumentations/blob/pytorch_augs/albumentations/pytorch/augmentations/README.md


Dipet:
Very large, need decomposition.

PR #523, opened 2020-01-31
Fix wrong varible usage in grid_distortion

steermomo:
'width' is used in both directions.
```
for idx, x in enumerate(np.linspace(0, width, num_steps)):
    ...
for idx, y in enumerate(np.linspace(0, width, num_steps)):
    ...
```
It should be `height`  in the row direction.


steermomo:
A mistake from #511

PR #524, opened 2020-02-02
Fix misprint in probabilities

Datasciensyash:
Found misprint in docs. `OneOf` hasn't attribute `p3`, so it must be `p=p3` and `p3=0.2`.


Datasciensyash:
Oh, sorry. I updated some files in my repo and i had to cancel them) (I have actually done it)

Issue #418, opened 2019-10-14
Cutout does nothing


igriffer:
## 🐛 Bug

Cutout with max_h_size == 1 or max_w_size == 1 does nothing

## To Reproduce

aug = Cutout(max_h_size=1, max_w_size=80, p=1) 
augment_and_show(aug, image)

## Expected behavior

8 black lines with height 1px

## Environment

 - Albumentations version (e.g., 0.1.8): 0.3.3
 - Python version (e.g., 3.7): 3.6.8
 - OS (e.g., Linux): Linux
 - How you installed albumentations (`conda`, `pip`, source): pip

skmatz:
First, `albumentations.Cutout` has been deprecated.
Use `albumentations.CoarseDropout` as in the warning.

skmatz:
I also found the cause of the bug.
The bug occurs not only when `max_h_size == 1` or `max_w_size == 1`, but also when they are odd.

Dipet:
Do not use Cutout it is deprecated. Use [CoarseDropout](https://albumentations.readthedocs.io/en/latest/api/augmentations.html#albumentations.augmentations.transforms.CoarseDropout)

PR #526, opened 2020-02-04
Fix Cutout bug #418

skmatz:
Fix Cutout bug #418.

This bug is caused by rounding down both pos and neg values when adding and subtracting an `odd // 2` from the standard coordinates.

I know this `albumentations.Cutout` has been deprecated, but I think it is a big bug that the box is not displayed when the size is set to 1, so I hope to merge it.

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -1483,9 +1483,15 @@ class Cutout(ImageOnlyTransform):
             x = random.randint(0, width)
 
             y1 = np.clip(y - self.max_h_size // 2, 0, height)
-            y2 = np.clip(y + self.max_h_size // 2, 0, height)
+            if self.max_h_size % 2 == 0:
+                y2 = np.clip(y + self.max_h_size // 2, 0, height)
+            else:
+                y2 = np.clip(y + self.max_h_size // 2 + 1, 0, height)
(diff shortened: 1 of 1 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -1483,9 +1483,15 @@ def get_params_dependent_on_targets(self, params):
...
+            if self.max_h_size % 2 == 0:
+                y2 = np.clip(y + self.max_h_size // 2, 0, height)
+            else:
+                y2 = np.clip(y + self.max_h_size // 2 + 1, 0, height)
             x1 = np.clip(x - self.max_w_size // 2, 0, width)
-            x2 = np.clip(x + self.max_w_size // 2, 0, width)
+            if self.max_w_size % 2 == 0:
+                x2 = np.clip(x + self.max_w_size // 2, 0, width)
+            else:
+                x2 = np.clip(x + self.max_w_size // 2 + 1, 0, width)
```
Dipet: I think is better to write:
```
y2 = np.clip(y1 + self.max_h_size, 0, height)

x2 = np.clip(x1 + self.max_w_size, 0, width)
```

skmatz: Exactly, thanks!

PR #533, opened 2020-02-12
#493 Fix case when label_fields in bbox_params of Compose has multiple values

tswr:
Hi, I also came across issue #493 . This patch seems to fix it for me.


Dipet:
It is similar to #504

Dipet:
Resolved in  #504

PR #536, opened 2020-02-14
Finetune GaussianBlur sigma

henrique:
Allows you to set min_max_sigma range instead of blur_limit (kernel size), which was inherited from the box blur.
The minimal kernel size 3 gives sigma 0.8 which can already be quite a lot for small images.
min_max_sigma also gives you now a continuous space, with way more variance than before...
I calculate the kernel size, using the inverse of the formula, which was previously used to calculate the sigma from the kernel size (with a minimum of size 3).
```
sigma = 0.3*((ksize-1)*0.5 - 1) + 0.8
ksize = max(3, math.ceil((sigma - 0.5)/0.3)*2 + 1)
```

### DIFF
--- a/albumentations/core/transforms_interface.py
+++ b/albumentations/core/transforms_interface.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 
 import random
 from warnings import warn
+from typing import Sequence
 
 import cv2
 from copy import deepcopy
@@ -33,7 +34,7 @@ def to_tuple(param, low=None, bias=None):
             param = -param, +param
         else:
             param = (low, param) if low < param else (param, low)
-    elif isinstance(param, (list, tuple)):
+    elif isinstance(param, Sequence):
         param = tuple(param)
     else:
         raise ValueError("Argument param must be either scalar (int, float) or tuple")


albumentations/augmentations/transforms.py
```diff
@@ -2471,12 +2473,24 @@ class GaussianBlur(Blur):
...
-    def apply(self, image, ksize=3, **params):
+    def apply(self, image, ksize=3, sigma=0, **params):
         return F.gaussian_blur(image, ksize)
 
+    def get_params(self):
+        if self.min_max_sigma is None:
+            return {"ksize": int(random.choice(np.arange(self.blur_limit[0], self.blur_limit[1] + 1, 2)))}
+        else:
+            sigma = random.uniform(self.min_max_sigma[0], self.min_max_sigma[1])
+            return {"sigma": sigma, "ksize": max(3, math.ceil((sigma - 0.5) / 0.3) * 2 + 1)}
```
Dipet: Please set `ksize` to zeros, `GaussianBlur` automatically produce `ksize` if sigma is set.

henrique: Thanks for checking it guys, and thanks for the comment @Dipet 

Unfortunately, the c++ function always dies on an assert, if I set ksize=0:
`OpenCV(4.1.2) /io/opencv/modules/imgproc/src/smooth.dispatch.cpp:189: error: (-215:Assertion failed) ksize.width > 0 && ksize.width % 2 == 1 && ksize.height > 0 && ksize.height % 2 == 1 in function 'createGaussianKernels'`

looking at the code:
`ksize.width = cvRound(sigma1*(depth == CV_8U ? 3 : 4)*2 + 1)|1;`
https://github.com/opencv/opencv/blob/4.1.2/modules/imgproc/src/smooth.dispatch.cpp#L184
or master:
https://github.com/opencv/opencv/blame/master/modules/imgproc/src/smooth.dispatch.cpp#L291
I guess the depth is not right, because it dies with any sigma I use (even really large ones, like 10+)
Anyway, their formula would not work with sigmas below 0.5, as it gives you a useless ksize == 1

Dipet: OpenCV wait tuple of values `(0, 0)`. In this case OpenCV works well with any sigma value.

henrique: @Dipet  gaussian_blur already transforms ksize into a tuple:
```
def gaussian_blur(img, ksize, sigma=0):
    # When sigma=0, it is computed as `sigma = 0.3*((ksize-1)*0.5 - 1) + 0.8`
    blur_fn = _maybe_process_in_chunks(cv2.GaussianBlur, ksize=(ksize, ksize), sigmaX=sigma)
    return blur_fn(img)
```

Do you guys approve the PR as it is, or should I do something else?

albumentations/augmentations/transforms.py
```diff
@@ -2471,12 +2473,24 @@ class GaussianBlur(Blur):
...
-    def __init__(self, blur_limit=7, always_apply=False, p=0.5):
-        super(GaussianBlur, self).__init__(blur_limit, always_apply, p)
+    def __init__(self, blur_limit=7, min_max_sigma=None, always_apply=False, p=0.5):
+        super(GaussianBlur, self).__init__(always_apply, p)
+        self.blur_limit = to_tuple(blur_limit, 3)
+        self.min_max_sigma = min_max_sigma
 
-    def apply(self, image, ksize=3, **params):
+    def apply(self, image, ksize=3, sigma=0, **params):
         return F.gaussian_blur(image, ksize)
```
Dipet: It is because you do not pass sigma value into `gaussian_blur`

Dipet: Please fix this and add simple test, to check that it is really works.

henrique: well spotted, I messed it up here
I will have a look on your tests and write some when I find some time
Although I like the idea of leaving opencv to decide on the kernel size, it doesn't help if it uses a size 1 :/
So I'd leave the formula as it is (or use any other). I hope you agree with that.

Dipet: > it doesn't help if it uses a size 1

I do not understand this.
Can you explain this in more detail.

henrique:
If we pass ksize = 0 we fall into this piece of code below right?
```
    // automatic detection of kernel size from sigma
    if( ksize.width <= 0 && sigma1 > 0 )
        ksize.width = cvRound(sigma1*(depth == CV_8U ? 3 : 4)*2 + 1)|1;
    if( ksize.height <= 0 && sigma2 > 0 )
        ksize.height = cvRound(sigma2*(depth == CV_8U ? 3 : 4)*2 + 1)|1;
```
and cvRound(sigma1*(depth == CV_8U ? 3 : 4)*2 + 1)|1 will be equal to 1 if sigma is below 0.25
e.g. round(0.2*1*2 + 1)|1 == 1
Any kernel of 1x1 is equal to [1.0] so it's the same as nothing (identity) so I'd like to limit the kernel size to at least 3x3
You can argue that sigma 0.25 is too small to make any difference anyway...
```
cv2.getGaussianKernel(ksize=3, sigma=0.25)
Out[3]: 
array([[3.35237708e-04],
       [9.99329525e-01],
       [3.35237708e-04]])
```
Then we could alternatively just limit the minimum sigma to 0.25, but I'd think we could leave that to the user to decide?
I'm not too fussy either way

BloodAxe:
Once https://github.com/albumentations-team/albumentations/pull/551 is merged, please update this PR to use new `gaussian_blur` with `sigma_x` and `sigma_y` parameter from there.

henrique:
- rebased master (which seems to have brought some deepsource error in core?) 
- aligned gaussian_blur with #551 as @BloodAxe suggested
- use default ksize from cv2 as @Dipet requested (ksize=0 when sigma is set)
- added unit-tests for GaussianBlur

albumentations/augmentations/functional.py
```diff
@@ -1087,7 +1087,7 @@ def grid_distortion(
     x_step = width // num_steps
     xx = np.zeros(width, np.float32)
     prev = 0
-    for idx in range(num_steps):
+    for idx in range(num_steps + 1):
```
henrique: conflicts rebased

henrique:
Hi guys,
Do you want me to rebase again from master, or start a fresh PR to clear to failing lint tests from core?
I can do either one, if you are still interested in merging this... just let me know
Thanks!

PR #537, opened 2020-02-19
AugMix augmentation

monuvio:
added AugMix aug, similar by https://arxiv.org/abs/1912.02781


Dipet:
Thank you for your PR. Our CI test catch some problems with code formatting. Please use pre-commit hook to avoid this problems, like described [there](https://github.com/albumentations-team/albumentations/blob/master/docs/contributing.rst) Or format code manually using this command `black --config=black.toml` and check format `flake8`

albumentations/augmentations/transforms.py
```diff
@@ -3320,3 +3321,61 @@ def get_transform_init_args_names(self):
...
+        return image
+
+    def apply(self, img, **params):
+        ws = np.float32(np.random.dirichlet([self.alpha] * self.width))
+        m = np.float32(np.random.beta(self.alpha, self.alpha))
+
+        flag_float = 0
+        if img.dtype in ['float64', 'float32']:
+            img_format = img.dtype
+            img = np.clip(img * 255., 0, 255).astype(np.uint8)
```
albu: why do we need this?

monuvio:
> Thank you for your PR. Our CI test catch some problems with code formatting. Please use pre-commit hook to avoid this problems, like described [there](https://github.com/albumentations-team/albumentations/blob/master/docs/contributing.rst) Or format code manually using this command `black --config=black.toml` and check format `flake8`

Have no problem with flake8, but black test didn't passed 
also, test for docs didn't pass either, but i add AugMix URL in Readme.md
how can i solve this?

albumentations/augmentations/transforms.py
monuvio: some augmentations works only with uint8 images, with this row i'm avoiding exceptions with this type of augs

why do we need float format flag?
in particular case we use augs for float images and expect float images with return(actually, that's one of the test)
you can see usage of float flag and image format in return of function

Dipet:
To fix black errors do `black --config=black.toml `
You mark transform like `ImageOnlyTransform` but you create docs for `DualTransform`
If Transform works with masks, change base class to `DualTransform`

albumentations/augmentations/transforms.py
```diff
@@ -3320,3 +3321,61 @@ def get_transform_init_args_names(self):
...
+        self.alpha = alpha
+        self.augmentations = augmentations
+
+    def apply_op(self, image, op):
+        image = op(image=image)['image']
+        return image
+
+    def apply(self, img, **params):
+        ws = np.float32(np.random.dirichlet([self.alpha] * self.width))
+        m = np.float32(np.random.beta(self.alpha, self.alpha))
```
Dipet: I think we need to create this variables in `get_params` if transform will support masks.

albumentations/augmentations/transforms.py
albu: I think it's better to do the same thing as in other augmentations because silent casting and unnormalization could lead to tricky problems (there are cases when you float image is not normalized by only division by 255.

In this particular case I would leave image format as is and one of subsequent augmentations will throw exception.

albumentations/augmentations/transforms.py
monuvio: Done

albumentations/augmentations/transforms.py
monuvio: sounds reasonable

monuvio:
Should i write tests, that includes every augmentation?(except integer only ones)

albumentations/augmentations/transforms.py
```diff
@@ -3320,3 +3321,56 @@ def get_transform_init_args_names(self):
...
+        return image
+
+    def apply(self, img, **params):
+        mix = np.zeros_like(img)
+        for i in range(self.width):
+            image_aug = img.copy()
+
+            for _ in range(self.depth):
+                op = np.random.choice(self.augmentations)
+                image_aug = self.apply_op(img, op)
```
BloodAxe: I feel this line should be `image_aug = self.apply_op(image_aug , op)`. Otherwise image_aug  will be an augmented image only by lastly-picked augmentation.

BloodAxe: I beleive this line should be changed to `image_aug = self.apply_op(img, op)`.

monuvio: that's actually huge mistake, don't know how that happened

albu:
yes, it's better to write at least some tests which ensures that augmentation does what expected.

albumentations/augmentations/transforms.py
```diff
@@ -3320,3 +3321,56 @@ def get_transform_init_args_names(self):
...
+
+            mix = np.add(mix, self.ws[i] * image_aug, out=mix, casting="unsafe")
+
+        mixed = (1 - self.m) * img + self.m * mix
+        if img.dtype in ["uint8", "uint16", "uint32", "uint64"]:
+            mixed = np.clip((mixed), 0, 255).astype(np.uint8)
+        return mixed
+
+    def get_transform_init_args_names(self):
+        return ("width", "depth", "alpha")
```
albu: looks like augmentations are not here so they would not be processed correctly during serialization. 
this augmentation looks `Compose`-ish so please take a look at `Compose`-like augmentations. 
and you can write test for serialization of this class

albumentations/augmentations/transforms.py
```diff
@@ -3320,3 +3321,56 @@ def get_transform_init_args_names(self):
...
+    """
+
+    def __init__(self, width=2, depth=2, alpha=0.5, augmentations=[HorizontalFlip()], always_apply=False, p=0.5):
+        super(AugMix, self).__init__(always_apply, p)
+        self.width = width
+        self.depth = depth
+        self.alpha = alpha
+        self.augmentations = augmentations
+        self.ws = np.float32(np.random.dirichlet([self.alpha] * self.width))
+        self.m = np.float32(np.random.beta(self.alpha, self.alpha))
```
BloodAxe: Sampling random numbers from numpy in c'tor is a bad practice for two reasons:
1) As we rely on default python random seed, it may have undesirable effect and break reproducible results. You can initialize numpy random generator from seed generated by python random and use it as pseudo-random generator state for `np.random.beta` and `np.random.dirichlet`
2) These weights will be the same for all augmented samples. Is this what was intended? As I understand this could be moved to `get_params`.

monuvio: 1.`
self.seed = random.randint(10,10000);
 np.random.seed(self.seed)
`
And then pass seed to init_args_names for reproducibility?
Did you mean something like that?

2. totally agree

monuvio:
this aug should be Compose-ish like, not ImageOnlyTransformation. should be better move this in different PR

PR #543, opened 2020-02-25
Fix grid distortion ranges

Dipet:
Linspace produce wrong step and we get artifacts on image.
image


albu:
why cannot we just revert https://github.com/albumentations-team/albumentations/pull/511?

Dipet:
In #511 you can see case when `range(0, height, y_step)` creates sequence with wrong number of elements, i.e. `len(list(range(0, height, y_step))) > num_steps`

Issue #477, opened 2019-11-26
Add new transforms from https://arxiv.org/abs/1903.12261


ternaus:
There is a set of Pixel Level transforms that is used in the work `Benchmarking Neural Network Robustness to Common Corruptions and Perturbations`

The authors also share the code => we can absorb some transforms that they have into the library.

https://github.com/hendrycks/robustness/blob/master/ImageNet-C/create_c/make_imagenet_c.py

 - [x] ShotNoise
 - [x] Defocus
 - [x] GlassBlur
 - [x] Zoom Blur
 - [x] Pixelate
 - [ ] Frost
 - [x] Spatter

ternaus:
Pixelate in the paper == Dowscale with `cv2.INTER_NEAREST`

ternaus:
ShotNoise in paper == ISONoise in the Library

akarsakov:
@ternaus I'd like to take up this enhancement.

akarsakov:
Hi @ternaus,

Frost transform requires precalculated images with ice crystals. 
What's the best way of storing this images in albumentations code (all images require ~2MB)?

Dipet:
It seems that in this transformation we can use any image templates..  I think it's best to name this transformation as a `TemplateTranform`. And give the opportunity to set any image as a template.

akarsakov:
That's great point! 
Moreover looks like it would be better to allow setting list of desired images as template and pick the random one.

p4rallax:
Has this been finished?

akarsakov:
@p4rallax The code is finished (please see PRs mentioned in this issue) but it haven't merged to master yet.

zetyquickly:
Based on the PRs seems like the implementation is actually missing Frost only

ternaus:
It looks like it was decided that Frost is the same as Template but with frost images.

Although this non-obvious fast should obviously covered in the blog post / tutorial.

PR #551, opened 2020-03-06
Add Defocus and ZoomBlur transforms

akarsakov:
Add Defocus and ZoomBlur transforms from 'Benchmarking Neural Network Robustness to Common Corruptions and Perturbations' (https://arxiv.org/abs/1903.12261)

Refer to issue: https://github.com/albumentations-team/albumentations/issues/477


albumentations/augmentations/transforms.py
```diff
@@ -3320,3 +3322,73 @@ def get_transform_init_args_names(self):
...
+
+    Image types:
+        Any
+    """
+
+    def __init__(self, severity=3, always_apply=False, p=0.5):
+        super(Defocus, self).__init__(always_apply, p)
+        self.severity = to_tuple(severity, low=1)
+
+        assert self.severity[0] >= 1, "Expected minimal severity greater than or equal to 1, got {}".format(
```
BloodAxe: Can you please change assertion to raising a run-time error in case incorrect range of argument? As a rule of thumb - asserts for tests, exceptions for the run-time errors.

akarsakov:
Hi @Dipet, @BloodAxe 
Sorry for late response. I've resolved all your comments.

albumentations/augmentations/transforms.py
```diff
@@ -3320,3 +3322,91 @@ def get_transform_init_args_names(self):
...
+
+        if self.alias_blur[0] <= 0:
+            raise ValueError("Parameter alias_blur must be non-negative")
+
+    def apply(self, img, radius=3, alias_blur=0.5, **params):
+        return F.defocus(img, radius, alias_blur)
+
+    def get_params(self):
+        return {
+            "radius": random.choice(np.arange(self.radius[0], self.radius[1] + 1)),
```
Dipet: Better to use `np.random.randint(self.radius[0], self.radius[1] + 1)`

albumentations/augmentations/transforms.py
```diff
@@ -3320,3 +3322,91 @@ def get_transform_init_args_names(self):
...
+    def __init__(self, radius=(3, 10), alias_blur=(0.1, 0.5), always_apply=False, p=0.5):
+        super(Defocus, self).__init__(always_apply, p)
+        self.radius = to_tuple(radius, low=1)
+        self.alias_blur = to_tuple(alias_blur, low=0)
+
+        if self.radius[0] < 0:
+            raise ValueError("Parameter radius must be positive")
+
+        if self.alias_blur[0] <= 0:
+            raise ValueError("Parameter alias_blur must be non-negative")
```
Dipet: I think this checks are redundant because you described the minimum values ​​in `to_tuple`

albumentations/augmentations/transforms.py
```diff
@@ -3320,3 +3322,91 @@ def get_transform_init_args_names(self):
...
+        Any
+    """
+
+    def __init__(self, max_factor=1.31, step_factor=(0.01, 0.03), always_apply=False, p=0.5):
+        super().__init__(always_apply, p)
+        self.max_factors = to_tuple(max_factor, low=1.0)
+        self.step_factors = to_tuple(step_factor, step_factor)
+
+        if self.max_factors[0] < 1:
+            raise ValueError("Max factor must be larger or equal 1")
```
Dipet: I think this check is redundant because you described the minimum values ​​in to_tuple

albumentations/augmentations/transforms.py
```diff
@@ -3320,3 +3322,91 @@ def get_transform_init_args_names(self):
...
+    def apply(self, img, zoom_factors=np.arange(1, 1.11, 0.01), **params):
+        return F.zoom_blur(img, zoom_factors)
+
+    def get_params(self):
+        max_factor = random.uniform(self.max_factors[0], self.max_factors[1])
+        step_factor = (
+            self.step_factors[0]
+            if self.step_factors[0] == self.step_factors[1]
+            else random.uniform(self.step_factors[0], self.step_factors[1])
+        )
```
Dipet: Do not need to check that values are equal.
`random.uniform(self.step_factors[0], self.step_factors[1])` works well, when `self.step_factors[0] == self.step_factors[1]`

Dipet:
Please, also add transforms to `test_serialiazation.py`

albumentations/augmentations/transforms.py
akarsakov: It's for the case when radius is set to negative value x, so tuple would be (x, 0).
On the other hand, user could set interval like (-5, 100) and this check will not allow it.

albumentations/augmentations/transforms.py
akarsakov: Oh, no reason, will fix.

akarsakov:
Hi @Dipet, @BloodAxe,
I've updated code according to your comments, but 'DeepSource: Python' check is failed and probably not in my code.

akarsakov:
Hi @Dipet, @BloodAxe,
I've rebased branch. Now all checks are passed.

akarsakov:
Hi @Dipet, @BloodAxe,
Have you got a chance to look into code review?

ahaliassos:
Hi, any news on when these two transforms will be added to albumentations?

Dipet (approved):

albumentations/augmentations/geometric/functional.py
```diff
@@ -377,22 +379,6 @@ def elastic_transform(
     return remap_fn(img)
 
 
-@preserve_channel_dim
-def resize(img, height, width, interpolation=cv2.INTER_LINEAR):
```
creafz, line 381: @Dipet Is it an intended change to move `resize` and `scale` from `albumentations/augmentations/geometric/functional.py` to `albumentations/augmentations/functional.py`?

Dipet: Yes, this is intentional because I don't know how to import them from `geometric.functional` to avoid infinite imports (`augmentations.geometric.functional` imports some names from `augmentations.functional`)

creafz (approved):

PR #553, opened 2020-03-09
Added config for deepsource

ternaus:


.deepsource.toml
```diff
@@ -0,0 +1,17 @@
...
+exclude_patterns = [
+
+]
+
+[[analyzers]]
+name = 'python'
+enabled = true
+runtime_version = '3.x.x'
+
+  [analyzers.meta]
```
BloodAxe: Tabulation issue?

ternaus: I think it is automatically generated this way: https://github.com/BloodAxe/pytorch-toolbelt/blob/develop/.deepsource.toml

Issue #288, opened 2019-07-05
KeyError: 'cropping_bbox' when using RandomCropNearBBox


Susia00:
I  use this method:
aug = get_aug([RandomCropNearBBox(max_part_shift=0.3,p=1)])
augmented = aug(**annotations)

but a keyError happened: 
albumentations-0.3.0-py3.5.egg/albumentations/core/composition.py", line 194, in __call__
albumentations-0.3.0-py3.5.egg/albumentations/core/transforms_interface.py", line 57, in __call__
albumentations-0.3.0-py3.5.egg/albumentations/core/transforms_interface.py", line 57, in <dictcomp>
KeyError: 'cropping_bbox'
 Does someone can solve it?

albu:
this method is a little unstable. you need to pass bounding box to crop by as argument "cropping_bbox" and run cropping before all other augmentations.

Susia00:
> this method is a little unstable. you need to pass bounding box to crop by as argument "cropping_bbox" and run cropping before all other augmentations.

Could you give me an example in code?   It still has some running errors. Thx.

vfdev-5:
Any updates on this ?

zhaonat:
also hoping for an update on this?

NOxDriver:
I'm having the same error

fadamsyah:
I have the same problem. Hope to see an update on this issue

Dipet:
Fixed in #993

PR #556, opened 2020-03-11
Fix RandomCropNearBBox

Dipet:
#288


NickLekkas01 (changes requested):
779: should be bbox = params[self.cropping_box_key] in get_params_dependent_on_targets

Dipet:
reopened new PR #993

PR #560, opened 2020-03-14
add assertion if passing data to augs as not named arguments

albu:


Dipet:
Maybe it will be nice to check that `image` and `mask`, if it is exists, is `np.ndarray`?

Issue #565, opened 2020-03-16
bug for Special enhancement


hhaAndroid:
Enhance the demo image as follows, a bug will appear
![image](https://user-images.githubusercontent.com/17425982/76735991-05a02b80-67a1-11ea-8caa-5f660e89a08b.png)

![image](https://user-images.githubusercontent.com/17425982/76736021-118bed80-67a1-11ea-8ecb-9e3d9de568b4.png)

The reason is obvious. There is no filter operation for each enhancement function.

The code that should be modified is：
![image](https://user-images.githubusercontent.com/17425982/76736169-47c96d00-67a1-11ea-9c43-04d9f69d75ae.png)

Dipet:
Please do not use screenshots for code, copy past it into issue as text and wrap it into code specifiers.

BloodAxe:
Please, use the suggested issue template. By explaining the issue in a structured way, you are helping maintainers to understand and address your issue faster. 

More specifically, there is parameter to `A.Compose` that controls how keypoints are removed during augmentations. It is not clear what parameter value you are using, therefore as it is now, the issue cannot be reproduced.

PR #566, opened 2020-03-16
Check each transform for bboxes and keypoints

Dipet:
#565

### DIFF
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -167,6 +167,9 @@ class Compose(BaseCompose):
             p.ensure_data_valid(data)
         transforms = self.transforms if need_to_run else self.transforms.get_always_apply(self.transforms)
         dual_start_end = transforms.start_end if self.processors else None
+        check_each_transform = any(
+            getattr(item.params, "check_each_transform", False) for item in self.processors.values()
+        )
 
         for idx, t in enumerate(transforms):
             if dual_start_end is not None and idx == dual_start_end[0]:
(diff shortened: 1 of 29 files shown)


albumentations/core/composition.py
```diff
@@ -379,16 +387,25 @@ class BboxParams(Params):
             visible area in pixels is less than this value will be removed. Default: 0.0.
         min_visibility (float): minimum fraction of area for a bounding box
             to remain this box in list. Default: 0.0.
+        check_each_transform (bool): if `True`, then bboxes will be checked after each dual transform.
+            Default: `True`
```
creafz, line 391: The docstring says that the default value is `True`, but the default value in `__init__` is set to `False`.

albumentations/core/composition.py
```diff
@@ -167,6 +167,9 @@ def __call__(self, force_apply=False, **data):
             p.ensure_data_valid(data)
         transforms = self.transforms if need_to_run else self.transforms.get_always_apply(self.transforms)
         dual_start_end = transforms.start_end if self.processors else None
+        check_each_transform = any(
+            getattr(item.params, "check_each_transform", False) for item in self.processors.values()
+        )
```
creafz, line 172: It looks like that if `check_each_transform` is set to `True` in any processor, then it will affect the behavior of all processors, and they all start to filter values at each step. If this a global setting that affects all processors, maybe we should set it in a separate place and not in a processor?

creafz:
For me, the previous behavior looks like a bug, so maybe we should enable `check_each_transform` by default by setting its default value to `True`?

albumentations/core/composition.py
Dipet: Changed default value to True

albumentations/core/composition.py
Dipet: Now this parameter will be checked only for selected processor.

PR #568, opened 2020-03-18
fixes #548

### LINKED ISSUE #548: RandomScale, ShiftScaleRotate bug
## 🐛 Bug

The logic in the `RandomScale`, `ShiftScaleRotate` is flawed - scale has a bias towards upscale. As can be seen from the [code](https://github.com/albumentations-team/albumentations/blob/bd2c655bc753a72cb48b49eceeabc8dbbc22b9d8/albumentations/augmentations/transforms.py#L568), scale is defined as `scale = uniform([1 - scale_limit + bias, 1 + scale_limit + bias]) - bias`. 

Let's take an example of `scale_limit =(0.5, 2)`:

1. formula above means that we will uniformly sample from [1.5, 3]. which means 2/3 of the sample will be within the range [2, 3] and 1/3 will be within the range [1.5, 2]. With the subtraction of bias this translates into 2/3 chance to upscale and 1/3 chance to downscale.
1. it is unclear what is the purpose of `bias` here, looks pointless (attempt to mitigate the issue above?).

dselivanov:
This is not yet ready to merge, but to foster discussion about #548. We can fix it properly, but it will break current behaviour. Please advise about your tolerance to breaking changes.

### DIFF
--- a/albumentations/__init__.py
+++ b/albumentations/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-__version__ = "0.4.5"
+__version__ = "0.4.5.1"
 
 from .core.composition import *
 from .core.transforms_interface import *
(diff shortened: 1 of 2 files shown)


albu:
breaking changes are not allowed at all. the only possible solution is to add parameter to select another behaviour

Dipet:
This PR changes default behavior of transform. In that case it is unjustified.

dselivanov:
Amen

PR #582, opened 2020-04-03
Added basic Point Clouds augmentations as pcaug

kumuji:
1. In the init of albumentations, I added import from an additional module called pcaug. The name pcaug was chosen because it is common to shorten "points clouds" into "pc". Also "aug" part comes from another module which already exists in the code, called "imgaug"
  
  
2. I had to add "PointCloudsTransform" in the transforms interface because any other didn't fulfill the requirements for the point cloud transformations. I also had to change "BasicTransform", because now not all the time "image" is passed thus it should be optional. All tests are passed.
But there are several problems and features need to be implemented in this part:
    - The most important is to able to work with color features in the point cloud using image interface, just passing it together with point cloud as a slice. But so far it is very hard to implement clearly, for example, because of dropout function. If the color features of the point cloud is passed as an image, the length of it will change during the process of augmentation. So it should be investigated more;
    - Add bounding boxes support (very simple, almost done it);
    - Add meshes support.
  
  
3. I added a few basic transformations, just for start + some notes:
    - "Scale3d" - implemented super simple, maybe should be reimplemented as general affine transformation, to support operations such as shear and movement
    - "RotateAroundAxis3d" - if 3 rotations needed should be combined by the user. Because different sequences might give different results for some inputs. Sometimes the standard is xyz, zyz', and so on. So, better give it to the user. Also, this form makes an easier interface
    - "Crop3d" - nothing important, but maybe also RandomCrop should be added
    - "RandomMove3d" - Moves all pc, nothing important, but, maybe can be reimplemented through functional from Scale3d as part of affine transformation
    - "Center3d" - Moves origin to mean of the point cloud, might not be optimal
    - "RandomDropout3d"
    - "Flip3d" - Implemented as rotation on 90deg. But also can be implemented using affine transformation. It Will take fewer computations, but in terms of optimization, it is not computed for very long anyway. Maybe optimized, but not necessary.
    - Should be considered to add, and would be easy:
        - "RandomCrop3d" - will add as soon as
        - "Frustum dropout" - similar as random crop as provided here: https://arxiv.org/pdf/2004.00831.pdf
        - "Noise" - just apply jitter to points. Also very easy. Same as move, just perturb every point by a small amount.
        - "AffineTransform" - should also be easy
        - "AugmentWithGroundThruth" - use some database to augment frame with more ground truth. It can be seen in SECOND paper. But it is a bit harder because if pc is crowded, it will take additional time.

4. Few functionals were added: scale, rotate_around_axis, crop, move, center. All of them can work in the way, that the first three columns in data are coordinates.
5. And parametrized tests for functionals.
6. Also small documentation for the new module.
7. So far I didn't add new lines to the main README and I don't reaaly know the best way to change it, so it is expected to fail one test.


Dipet:
Thank you for PR.
Can you explain to me, this transforms works without images, only on point clouds?
As I see it, this transforms will work with images and point clouds from lidar. So we need there interface to work with images to.

kumuji:
Yes, it is working only on point clouds and gives a basic interface to it.
Images and point clouds have a different nature, one data is sparse, another is dense. Connecting them is not trivial. 

There two types of data I saw:
1. For 3d object detection datasets we usually have lidar and images from different points in space. While it might be easy to apply some transformations like scale or rotation to both of them, applying something like elastic transformation or populating point cloud with ground truth might not be that easy to combine. Difficulties in fusion are partly the reason why most of the methods so far are working only with lidar data.
2. Segmentation or classification datasets add color information as one of the features of a point. And if dropout applied on the point level it would also affect color information. Which requires the whole package to be able to work with variable size images. And adding an image interface wouldn't do anything for these types of datasets.

My idea was to provide the same interface as albumentations, but for working with point clouds. Not all of the tasks require working with images and point clouds together. And most of the methods are working only with point clouds.

What might be a compromise is adding an interface for extrinsic/intrinsic camera parameters later. It will ensure that camera positions are moved correctly for the task of 3d object detection.

Dipet:
Well, in many cases we will work only with points. But what about `Scale` transform? I think than when we scale points, we need scale image as well. We need an interface to for image processing or some mechanism to get the applied transform params.

Dipet:
Thank for your PR. But we decide to focus only on images and targets on images. So this functionality is overhead for single library.
You can create your own package with supports this targets.

kumuji:
Sure, in the middle of my work I realized that it is not fitting very good fit in the current package, so I started to work slowly on my stuff taking a lot from your code(mostly cutting things I don't need), because for me images are not very helpful and my work lays in a bit different domain. And your code for me is like a good standard where I should aim at.
Thank you for the great package!

Would it be okay if I can use your code as a base for the package on 3d data? Sorry for not saying this few days later. I would try to give as much credit as I can

Dipet:
It is great if you will use albumentations as a base for your package

PR #590, opened 2020-04-13
BboxFormatConvert transform

aclex:
I've written a little additional transform converting `bboxes` target from one bounding box format to another. I use it to make albumentations pipeline compatible with the existing model expecting its input in one particular format (similar to `albumentations` format in my case), while receiving samples in some different format from the initial dataset class (e.g. `pascal_voc`). Maybe you would find it useful for the project as well.

Please, feel free to correct me on naming or source code location, I've tried my best to find the proper place for it, but maybe I've just not get it)

And yes, many thanks for this great project!

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -12,6 +12,7 @@ from skimage.measure import label
 
 from . import functional as F
 from .bbox_utils import denormalize_bbox, normalize_bbox, union_of_bboxes
+from .bbox_utils import convert_bbox_from_albumentations, convert_bbox_to_albumentations, check_bbox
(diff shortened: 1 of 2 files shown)


Dipet:
Hi, thanks for PR. Transforms is not suitable for these purposes because bboxes formats changes in [BbboxProcessor](https://github.com/albumentations-team/albumentations/blob/9c225a99a379594098dbea2a077fd22da684ade9/albumentations/augmentations/bbox_utils.py#L21)

I am resolved it in #594

aclex:
@Dipet thanks for taking a look at it! Indeed, there's probably more neat and consistent approach in #594, hope it would be merged soon and become available for use.

PR #596, opened 2020-04-20
Do not show deprecated functions in README.md

Dipet:


tools/make_transforms_docs.py
```diff
@@ -156,7 +158,7 @@ def check_docs(filepath, image_only_transforms_links, dual_transforms_table):
     )
 
 
-if __name__ == "__main__":
+def _main():
```
creafz, line 161: Not sure, but maybe this function should me named `main()` (without an underscore), because it is a function that serves as an entrypoint to this file and not a private or protected one?

Dipet: done

Issue #478, opened 2019-11-27
Fill Color: Random ?


MentalGear:
## Question / Improvement

In the api, some parameters have a fill_value. Is there an option for a random color fill ?

akarsakov:
Hi all,
I would like to take up this enhancement. 
I propose to set fill_value parameter to None (or 'random' string for example) in order to generate it randomly. It's gonna to affect following transforms:
- PadIfNeeded
- ShiftScaleRotate
- OpticalDistortion
- GridDistortion
- ElasticTransform
- Cutout
- CoarseDropout
- ChannelDropout
- MaskDropout
- GridDropout
Probably I missed something.. 

@albu, @ternaus, @kirillbobyrev, @BloodAxe, @tatigabru,
What do you think?

akarsakov:
I've decided to add this ability only for CoarseDropout, MaskDropout, and GridDropout. Looks like it doesn't make a lot of sense for others.
Created PR: #597

ternaus:
Added random filling for `CoarseDropout` in  https://github.com/albumentations-team/albumentations/pull/1677

For other transforms - could be added.

We added stricted checks for types, hence it is a bit trickier to add random color than in https://github.com/albumentations-team/albumentations/pull/597

Also we want to support different image types, hence random sampling should be extended to other types beyond `float32` and `unit8`. Still kind of similar to the PR fro @akarsakov

ogencoglu:
Turning `fill` into a range to randomly sample from in `ChannelDropout` would be really useful indeed

ternaus:
@ogencoglu you can get this functionality using 

`OneOf( [ChannelDropout(fill = fill) for fill in <list of fill values>] )`

PR #597, opened 2020-04-21
Random fill_value

akarsakov:
Refer to #478 

Add ability to set fill_value='random' and randomly generate fill color.


akarsakov:
Deep source has been failed not in my changes. Not sure how to properly suppress it.

PR #609, opened 2020-04-30
Update ToTensorV2 in transforms.py

svdcvt:
ValueError: At least one stride in the given numpy array is negative, and tensors with negative strides are not currently supported. (You can probably work around this by making a copy of your array  with array.copy().)


Dipet:
Hi
Thanks for PR!
You catch this error with our transforms?
This errors occurs when image is not contiguous, better to call `np.ascontiguousarray`

Dipet:
The author did not provide a reproducible example.

PR #616, opened 2020-05-04
Fix #615 bboxes with min_visibility are excluded

### LINKED ISSUE #615: bboxes with min_visibility are excluded
`min_visibility` bbox parameter works as non-strict limit to exclude bboxes. I.e. bboxes with visibility ratio == min_visibility  are excluded.
It not only contradicts to documentation ('min_visibility (float): Minimum fraction of area for a bounding box to remain this box in list.') but causes the problem that if I want to retain only not clipped bboxes, setting `min_visibility=1.0` results in all bboxes are filtered

IlyaOvodov:
Fix #615 bboxes with min_visibility are excluded.
Both cases min_visibility=0 and min_visibility=1 are now handled correctly.


albumentations/augmentations/bbox_utils.py
```diff
@@ -363,10 +363,9 @@ def filter_bboxes(bboxes, rows, cols, min_area=0.0, min_visibility=0.0):
         transformed_box_area = calculate_bbox_area(bbox, rows, cols)
         bbox, tail = tuple(np.clip(bbox[:4], 0, 1.0)), tuple(bbox[4:])
         clipped_box_area = calculate_bbox_area(bbox, rows, cols)
-        if not transformed_box_area or clipped_box_area / transformed_box_area <= min_visibility:
+        if clipped_box_area == 0 or clipped_box_area / transformed_box_area < min_visibility:
```
Dipet, line 366: I think better to check `transformed_box_area == 0` because it is needed to avoid division by zero.

albumentations/augmentations/bbox_utils.py
```diff
@@ -363,10 +363,9 @@ def filter_bboxes(bboxes, rows, cols, min_area=0.0, min_visibility=0.0):
...
             continue
-        else:
-            bbox = tuple(np.clip(bbox[:4], 0, 1.0))
+        bbox = tuple(np.clip(bbox[:4], 0, 1.0))
```
Dipet, line 368: We can remove this line. We did this operation above.

albumentations/augmentations/bbox_utils.py
```diff
@@ -363,10 +363,9 @@ def filter_bboxes(bboxes, rows, cols, min_area=0.0, min_visibility=0.0):
...
         if calculate_bbox_area(bbox, rows, cols) <= min_area:
```
Dipet, line 369: I think, here we should also change `<=` to `<`

albumentations/augmentations/bbox_utils.py
IlyaOvodov: I don't agree, I've changed it intentionally.
`transformed_box_area==0` => `clipped_box_area == 0`, but not `clipped_box_area == 0` => `transformed_box_area==0` .
If you check `transformed_box_area` and it is != 0 as before, then for `min_visibility=0` all boxes including invisible ones with `clipped_box_area == 0` will pass the filter. That's why we need check `clipped_box_area == 0` separately.
Of cause, we can check both just to demonstrate that we avoid division by 0, but it seems overkill :)

Dipet: Ok, I agree. But, I think, need to add comment that we check `clipped_box_area == 0` to properly process this  case, in future we will forget this detail.
Something like this:
```python
# Check `clipped_box_area == 0` to process case when `min_visibility=0`
# and strict comparison for case when `min_visibility=1`
if clipped_box_area == 0 or clipped_box_area / transformed_box_area < min_visibility:
```

albumentations/augmentations/bbox_utils.py
IlyaOvodov: Yes, as well as `calculate_bbox_area(bbox, rows, cols)` :). Done.

albumentations/augmentations/bbox_utils.py
IlyaOvodov: Done

Dipet:
Please add this test to `test_bbox`
```
@pytest.mark.parametrize(
    ["transforms", "bboxes", "result_bboxes", "min_area", "min_visibility"],
    [[Crop(10, 10, 20, 20)], [[0, 0, 10, 10, 0]], [], 0, 0],
    [[Crop(0, 0, 90, 90)], [[0, 0, 90, 90, 0], [0, 0, 89, 89, 0]], [[0, 0, 89, 89, 0]], 0, 1],
    [[Crop(0, 0, 90, 90)], [[0, 0, 1, 10, 0], [0, 0, 1, 11, 0]], [[0, 0, 1, 10, 0], [0, 0, 1, 11, 0]], 10, 0],
)
def test_bbox_params_edges(transforms, bboxes, result_bboxes, min_area, min_visibility):
    image = np.empty([100, 100, 3], dtype=np.uint8)
    aug = Compose(transforms, bbox_params=BboxParams("pascal_voc", min_area=min_area, min_visibility=min_visibility))

    res = aug(image=image, bboxes=bboxes)["bboxes"]
    assert np.allclose(res, result_bboxes)
```

IlyaOvodov:
Done. But had to change the 2nd test: rect(0,0,90,90) is not cropped as rects are considered as semi-open, only rect(0,0,91,91) is

PR #620, opened 2020-05-05
Fix #619. If `bbox_params` is set, fails if no bbox is provided

### LINKED ISSUE #619: If `bbox_params` is set in constructor, processing fails if no bbox data is provided
## 🐛 Bug
```
img = np.zeros((100,100,3), dtype=np.uint8)
augs_list = [HorizontalFlip()]
aug = Compose(augs_list)
x = aug(image=img, bboxes=[])   # OK
x = aug(image=img)   # OK
```
runs ok, but
```
aug = Compose(augs_list, bbox_params={'format': 'albumentations'})
x = aug(image=img, bboxes=[])   # OK
x = aug(image=img)   # Fail
```
fails in the 2nd call.
If somebody configures bbox or other data conversion, it doesn't mean that conversion of it is required every time. Sometimes not.

IlyaOvodov:
Fix #619. If `bbox_params` is set in constructor, processing fails if no bbox data is provided.
The same problem fixed for keypoints and additional targets.

### DIFF
--- a/albumentations/core/utils.py
+++ b/albumentations/core/utils.py
@@ -48,8 +48,9 @@ class DataProcessor:
         rows, cols = data["image"].shape[:2]
 
         for data_name in self.data_fields:
-            data[data_name] = self.filter(data[data_name], rows, cols)
-            data[data_name] = self.check_and_convert(data[data_name], rows, cols, direction="from")
+            if data_name in data:
+                data[data_name] = self.filter(data[data_name], rows, cols)
+                data[data_name] = self.check_and_convert(data[data_name], rows, cols, direction="from")
 
         data = self.remove_label_fields_from_data(data)
         return data
@@ -59,7 +60,8 @@ class DataProcessor:
 
         rows, cols = data["image"].shape[:2]
         for data_name in self.data_fields:
-            data[data_name] = self.check_and_convert(data[data_name], rows, cols, direction="to")
+            if data_name in data:
+                data[data_name] = self.check_and_convert(data[data_name], rows, cols, direction="to")
 
     def check_and_convert(self, data, rows, cols, direction="to"):
         if self.params.format == "albumentations":


albumentations/core/utils.py
```diff
@@ -48,8 +48,9 @@ def postprocess(self, data):
         rows, cols = data["image"].shape[:2]
 
         for data_name in self.data_fields:
-            data[data_name] = self.filter(data[data_name], rows, cols)
-            data[data_name] = self.check_and_convert(data[data_name], rows, cols, direction="from")
+            if data_name in data:
+                data[data_name] = self.filter(data[data_name], rows, cols)
+                data[data_name] = self.check_and_convert(data[data_name], rows, cols, direction="from")
```
Dipet, line 53: I agree with this changes if you will add warning message:
```python
if data_name in data:
    ...
else:
    warnings.warn("Can not find {} in provided data.".format(data_name), RuntimeWarning)
```

IlyaOvodov: This warning will be desplayed every time augmentation is called. In most applications it is not 1 or 2 times :). What for? I can't imagine use case when a didn't add some parameter and will not see it without such a warning. Not to mention the fact that warnings not always are noticed and correctly interpreted.
An only case when it can be usefull is when I misspelled parameter name. Now such a parameter data is just passed through without processing. And it is quite another case because this code even will not be accessed, if no pre- and post- processing is required.
For example:
```
aug = Compose([HorizontalFlip(), RandomCrop(100,100)])  # no processors are created
res = aug(img, bbooks=boxes)  # misspelled
img, boxes = res["image"], res["bbooks"]  # bbooks copy-pasted from above
```
- produce no error or warnings but results in quite unexpected way.
I suppose an exeption should be generated in such case. If somebody needs extra data `bbooks` to be passed through without modification, it should be explicitly defined through `additional_targets` argument, that should supoport not only bbox and keypoint, but also targets with no processing.
But in my oppinion it is another issue.

Dipet:
You are really need this changes?
Why you can not set bboxes on call to empty list?
With this changes code would be less predictable.

IlyaOvodov:
From my point of view behaviour will become more intuitive and will not tend to throw exceptions were they are not expected. Of cause pitfalls like this can be overcomed with some 'plyaski s bubnom' :) if you already know where they are, but more intuitive way is more comfortable. At least it takes some amount of time to find why fhings went wrong, when you face the problem 1st time (as it was with me).

IlyaOvodov:
Currently call to `res = aug(image=img)` results in:
```
  File "D:\Programming\3rd_party\albumentations\albumentations\core\composition.py", line 174, in __call__
    p.preprocess(data)
  File "D:\Programming\3rd_party\albumentations\albumentations\core\utils.py", line 62, in preprocess
    data[data_name] = self.check_and_convert(data[data_name], rows, cols, direction="to")
KeyError: 'bboxes'
```
that is really hard to interpret if you are not acquainted with albumentations' innards.

IlyaOvodov:
I've opened a separate issue #624 about argument names checking.

Dipet:
@IlyaOvodov could you update this PR?

Dipet:
Closed as abandoned

PR #627, opened 2020-05-07
remove np.ascontiguousarray() calls

MichaelMonashev:
np.ascontiguousarray() makes slow memory allocation.


BloodAxe:
This PR definitely needs informative description, motivation section and benchmarks.

Dipet:
Some functions can not work with non-contiguous arrays, for example `torch.from_numpy`

PR #632, opened 2020-05-13
bboxes clipped between 0 and 1. 

snlpatel001213:


Dipet:
Function `check_validity` need to check that all bboxes are correct and in range `[0, 1]`.

PR #637, opened 2020-06-03
Add limit_to_bbox option to GridDistortion

poke1024:
Currently GridDistortion produces distortions than can shift content outside of the image bounding box (i.e. after the distortion, parts of the original content might not be visible). This PR adds a new parameter `limit_to_bbox` that restricts this behaviour such that no content ever gets moved outside (by making sure the sum of distortions does not exceed 1 on each axis). It produces a distortion constrained by the bounding box.

Now, to get this to work properly, I realized just now I had to change some underlying stuff in `grid_distortion`, which gets pretty tricky unfortunately.

I tried to isolate the current way (master) and another new way (this PR) of doing things. Currently things are always calculated with `num_steps + 1` internally, but then divided by `num_steps` when it comes to `x_step`. This is the rough code:

```
import numpy as np

# simplified algorithm for demonstration.

def linear_old(width, num_steps):  # current master
	xsteps = [1 for i in range(num_steps + 1)]  # usually random.
	x_step = width // num_steps
	prev = 0
	for idx in range(num_steps + 1):
		x = idx * x_step
		start = int(x)
		end = int(x) + x_step
		if end > width:
			end = width
			cur = width
		else:
			cur = prev + x_step * xsteps[idx]
		yield (start, end, prev, cur)
		prev = cur

def linear_new(width, num_steps):  # this PR
	xsteps = [1 for i in range(num_steps)]  # usually random.
	x_step = width / num_steps
	prev = 0
	for idx in range(num_steps):
		x = idx * x_step
		start = int(x)
		end = int(x + x_step)
		if idx >= num_steps - 1 or end >= width:
			end = width
			cur = width
		else:
			cur = prev + x_step * xsteps[idx]
		yield (start, end, int(prev), int(cur))
		prev = cur

print(list(linear_old(100, 3)))
print(list(linear_new(100, 3)))
```

With outputs:

```
[(0, 33, 0, 33), (33, 66, 33, 66), (66, 99, 66, 99), (99, 100, 99, 100)]
[(0, 33, 0, 33), (33, 66, 33, 66), (66, 100, 66, 100)]
```

The current solution always produces a strange bit at the end (due to the `//` rounding). So, I wonder, due to these changes, whether there could be an all new `ConstrainedGridDistortion`, that does not need to change the existing code base.

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -1178,6 +1178,8 @@ class GridDistortion(DualTransform):
         num_steps (int): count of grid cells on each side.
         distort_limit (float, (float, float)): If distort_limit is a single float, the range
             will be (-distort_limit, distort_limit). Default: (-0.03, 0.03).
+        limit_to_bbox (bool, (bool, bool)): If limit_to_bbox is False, the distortion may move content outside of
+            the image's bounding box.
         interpolation (OpenCV flag): flag that is used to specify the interpolation algorithm. Should be one of:
             cv2.INTER_NEAREST, cv2.INTER_LINEAR, cv2.INTER_CUBIC, cv2.INTER_AREA, cv2.INTER_LANCZOS4.
             Default: cv2.INTER_LINEAR.
@@ -1200,6 +1202,7 @@ class GridDistortion(DualTransform):
         self,
         num_steps=5,
         distort_limit=0.3,
+        limit_to_bbox=False,
         interpolation=cv2.INTER_LINEAR,
         border_mode=cv2.BORDER_REFLECT_101,
         value=None,
@@ -1210,6 +1213,9 @@ class GridDistortion(DualTransform):
         super(GridDistortion, self).__init__(always_apply, p)
         self.num_steps = num_steps
         self.distort_limit = to_tuple(distort_limit)
+        self.limit_to_bbox = (
+            (limit_to_bbox, limit_to_bbox) if isinstance(limit_to_bbox, bool) else tuple(limit_to_bbox)
+        )
         self.interpolation = interpolation
         self.border_mode = border_mode
         self.value = value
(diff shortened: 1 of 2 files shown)


poke1024:
Tests are failing, but this not seem related to this PR:

```
tests/test_augmentations.py:522: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
albumentations/core/transforms_interface.py:87: in __call__
    return self.apply_with_params(params, **kwargs)
albumentations/core/transforms_interface.py:100: in apply_with_params
    res[key] = target_function(arg, **dict(params, **target_dependencies))
albumentations/augmentations/transforms.py:2508: in apply
    return F.median_blur(image, ksize)
albumentations/augmentations/functional.py:54: in wrapped_function
    result = func(img, *args, **kwargs)
albumentations/augmentations/functional.py:752: in median_blur
    return blur_fn(img)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

img = array([[[0, 0, 0, 0, 0, 0],
        [0, 0, 0, 0, 0, 0],
        [0, 0, 0, 0, 0, 0],
        ...,
        [0, 0, 0, 0, ... 0, 0],
        ...,
        [0, 0, 0, 0, 0, 0],
        [0, 0, 0, 0, 0, 0],
        [0, 0, 0, 0, 0, 0]]], dtype=uint8)

def __process_fn(img):
        num_channels = get_num_channels(img)
        if num_channels > 4:
            chunks = []
            for index in range(0, num_channels, 4):
                chunk = img[:, :, index : index + 4]
>               chunk = process_fn(chunk, **kwargs)
E               cv2.error: OpenCV(4.2.0) /io/opencv/modules/imgproc/src/median_blur.simd.hpp:985: error: (-215:Assertion failed) src.depth() == CV_8U && (cn == 1 || cn == 3 || cn == 4) in function 'medianBlur'

albumentations/augmentations/functional.py:176: error
```

Dipet:
Hi! Thank you for your PR.
Tests found bug in our code. This bug will be fixed after merge #638

albumentations/augmentations/transforms.py
```diff
@@ -1224,8 +1230,12 @@ def apply_to_mask(self, img, stepsx=(), stepsy=(), **params):
...
 
     def get_params(self):
-        stepsx = [1 + random.uniform(self.distort_limit[0], self.distort_limit[1]) for i in range(self.num_steps + 1)]
-        stepsy = [1 + random.uniform(self.distort_limit[0], self.distort_limit[1]) for i in range(self.num_steps + 1)]
+        stepsx = [1 + random.uniform(self.distort_limit[0], self.distort_limit[1]) for i in range(self.num_steps)]
+        stepsy = [1 + random.uniform(self.distort_limit[0], self.distort_limit[1]) for i in range(self.num_steps)]
+        if self.limit_to_bbox[0]:
+            stepsx = np.array(stepsx) * max(1, self.num_steps / np.sum(stepsx))
+        if self.limit_to_bbox[1]:
+            stepsy = np.array(stepsy) * max(1, self.num_steps / np.sum(stepsy))
```
Dipet, line 1238: Why you changed `num_steps`? `num_steps` represents count of grids.

poke1024:
I'm closing this in favor of https://github.com/albumentations-team/albumentations/pull/722

PR #658, opened 2020-07-09
img aug version bumped

ternaus:


setup.py
```diff
@@ -5,7 +5,7 @@
 from pkg_resources import DistributionNotFound, get_distribution
 
 
-INSTALL_REQUIRES = ["numpy>=1.11.1", "scipy", "imgaug>=0.2.5,<0.2.7", "PyYAML"]
+INSTALL_REQUIRES = ["numpy>=1.11.1", "scipy", "imgaug==0.4.0", "PyYAML"]
```
creafz, line 8: Maybe we should use `"imgaug>=0.4.0"` (greater than or equal to comparison)? So next time, when the new version of ImgAug is released, we won't downgrade it during the installation.

Issue #673, opened 2020-07-31
Support passing 'sigma' as an argument for GaussianBlur


yaox12:
In recent self-supervised learning papers, they usually adopt the `GaussianBlur` augmentation with a random standard
deviation `sigma`. For example, in SimCLR [[code]](https://github.com/google-research/simclr) [[paper]](https://arxiv.org/abs/2002.05709), they randomly sample sigma in [0.1, 2.0]. I hope you can support this feature.

Dipet:
Unfortunately at this time you can not set range for sigma value. In current implementation sigma value computes from kernel size.
For example:
- `ksize=3, sigma=0.8`
- `ksize=5, sigma=1.1`
- `ksize=7, sigma=1.4`

yaox12:
@Dipet Thank you for your fast response!
I have read the updated code and I'm wondering whether we can use a **fixed kernel size** while **random sigma** with the new code? For example, in SimCLR, they use a fixed kernel size of 23 with sigma uniformly sampled from [0.1, 2.0].

Dipet:
Yes, you can set fixed kernel size like `blur_limit=[23, 23]`, it is docs issue.

Dipet:
The changes are merged.

PR #674, opened 2020-07-31
sigma_limit for GaussianBlur

Dipet:
Now we can set sigma range for GaussianBlur

#673


Dipet:
This PR similar to #536, but int #536 if sigma limits is set, then ksize will be computed from sigma.
We can do the same thing if we will set `blur_limit=(0, 0)` and `sigma != (0, value)`

PR #681, opened 2020-08-15
Add .readthedocs.yml

creafz:
Current automatic Readthedocs builds are broken and it seems that the only way to fix them is to use a custom config.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7]
+        python-version: [3.8]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


Dipet:
@creafz Can you check what a problem with spinx build?

creafz:
@Dipet Current builds on ReadTheDocs raise the `ModuleNotFoundError: No module named 'skbuild'` exception. https://readthedocs.org/projects/albumentations/builds/11667932/

According to [opencv-python's FAQ](https://github.com/skvark/opencv-python#frequently-asked-questions), the correct way to fix this error is to use pip >= 19.3. However, according to the output of the ReadTheDocs' build pipeline, it already uses the latest version of pip (20.2.2). Since there is little control over the build pipeline in a default configuration, it is challenging to debug the issue further. So I switched to `.readthedocs.yml` file for configuring a build pipeline.

PR #685, opened 2020-08-23
add FDA

gavrin-s:
Paper: https://openaccess.thecvf.com/content_CVPR_2020/papers/Yang_FDA_Fourier_Domain_Adaptation_for_Semantic_Segmentation_CVPR_2020_paper.pdf

Repo: https://github.com/YanchaoYang/FDA

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -2081,4 +2081,4 @@ def fourier_domain_adaptation(img, target_img, beta):
     src_image_transformed = np.fft.ifft2(amplitude_src * np.exp(1j * phase_src), axes=(0, 1))
     src_image_transformed = np.real(src_image_transformed)
 
-    return np.clip(src_image_transformed, 0, 255).astype(np.uint8)
+    return np.clip(src_image_transformed, 0, 255).astype(img.dtype)
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -2041,3 +2041,47 @@ def glass_blur(img, sigma, max_delta, iterations, dxy, mode):
             x[h, w], x[h + dy, w + dx] = x[h + dy, w + dx], x[h, w]
 
     return np.clip(cv2.GaussianBlur(x / coef, sigmaX=sigma, ksize=(0, 0)), 0, 1) * coef
+
+
+def fourier_domain_adaptation(img, target_img, beta):
+    """
+    Fourier Domain Adaptation from https://github.com/YanchaoYang/FDA
+
+    Args:
+        img:  source image
+        target_img:  target image for domain adaptation
+        beta: coefficient from source paper
+
+    Returns:
+        transformed image
+    """
+    # get fft of both source and target
+    fft_src = np.fft.fft2(img.astype(np.float32), axes=(0, 1))
+    fft_trg = np.fft.fft2(target_img.astype(np.float32), axes=(0, 1))
+
+    # extract amplitude and phase of both fft-s
+    amplitude_src, phase_src = np.abs(fft_src), np.angle(fft_src)
+    amplitude_trg = np.abs(fft_trg)
+
+    # mutate the amplitude part of source with target
+    amplitude_src_shifted = np.fft.fftshift(amplitude_src, axes=(0, 1))
+    amplitude_trg_shifted = np.fft.fftshift(amplitude_trg, axes=(0, 1))
+
+    height, width, _ = amplitude_src_shifted.shape
+    border = (np.floor(np.min((height, width)) * beta)).astype(int)
+    center_x, center_y = np.floor(height / 2.0).astype(int), np.floor(width / 2.0).astype(int)
+
+    y1, y2 = center_y - border, center_y + border + 1
+    x1, x2 = center_x - border, center_x + border + 1
+
+    amplitude_src_shifted[y1:y2, x1:x2] = amplitude_trg_shifted[y1:y2, x1:x2]
+    amplitude_src_mutated = np.fft.ifftshift(amplitude_src_shifted, axes=(0, 1))
+
+    # mutated fft of source
+    fft_src_mutated = amplitude_src_mutated * np.exp(1j * phase_src)
+
+    # get mutated image
+    src_image_transformed = np.fft.ifft2(fft_src_mutated, axes=(0, 1))
+    src_image_transformed = np.real(src_image_transformed)
+
+    return np.clip(src_image_transformed, 0, 255).astype(np.uint8)
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -2054,6 +2054,7 @@ def fourier_domain_adaptation(img, target_img, beta):
 
     Returns:
         transformed image
+
     """
     # get fft of both source and target
     fft_src = np.fft.fft2(img.astype(np.float32), axes=(0, 1))
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -2065,24 +2065,20 @@ def fourier_domain_adaptation(img, target_img, beta):
     amplitude_trg = np.abs(fft_trg)
 
     # mutate the amplitude part of source with target
-    amplitude_src_shifted = np.fft.fftshift(amplitude_src, axes=(0, 1))
-    amplitude_trg_shifted = np.fft.fftshift(amplitude_trg, axes=(0, 1))
-
-    height, width, _ = amplitude_src_shifted.shape
-    border = (np.floor(np.min((height, width)) * beta)).astype(int)
-    center_x, center_y = np.floor(height / 2.0).astype(int), np.floor(width / 2.0).astype(int)
+    amplitude_src = np.fft.fftshift(amplitude_src, axes=(0, 1))
+    amplitude_trg = np.fft.fftshift(amplitude_trg, axes=(0, 1))
+    height, width = amplitude_src.shape[:2]
+    border = np.floor(min(height, width) * beta).astype(int)
+    center_x, center_y = np.floor([height / 2.0, width / 2.0]).astype(int)
 
     y1, y2 = center_y - border, center_y + border + 1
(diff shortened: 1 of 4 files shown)


BloodAxe (approved):

Dipet:
Great work!
Can you add simple example how to use this transform and reference to article into docstring?
For many peoples this API is not clear.
Something like this:
```python
class FDA(ImageOnlyTransform):
    """
    Fourier Domain Adaptation from https://github.com/YanchaoYang/FDA
    Source paper "FDA: Fourier Domain Adaptation for Semantic Segmentation"
    Important: you need to pass target image as a parameter `target_image` in __call__

    Args:
        beta_limit (float or tuple of float): coefficient beta from paper. Recommended less 0.3.

    Targets:
        image

    Image types:
        3-channel uint8 images only

    Reference:
        https://github.com/YanchaoYang/FDA
        https://openaccess.thecvf.com/content_CVPR_2020/papers/Yang_FDA_Fourier_Domain_Adaptation_for_Semantic_Segmentation_CVPR_2020_paper.pdf

    Example:
        >>> import numpy as np
        >>> import albumentations as A
        >>> image = np.random.randint(0, 256, [100, 100, 3], dtype=np.uint8)
        >>> target_image = np.random.randint(0, 256, [100, 100, 3], dtype=np.uint8)
        >>> aug = A.Compose([A.FDA(p=1)])
        >>> result = aug(image=image, target_image=target_image)

    """
```

albumentations/augmentations/functional.py
```diff
@@ -2041,3 +2041,48 @@ def glass_blur(img, sigma, max_delta, iterations, dxy, mode):
...
+    fft_src = np.fft.fft2(img.astype(np.float32), axes=(0, 1))
+    fft_trg = np.fft.fft2(target_img.astype(np.float32), axes=(0, 1))
+
+    # extract amplitude and phase of both fft-s
+    amplitude_src, phase_src = np.abs(fft_src), np.angle(fft_src)
+    amplitude_trg = np.abs(fft_trg)
+
+    # mutate the amplitude part of source with target
+    amplitude_src_shifted = np.fft.fftshift(amplitude_src, axes=(0, 1))
+    amplitude_trg_shifted = np.fft.fftshift(amplitude_trg, axes=(0, 1))
```
Dipet, line 2069: Maybe there better to reuse `amplitude_src` and `amplitude_trg` to reduce memory usage?

albumentations/augmentations/functional.py
```diff
@@ -2041,3 +2041,48 @@ def glass_blur(img, sigma, max_delta, iterations, dxy, mode):
...
+
+    height, width, _ = amplitude_src_shifted.shape
+    border = (np.floor(np.min((height, width)) * beta)).astype(int)
```
Dipet, line 2072: We can use there built in `min` function, I think this code a little bit clearly:
```python
np.floor(min(height, width) * beta).astype(int)
```

albumentations/augmentations/functional.py
```diff
@@ -2041,3 +2041,48 @@ def glass_blur(img, sigma, max_delta, iterations, dxy, mode):
...
+    center_x, center_y = np.floor(height / 2.0).astype(int), np.floor(width / 2.0).astype(int)
```
Dipet, line 2073: Maybe better write this?
```python
center_x, center_y = np.floor([height / 2.0, width / 2.0]).astype(int)
```

albumentations/augmentations/functional.py
```diff
@@ -2041,3 +2041,48 @@ def glass_blur(img, sigma, max_delta, iterations, dxy, mode):
...
+    x1, x2 = center_x - border, center_x + border + 1
+
+    amplitude_src_shifted[y1:y2, x1:x2] = amplitude_trg_shifted[y1:y2, x1:x2]
+    amplitude_src_mutated = np.fft.ifftshift(amplitude_src_shifted, axes=(0, 1))
+
+    # mutated fft of source
+    fft_src_mutated = amplitude_src_mutated * np.exp(1j * phase_src)
+
+    # get mutated image
+    src_image_transformed = np.fft.ifft2(fft_src_mutated, axes=(0, 1))
```
Dipet, line 2085: Maybe we can to reduce the copies of images?
In many rows we store image copy that we use only in next row.
It is not necessary, but would be great if we will use in place operations and reduce optional copies of images. Memory operations are one of the slowest operations for system.

albumentations/augmentations/functional.py
```diff
@@ -2041,3 +2041,48 @@ def glass_blur(img, sigma, max_delta, iterations, dxy, mode):
...
+
+    # extract amplitude and phase of both fft-s
+    amplitude_src, phase_src = np.abs(fft_src), np.angle(fft_src)
+    amplitude_trg = np.abs(fft_trg)
+
+    # mutate the amplitude part of source with target
+    amplitude_src_shifted = np.fft.fftshift(amplitude_src, axes=(0, 1))
+    amplitude_trg_shifted = np.fft.fftshift(amplitude_trg, axes=(0, 1))
+
+    height, width, _ = amplitude_src_shifted.shape
```
Dipet, line 2071: Rewrite to:
```python
height, width = amplitude_src_shifted.shape[:2]
```

After that this transform will work with grayscale images. But maybe will be need to add check that images contains the same count of channels.

albumentations/augmentations/transforms.py
```diff
@@ -3405,3 +3406,65 @@ def get_transform_init_args_names(self):
...
+    def get_params_dependent_on_targets(self, params):
+        img = params["image"]
+        target_img = params["target_image"]
+        target_img = cv2.resize(target_img, img.shape[1::-1])
+
+        if target_img.shape[2] != img.shape[2]:
+            raise ValueError(
+                "The source and target images must contain the same number of channels,"
+                f" but given {img.shape[2]} and {target_img.shape[2]} respectively."
+            )
```
Dipet: I think better to check that image contains third dimension:
```python
if target_img.ndim != img.ndim or (target_img.ndim == 3 and target_img.shape[2] != img.shape[2]):
            raise ValueError(
                "The source and target images must contain the same number of channels,"
                f" but given {img.shape[2]} and {target_img.shape[2]} respectively."
            )
```

gavrin-s: done.

gavrin-s: I will check all dimensions `target_img.shape != img.shape`. Ok?

albumentations/augmentations/functional.py
gavrin-s: done.

albumentations/augmentations/functional.py
gavrin-s: done.

albumentations/augmentations/functional.py
gavrin-s: done.

albumentations/augmentations/functional.py
gavrin-s: done.

albumentations/augmentations/functional.py
gavrin-s: done.

albumentations/augmentations/transforms.py
Dipet: Oh yes, that's more correct!

albumentations/augmentations/transforms.py
```diff
@@ -3405,3 +3406,65 @@ def get_transform_init_args_names(self):
...
+    Important: you need to pass target image as a parameter `target_image` in __call__, see example
+
+    Args:
+        beta_limit (float or tuple of float): coefficient beta from paper. Recommended less 0.3.
+
+    Targets:
+        image
+
+    Image types:
+        3-channel uint8 images only
```
Dipet, line 3424: If we now support grayscale image we can remove this remark.

gavrin-s: done

Dipet: I think this transform can work with any channel images.

albumentations/augmentations/transforms.py
```diff
@@ -3405,3 +3406,65 @@ def get_transform_init_args_names(self):
...
+
+    """
+
+    def __init__(self, beta_limit=0.1, always_apply=False, p=0.5):
+        super(FDA, self).__init__(always_apply=always_apply, p=p)
+        self.beta_limit = to_tuple(beta_limit, low=0)
+
+    def apply(self, img, target_image=None, beta=0.1, **params):
+        if target_image is None:
+            return img
```
Dipet, line 3446: I think we do not need this check there. I think if we use this transform in out pipeline we always must provide target image.
Also this check is unnecessary because at first error will be raised in `get_params_dependent_on_targets`

gavrin-s: done

albumentations/augmentations/functional.py
```diff
@@ -2041,3 +2041,44 @@ def glass_blur(img, sigma, max_delta, iterations, dxy, mode):
...
+    x1, x2 = center_x - border, center_x + border + 1
+
+    amplitude_src[y1:y2, x1:x2] = amplitude_trg[y1:y2, x1:x2]
+    amplitude_src = np.fft.ifftshift(amplitude_src, axes=(0, 1))
+
+    # get mutated image
+    src_image_transformed = np.fft.ifft2(amplitude_src * np.exp(1j * phase_src), axes=(0, 1))
+    src_image_transformed = np.real(src_image_transformed)
+
+    return np.clip(src_image_transformed, 0, 255).astype(np.uint8)
```
Dipet, line 2084: If store image datatype at function start and change this line this transform will supports any datatype.
```python
return np.clip(src_image_transformed, 0, 255).astype(dtype)
```

Dipet: Do not forget after that change docstring about image types

Dipet:
Also, please, add transform to serialization tests

albumentations/augmentations/functional.py
gavrin-s: done

Dipet (approved):

PR #686, opened 2020-08-25
Added keypoints support to RandomSizedBBoxSafeCrop.

IlyaOvodov:
Added keypoints support to `RandomSizedBBoxSafeCrop`.
Also removed misleading unused parameters from `crop_keypoint_by_coords(...)`: it had the same signature as `crop_bbox_by_coords(...)`, but `crop_bbox_by_coords` scaled resulting bboxes to `(crop_height, crop_width)` while `crop_keypoint_by_coords` didn't, so parameters `crop_height` and `crop_width` were not only excessive but misleading.

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -200,11 +200,7 @@ class Crop(DualTransform):
     def apply_to_keypoint(self, keypoint, **params):
         return F.crop_keypoint_by_coords(
             keypoint,
-            crop_coords=(self.x_min, self.y_min, self.x_max, self.y_max),
-            crop_height=self.y_max - self.y_min,
-            crop_width=self.x_max - self.x_min,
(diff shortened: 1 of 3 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -953,6 +939,8 @@ def get_transform_init_args_names(self):
 
 class RandomSizedBBoxSafeCrop(DualTransform):
     """Crop a random part of the input and rescale it to some size without loss of bboxes.
+    Keypoints are processed but does not affect selection of crop region. So if you need to crop preserving
+    keypoints you have to pass fictive bbox enclosing keypoints.
```
Dipet, line 943: Thank you, great work! 
But I don't understand this clarification. Can you clarify this with an example?

ternaus:
Added

Issue #679, opened 2020-08-11
ValueError: Expected x_min for bbox (-0.02666666666666667, 0.3022222222222222, 0.10666666666666667, 0.48444444444444446, tensor(1)) to be in the range [0.0, 1.0], got -0.02666666666666667.


SlowMonk:
## To Reproduce

1.
1.
1.

## Environment

 - Albumentations version (e.g., 0.1.8):
 - Python version (e.g., 3.7): 3.7
 - OS (e.g., Linux): ubuntu 
 - How you installed albumentations (`conda`, `pip`, source): pip
 - Any other relevant information:

## Additional context

I'm getting ValueError: `Expected x_min for bbox (-0.02666666666666667, 0.3022222222222222, 0.10666666666666667, 0.48444444444444446, tensor(1)) to be in the range [0.0, 1.0], got -0.02666666666666667.`
 message  <br/>

my original box is 

`new_boxes->[[442.  79. 972. 564.]]`

its x_min, y_min, x_max, y_max format 
then it only give this error message when 
```

        if random.random() < 0.2:
            print('horizontalFlip')
            print(f'new_boxes->{new_boxes}')
            augment = albumentations.HorizontalFlip(p=0.5)
            aug = albumentations.Compose([augment],bbox_params={'format': 'pascal_voc', 'label_fields': ['labels']})
            aug_result = aug(image=np.array(new_image), bboxes=new_boxes, labels=labels)
            new_image = aug_result['image']
            new_boxes = aug_result['bboxes']
            new_image = FT.to_pil_image(new_image)
```

ternaus:
What is the shape of the image?

marcocaccin:
This unfortunately is regularly happening for me as well. Here's my env along with a reproducible script.

It seems to me that all of these errors seem to deal with numerical errors when doing geometric operations on totally legit bounding boxes right at the image boundary. 
One way to deal with the issue would be to allow an "unsafe" transformation by letting the user control the `check_validity` flag from `BBoxParams` and deal with truncations manually afterwards: that's still a better option than having a training job fail because of uncontrollable exceptions.

CC: @BeckerFelix

### Environment
- Albumentations version: 0.4.6, 0.3.3
- Python version: 3.6.12
- OS (e.g., Linux): ubuntu 20.04
- How you installed albumentations (conda, pip, source): pip

```python
import numpy as np
import albumentations as A

np.random.seed(123)
HEIGHT, WIDTH = 720, 1280

[...]

PR #688, opened 2020-08-28
Fixed rounding order of bbox values for yolo format

Dipet:
Issue: #679 

Fixed incorrect rounding order for yolo bbox format.

Example:
```python
# Current master
x, y, width, height = (1200.0, 367.5, 158.0, 703.0)
x_min, y_min, x_max, y_max = (1122.0, 17.5, 1280.0, 720.5)

# PR results
x, y, width, height = (1200.0, 367.5, 158.0, 703.0)
x_min, y_min, x_max, y_max = (1122, 17, 1280, 720)
```


creafz:
For all other bounding boxes formats, Albumentations returns augmented coordinates as floats. May we should also convert int coordinates to float before returning them to preserve the same output data type?

Dipet:
@creafz You a little bit confused me :)
Yolo format is normalized float format, so after transforms we always get floats.
As you can see after conversion to int I call `normalize_bbox` function that returns floats.

marcocaccin:
@creafz your suggestion can be a welcome addition, but it is out of scope for fixing this bug (and yes, yolo is float and normalised).

I think it's low on the priority list, but I do find myself having to query the data types of the albu inputs and restoring them afterwards - reason: albumentations transforms are performed inside a `tf.numpy_function`, which requires strict output types.
Alternatively, just use type annotations and specify that the output bboxes of albumentations will always be float64, that's fair as well

creafz (approved):

PR #699, opened 2020-09-10
Add support for mask target for CoarseDropout augmentation.

BloodAxe:
By settings non-None value for mask_fill_value, it will erase mask regions as well.
The default behavior of the CoarseDropout was not changed.

### DIFF
--- a/albumentations/core/transforms_interface.py
+++ b/albumentations/core/transforms_interface.py
@@ -141,6 +141,8 @@ class BasicTransform:
             params["interpolation"] = self.interpolation
         if hasattr(self, "fill_value"):
             params["fill_value"] = self.fill_value
+        if hasattr(self, "mask_fill_value"):
+            params["mask_fill_value"] = self.mask_fill_value
         params.update({"cols": kwargs["image"].shape[1], "rows": kwargs["image"].shape[0]})
         return params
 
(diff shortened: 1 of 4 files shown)


ternaus (approved):

albumentations/core/transforms_interface.py
```diff
@@ -141,6 +141,8 @@ def update_params(self, params, **kwargs):
             params["interpolation"] = self.interpolation
         if hasattr(self, "fill_value"):
             params["fill_value"] = self.fill_value
+        if hasattr(self, "mask_fill_value"):
+            params["mask_fill_value"] = self.mask_fill_value
```
Dipet, line 145: Are you sure that we need these changes? Looks like `CoarseDropout` can work with masks without these changes.

BloodAxe: There two ways to propagate params to apply functions - via "get_params" function in the class itself, or via this global "hack". For the sake of consistency with `fill_value` I thought to put it here. However we can discuss this decision :)

Dipet: Ok, I see about what you say. I think wee need some refactor and use same approach in all transforms, because today different transforms uses different methods.

Dipet (approved):

PR #700, opened 2020-09-11
Add new features to A.PadIfNeeded

BloodAxe:
This PR brings support of making image size dividable by `width_divisor` & `height_divisor`. This is helpful when model requires that image has size that is dividable by 32/64/128 etc.

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -3165,7 +3165,6 @@ class FancyPCA(ImageOnlyTransform):
     """
 
     def __init__(self, alpha=0.1, always_apply=False, p=0.5):
-        """"""
         super(FancyPCA, self).__init__(always_apply=always_apply, p=p)
         self.alpha = alpha
 
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -147,11 +147,8 @@ class PadIfNeeded(DualTransform):
                 h_pad_top = 0
                 h_pad_bottom = 0
         else:
-            if rows > self.pad_height_divisor:
-                pad_rows = rows % self.pad_height_divisor
-                pad_rows = self.pad_height_divisor - pad_rows if pad_rows > 0 else 0
-            else:
-                pad_rows = self.pad_height_divisor - rows
+            pad_remained = rows % self.pad_height_divisor
+            pad_rows = self.pad_height_divisor - pad_remained if pad_remained > 0 else 0
 
             h_pad_top = pad_rows // 2
             h_pad_bottom = pad_rows - h_pad_top
@@ -164,11 +161,8 @@ class PadIfNeeded(DualTransform):
                 w_pad_left = 0
                 w_pad_right = 0
         else:
-            if cols > self.pad_width_divisor:
-                pad_cols = cols % self.pad_width_divisor
-                pad_cols = self.pad_width_divisor - pad_cols if pad_cols > 0 else 0
-            else:
-                pad_cols = self.pad_width_divisor - cols
+            pad_remainder = cols % self.pad_width_divisor
+            pad_cols = self.pad_width_divisor - pad_remainder if pad_remainder > 0 else 0
 
             w_pad_left = pad_cols // 2
             w_pad_right = pad_cols - w_pad_left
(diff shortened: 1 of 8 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -3124,9 +3165,7 @@ class FancyPCA(ImageOnlyTransform):
     """
 
     def __init__(self, alpha=0.1, always_apply=False, p=0.5):
-        """
-
-        """
+        """"""
```
creafz, line 3168: Looks like this row could be safely deleted since it is an empty docstring.

BloodAxe: Done!

creafz (approved):

albumentations/augmentations/transforms.py
```diff
@@ -106,17 +109,27 @@ class PadIfNeeded(DualTransform):
...
         mask_value=None,
         always_apply=False,
         p=1.0,
     ):
+        if (min_height is None) == (pad_height_divisor is None):
+            raise ValueError("Only one of 'min_height' and 'pad_height_divisor' parameters must be set")
+
+        if (min_width is None) == (pad_width_divisor is None):
+            raise ValueError("Only one of 'min_width' and 'pad_width_divisor' parameters must be set")
+
```
Dipet, line 127: I think that better to compare boolean values using boolean functions
```python
if (min_height is None) and (pad_height_divisor is None):
    raise ValueError("Only one of 'min_height' and 'pad_height_divisor' parameters must be set")

if (min_width is None) and (pad_width_divisor is None):
    raise ValueError("Only one of 'min_width' and 'pad_width_divisor' parameters must be set")
```

albumentations/augmentations/transforms.py
```diff
@@ -126,19 +139,39 @@ def update_params(self, params, **kwargs):
...
+                h_pad_top = 0
+                h_pad_bottom = 0
         else:
-            h_pad_top = 0
-            h_pad_bottom = 0
+            if rows > self.pad_height_divisor:
+                pad_rows = rows % self.pad_height_divisor
+                pad_rows = self.pad_height_divisor - pad_rows if pad_rows > 0 else 0
+            else:
+                pad_rows = self.pad_height_divisor - rows
```
Dipet, line 154: I think we can simplify that code by remove condition, because first first branch works fine in all situations.
We need this condition if we want to speedup code, but I not sure that we achieve real speedup there.

albumentations/augmentations/transforms.py
```diff
@@ -126,19 +139,39 @@ def update_params(self, params, **kwargs):
...
+                w_pad_left = 0
+                w_pad_right = 0
         else:
-            w_pad_left = 0
-            w_pad_right = 0
+            if cols > self.pad_width_divisor:
+                pad_cols = cols % self.pad_width_divisor
+                pad_cols = self.pad_width_divisor - pad_cols if pad_cols > 0 else 0
+            else:
+                pad_cols = self.pad_width_divisor - cols
```
Dipet, line 171: Same thing as for rows.

albumentations/augmentations/transforms.py
```diff
@@ -126,19 +139,39 @@ def update_params(self, params, **kwargs):
...
+            h_pad_top = pad_rows // 2
+            h_pad_bottom = pad_rows - h_pad_top
+
+        if self.min_width is not None:
+            if cols < self.min_width:
+                w_pad_left = int((self.min_width - cols) / 2.0)
+                w_pad_right = self.min_width - cols - w_pad_left
+            else:
+                w_pad_left = 0
+                w_pad_right = 0
```
Dipet, line 165: Also I do not think that this if statement give to us real speedup, for this reason we can remove it for readability.

albumentations/augmentations/transforms.py
Dipet: Ok, I understand that `==` works here like `xor` function, that is more correct.

Dipet:
Also would be great to write tests for new functionality.

albumentations/augmentations/transforms.py
BloodAxe: This check is indeed needed. Without it, transform will work incorrectly when image is dimensions greater than min size. It this case it will return negative padding values.

Dipet (approved):

Issue #698, opened 2020-09-09
Is there a HueSaturationLightness (to recreate torchvision.transforms.ColorJitter)?


cgebbe:
I wanted to reproduce the behavior of torchvision.transforms.ColorJitter, similar to these existing threads
- https://github.com/albumentations-team/albumentations/issues/693
- https://github.com/albumentations-team/albumentations/issues/672
- https://github.com/albumentations-team/albumentations/issues/589

For now, I am only focusing on adjusting the **saturation** using albumentations in (nearly) the same way as using torchvision:
```
sat_shift = 0.5
torch_aug = torchvision.transforms.ColorJitter.get_params(saturation=sat_shift)
album_aug = A.HueSaturationValue(hue_shift_limit=0,
                                 sat_shift_limit=int(sat_shift * 255),
                                 val_shift_limit=0,
                                 p=1.0,
                                 ),
```
# Expected result
Augmented images are more or less the same

# Actual result
![test_aug_5](https://user-images.githubusercontent.com/4158176/92602131-05936b00-f2ae-11ea-92d7-ffb96e36e1a2.png)

- Torchvision seems to apply saturation shift in something like HSL space (they actually convert to black-white 'L' space and then blend using PIL, so not 100% sure whether it is really HSL space). Anyhow, white stays white
- In contrast, albumentations applies saturation shift in HSV space. Thereby, white can become a color (e.g. red)! I was actually quite surprised to see this. It does make total sense, since we are using HSV space. But this behavior is definitely different than changing the saturation slider in photoshop/GIMP.

# Conclusion
Is there any other transformation which changes saturation in the HSL space or HL-Chroma space similar to photoshop/GIMP?

Dipet:
We now have our version of `ColorJitter`, look to PR #705 fo details

Dipet:
Ooops, my fault, PR is not merged....

Dipet:
Finally we are added our implementation of `CollorJitter` that equal to torchvision transform.

PR #704, opened 2020-09-14
Added RandomColorShifts transform

Dipet:
Generalized version of `HueSaturationValue`. This transform can work with different color spaces. Also supported `random` color space, when we choose color space randomly.
 
Supported color spaces:
```python
class Colorspace(Enum):
        RANDOM = "random"
        HLS = "HLS"
        HSV = "HSV"
        LAB = "LAB"
        LUV = "LUV"
        YCrCb = "YCrCb"
        XYZ = "XYZ"
        YUV = "YUV"
```

This transform can resolve this issue #698


ternaus:
If it helps to resolve https://github.com/albumentations-team/albumentations/issues/698 we need a test that shows that for a set of the parameters CollorJitter and RandomColorShift match.

Dipet:
Thanks @ternaus for reminding me that we need a test to compare this transforms. `ColorJitter` works with image in different way.

sachit-menon:
Would changing H and S using HLS colorspace with this (composed with `RandomBrightnessContrast`) correspond to the perturbation `ColorJitter` does? A 1:1 comparison with that augmentation would be very useful for people hoping to migrate from torchvision.

Dipet:
I wrote our version of `ColorJitter` #705

albumentations/augmentations/functional.py
```diff
@@ -418,6 +418,62 @@ def shift_hsv(img, hue_shift, sat_shift, val_shift):
...
+
+        lut = np.arange(0, 256, dtype=np.int16)
+        if is_a:
+            lut = np.mod(lut + shift, limit[1]).astype(np.uint8)
+        else:
+            lut = np.clip(lut + shift, limit[0], limit[1]).astype(np.uint8)
+
+        cv2.LUT(channel, lut, dst=channel)
+
+    img = cv2.merge(img)
```
BloodAxe, line 437: Can we avoid cv2.split/cv2.merge by doing cv2.LUT inplace?  `cv2.LUT(img[..., channel], lut, dst=img[..., channel])` ?

BloodAxe (approved):

albumentations/augmentations/functional.py
Dipet: This doesn't work, because numpy is creating a new array.

Dipet: This version is slower.
Current version: 120 FPS
`img[..., i] = cv2.LUT(img[..., i], lut)`: 110 FPS

Dipet: Version `img[..., i] = lut[img[..., i]]` gives 60 FPS

Dipet:
I need some tests with `XYZ`, `YUV` and `YCrCb` colorspaces

ternaus (approved):

Dipet:
I close this PR, because at this time it is work different for uint8 and float32 data types

PR #705, opened 2020-09-14
Added ColorJitter transform

### LINKED ISSUE #698: Is there a HueSaturationLightness (to recreate torchvision.transforms.ColorJitter)?

Dipet:
This transform works like torchvision `ColorJitter`, but with some differences:
- Pillow and OpenCV works a little bit different with `HSV` color format.
- torchvision use `uint8` overflow, but I am use value saturation. It is important when factors greater than 1.

I do not know how to add comparison to torchvision into our tests, because for tests we need torchvision. Also in future behavior may change.
This test works fine:
```python
import random
import numpy as np
import albumentations as A
from PIL import Image
from torchvision.transforms import ColorJitter
from tqdm import tqdm


def check(img, brightness=1, contrast=1, saturation=1, hue=0):
    pil_image = Image.fromarray(img)

    transform = A.Compose([A.ColorJitter(brightness=[brightness, brightness], contrast=[contrast, contrast],
                                         saturation=[saturation, saturation], hue=[hue, hue], p=1)])
    pil_transform = ColorJitter(brightness=[brightness, brightness], contrast=[contrast, contrast],
                                saturation=[saturation, saturation], hue=[hue, hue])

    res1 = transform(image=img)['image']
    res2 = np.array(pil_transform(pil_image))

    _max = np.abs(res1.astype(np.int16) - res2.astype(np.int16)).max()

    if hue == 0:
        assert  _max <= 2, _max
    else:
        assert _max <= 255 / 10, _max


for i in tqdm(range(100)):
    img = np.random.randint(0, 256, [1000, 1000, 3], dtype=np.uint8)

    check(img, random.uniform(0, 1))
    check(img, 1, random.uniform(0, 1))
    check(img, 1, 1, random.uniform(0, 1))
    check(img, 1, 1, 1, random.uniform(-0.5, 0.5))
```

This PR will fix #698


ternaus (approved):

ternaus:
I do not remember, but what is the issue to add torchvision to tests?

Dipet:
In the past we especially removed Pillow from dependencies #370

Dipet:
Another problem that current transform is not fully equal to torchvisions `ColorJitter`

Dipet:
At first in OpencCV hue in range `[0, 180]` for `HSV` format. But for Pillow this range in `[0, 255]`
```python
import random
import numpy as np
import cv2 as cv
from PIL import Image

seed = 0
random.seed(seed)
np.random.seed(0)

image = np.random.randint(0, 256, [3, 1, 3], dtype=np.uint8)

res1 = cv.cvtColor(image, cv.COLOR_RGB2HSV)
res2 = np.array(Image.fromarray(image).convert('HSV'))

print("OpenCV:")
print(res1)
print()
print("Pillow")
print(res2)

# OpenCV:
# [[[158 240 172]]
# 
#  [[143 185 170]]
# 
#  [[ 13 103 196]]]
# 
# Pillow
# [[[224 240 172]]
# 
#  [[202 184 170]]
# 
#  [[ 18 102 196]]]
```

Dipet:
ok i think that's just one difference. I fixed some bugs in my code and now the differences between this implementation and torchvision version less than 2 for `uint8` for `brightness`, `saturation` and `contrast`

creafz (approved):

PR #722, opened 2020-10-15
An implementation of normalized_grid_distortion

poke1024:
This is a follow up of https://github.com/albumentations-team/albumentations/pull/637 with a new more focused implementation.

In contrast to #637, this implementation is based on the existing `grid_distortion`, i.e. it does not change its implementation of the `num_steps`.

Basing this functionality on the existing `grid_distortion` makes things a bit tricky, as the inner workings of `grid_distortion` are not super trivial. I've written a rationale of how this implementation comes up with its scaling factors in https://github.com/poke1024/albumentations-normalized-grid/blob/master/albumentations.pdf

I'm not a math whizz, and probably there is some more elegant way to tackle this, but I hope that these explanations make the implementation more clear.

A demo ipynb showing the new `NormalizedGridDistortion` functionality is here:

https://github.com/poke1024/albumentations-normalized-grid/blob/master/demo.ipynb

As you can see (on the right column at the graphics at the bottom), the new `NormalizedGridDistortion` will warp the image internally, but never project it outside its bounds. The current default `GridDistortion` is in the left column.

A standalone version of the notebook (not needing a patched albumentations library) is here: https://github.com/poke1024/albumentations-normalized-grid/blob/master/demo_standalone.ipynb


poke1024:
In the last PR this thing was implemented as `limit_bbox` parameter to `GridDistortion`, now it's a new class `NormalizedGridDistortion`. I'm not really sure which approach is better. I ended up creating a new class since I started implementing this outside of albumentations and deriving from `GridDistortion`.

Dipet (approved):

creafz (approved):

Issue #527, opened 2020-02-05
Speed Benchmark vs. Kornia


ogencoglu:
## 📚 Documentation

Would be nice to add _kornia_ to the benchmarks : https://kornia.readthedocs.io/en/latest/augmentation.html

ternaus:
Already there: https://albumentations.ai/docs/benchmarking_results/

PR #723, opened 2020-10-17
Add Kornia to benchmark

megaserg:
Added `kornia` library to benchmark suite, with a few of the augmentations. This was requested in #527.
Kornia operates on PyTorch Tensors, so I introduced a method to read images as tensors.
I'm actually not super familiar with Kornia or albumentations, I hope someone knowledgeable will implement more augmentations for comparison. I also might have used some operations incorrectly.

**Note**: nailed the version of `solt`: the version being installed by default is `0.1.9`, which had breaking changes from `0.1.8`, e.g. removed/renamed `solt.data` subpackage.

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -18,12 +18,15 @@ import cv2
 from tqdm import tqdm
 import numpy as np
 import pandas as pd
+import torch
+from torchvision import transforms
 import torchvision.transforms.functional as torchvision
 import keras_preprocessing.image as keras
 from imgaug import augmenters as iaa
 import solt.core as slc
 import solt.transforms as slt
 import solt.data as sld
+import kornia.augmentation as K
 
 import albumentations.augmentations.functional as albumentations
 import albumentations as A
@@ -38,7 +41,7 @@ os.environ["VECLIB_MAXIMUM_THREADS"] = "1"  # noqa E402
 os.environ["NUMEXPR_NUM_THREADS"] = "1"  # noqa E402
 
 
-DEFAULT_BENCHMARKING_LIBRARIES = ["albumentations", "imgaug", "torchvision", "keras", "augmentor", "solt"]
+DEFAULT_BENCHMARKING_LIBRARIES = ["albumentations", "imgaug", "torchvision", "keras", "augmentor", "solt", "kornia"]
 
 
 def parse_args():
@@ -69,6 +72,7 @@ def get_package_versions():
         "imgaug",
         "torchvision",
         "keras",
+        "kornia",
         "numpy",
         "opencv-python",
         "scikit-image",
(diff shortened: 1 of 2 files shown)


benchmark/benchmark.py
```diff
@@ -335,6 +361,19 @@ def keras(self, img):
...
+    #         "translations": torch.Tensor([50, 50]),
+    #         "center": torch.Tensor([[0, 0]]),
+    #         "scale": torch.Tensor([2]),
+    #         "sx": torch.Tensor([0.0]),
+    #         "sy": torch.Tensor([0.0]),
+    #         "resample": torch.Tensor([1]),  # BILINEAR
+    #         "align_corners": torch.tensor(False),
+    #     }
+    #     return K.functional.apply_affine(img, self.kornia_affine_params)
+
```
Dipet, line 376: Please remove this TODO

benchmark/benchmark.py
```diff
@@ -274,19 +293,26 @@ def torchvision_transform(self, img):
...
+    def kornia_op(self, img):
+        return K.functional.apply_adjust_brightness(img, self.kornia_brightness_params)
+
 
 class Contrast(BenchmarkTest):
     def __init__(self):
         self.imgaug_transform = iaa.Multiply((1.5, 1.5), per_channel=False)
         self.augmentor_op = Operations.RandomContrast(probability=1, min_factor=1.5, max_factor=1.5)
         self.solt_stream = slc.Stream([slt.ImageRandomContrast(p=1, contrast_range=(1.5, 1.5))])
+        self.kornia_contrast_params = {"contrast_factor": torch.Tensor([1.5])}
```
Dipet, line 305: better to use `torch.tensor`. `torch.Tensor` is not recommended to create tensors

creafz:
Hey, @megaserg.

Could you please add the following changes to PR?

1. Rebase the branch on `master`.  The master branch already contains fixes for working with the latest version of solt.
2. Add a command-line argument that specifies a device for Kornia (should be `cpu` by default). The benchmarking code should move PyTorch Tensors to the specified device.

benchmark/benchmark.py
Dipet: I think you can use this code. But need to check that it is get correct results
```python
    def kornia_op(self, img):
        self.kornia_affine_params = {
            "angle": torch.tensor([45.0]),
            "translations": torch.tensor([50.0, 50.0]),
            "center": torch.tensor([[0.0, 0.0]]),
            "scale": torch.tensor([2.0]),
            "sx": torch.tensor([0.0]),
            "sy": torch.tensor([0.0]),
            "resample": torch.tensor([1.0]),  # BILINEAR
            "align_corners": torch.tensor(False),
        }
        return K.functional.apply_affine(img, self.kornia_affine_params)
```

Dipet:
`BrightnessContrast`
```python
    def kornia_op(self, img):
        img = KF.apply_adjust_brightness(img, params={"brightness_factor": torch.tensor([1.5])})
        img = KF.apply_adjust_contrast(img, params={"contrast_factor": torch.tensor([1.5])})
        return img
```

Dipet:
If we will use `kornia==0.4.1` we can benchmark more transforms
`Posterize`
```python
    def kornia_op(self, img):
        return KF.apply_posterize(img, {"bits_factor": 4})
```

Dipet:
Closed as abandoned

PR #734, opened 2020-10-22
New interface for FDA transform and fix some bugs for HistogramMatching

Dipet:
- New interface for FDA transform, equal to `HistogramMatching`
- Fixed problems with grayscale images
- Predefined read functions
- Added transforms to tests
- Added exception on try to serialize this transforms

### DIFF
--- a/albumentations/augmentations/domain_adaptation.py
+++ b/albumentations/augmentations/domain_adaptation.py
@@ -35,7 +35,7 @@ def fourier_domain_adaptation(img: np.ndarray, target_img: np.ndarray, beta: flo
 
     if target_img.shape != img.shape:
         raise ValueError(
-            "The source and target images must contain the same shape,"
+            "The source and target images must have the same shape,"
             " but got {} and {} respectively.".format(img.shape, target_img.shape)
         )
 
--- a/albumentations/augmentations/domain_adaptation.py
+++ b/albumentations/augmentations/domain_adaptation.py
@@ -9,16 +9,9 @@ from .functional import clipped, preserve_shape
 
 from skimage.exposure import match_histograms
 
-__all__ = ["HistogramMatching", "FDA", "fourier_domain_adaptation", "read_bgr_image", "read_rgb_image"]
+from .utils import read_rgb_image
 
-
-def read_bgr_image(path):
-    return cv2.imread(path, cv2.IMREAD_COLOR)
-
-
-def read_rgb_image(path):
-    image = cv2.imread(path, cv2.IMREAD_COLOR)
-    return cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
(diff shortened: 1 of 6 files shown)


albumentations/augmentations/domain_adaptation.py
```diff
@@ -1,16 +1,28 @@
...
+from skimage.exposure import match_histograms
+
+__all__ = ["HistogramMatching", "FDA", "fourier_domain_adaptation", "read_bgr_image", "read_rgb_image"]
+
+
+def read_bgr_image(path):
+    return cv2.imread(path, cv2.IMREAD_COLOR)
+
+
+def read_rgb_image(path):
```
creafz, line 19: Maybe we could move those 2 functions to another module with generic functions such as `albumentations.augmentations.utils` since those functions are not tied to transforms in `domain_adaptation.py`?

albumentations/augmentations/domain_adaptation.py
```diff
@@ -80,21 +111,21 @@ class HistogramMatching(ImageOnlyTransform):
         uint8, uint16, float32
     """
 
-    def __init__(self, reference_images: List[str], blend_ratio=(0.5, 1.0), read_fn=cv2.imread, p=0.5):
-        super().__init__(p=p)
+    def __init__(
+        self,
+        reference_images: List[Union[str, np.ndarray]],
+        blend_ratio=(0.5, 1.0),
+        read_fn=read_bgr_image,
```
creafz, line 118: Maybe use `read_rgb_image` as a default `read_fn` both for `HistogramMatching` and `FDA`? I think most users are working with RGB images, and this will be a better default.

albumentations/augmentations/domain_adaptation.py
Dipet: done

albumentations/augmentations/domain_adaptation.py
Dipet: done

creafz (approved):

albumentations/augmentations/domain_adaptation.py
```diff
@@ -24,6 +29,16 @@ def fourier_domain_adaptation(img: np.ndarray, target_img: np.ndarray, beta: flo
         transformed image
 
     """
+
+    img = np.squeeze(img)
+    target_img = np.squeeze(target_img)
+
+    if target_img.shape != img.shape:
+        raise ValueError(
+            "The source and target images must contain the same shape,"
```
BloodAxe, line 38: A small improvement of the exception message "The source and target images must _have_ the same shape,"

BloodAxe (approved):

albumentations/augmentations/domain_adaptation.py
Dipet: done

PR #735, opened 2020-10-23
Update transforms for AutoAlbument support

creafz:
1. Add arguments to `ShiftScaleRotate` to separately set values for `shift_x` and `shift_y`.
2. Add the `Sequential` container that applies its augmentations in a sequence.
3. Add the `transpose_mask` argument to `ToTensorV2`. if it is set to `True` `ToTensorV2` will transpose not only dimensions of an image but also dimensions of an associated mask.
4. Add a function to register transforms from `albumentations.pytorch` for deserialization. The idea is that now transforms are registered for deserialization (so we can call A.load('some_yaml_or_json_file') to deserialize a pipeline) during import time. All transforms that are available through the `albumentations` module are registered for deserialization because those lines import transforms from their modules - https://github.com/albumentations-team/albumentations/blob/master/albumentations/__init__.py#L5-L9. But since we don't import transforms from `albumentations.pytorch` into the main module I need additional logic to register those transforms. As an alternative, I could add the import directly into https://github.com/albumentations-team/albumentations/blob/master/albumentations/__init__.py


albumentations/core/composition.py
```diff
@@ -474,3 +474,15 @@ def _to_dict(self):
...
+class Sequential(BaseCompose):
+    """Sequentially applies all transforms to targets."""
+
+    def __init__(self, transforms, p=0.5):
+        super().__init__(transforms, p)
+
+    def __call__(self, **data):
+        for t in self.transforms:
+            data = t(**data)
+        return data
```
Dipet, line 488: I am not sure that `DualTransforms` will work correctly with this wrapper, because we are do not convert bboxes and keypoints to our format.
Also this call doesn't contain preprocessing and postprocessing

creafz: Not sure if I understood you correctly. `Sequential` is not intended to be a replacement for `Compose`. Instead, it should be used inside `Compose` the same way `OneOf` or `OneOrOther` are used now. 

The main use case for AutoAlbument is the following: we have N sub-policies, each sub-policy contains M augmentations.
We want to choose one sub-policy at random and apply its augmentations in a sequence. With `Sequential` we can implement this pipeline. For example, here is a definition if a pipeline that contains 3 sub-policies and each subpolicy has 2 sequential augmentations:

```
transform = A.Compose([
    A.OneOf([
        A.Sequential([
            A.HorizontalFlip(p=0.5),
            A.ShiftScaleRotate(p=0.5),
        ]),
        A.Sequential([
            A.VerticalFlip(p=0.5),
            A.RandomBrightnessContrast(p=0.5),
        ]),
        A.Sequential([
            A.RGBShift(p=0.5),
            A.Cutout(p=0.5),
        ])
    ], p=1)
])
```

Dipet: I think we need write this tips in `Sequentional` documentation, that it is must be inside `Compose` and can not work alone.

creafz: Done

Dipet (approved):

ternaus (approved):

PR #736, opened 2020-10-26
Skip opencv requirement if it's already installed in the env (i.e. via conda)

daavoo:


sergeyk:
It would be very useful to avoid installing a different version of opencv if one already exists

Issue #752, opened 2020-11-14
ImportError: cannot import name 'match_histograms'


hhawary:
## 🐛 Bug
After installing Albumentations using conda i get this error `ImportError: cannot import name 'match_histograms'` when i try to import the library

### 2. `conda install -c conda-forge imgaug`

Collecting package metadata (current_repodata.json): done
Solving environment: done

==> WARNING: A newer version of conda exists. <==
  current version: 4.8.3
  latest version: 4.9.2

Please update conda by running

`$ conda update -n base -c defaults conda`

\## Package Plan ##

  environment location: /home/<username>/anaconda2/envs/wmlce_env

  added / updated specs:
    - imgaug

The following packages will be downloaded:

```
    package                    |            build
    ---------------------------|-----------------
    ca-certificates-2020.11.8  |       h1084571_0         145 KB  conda-forge
    certifi-2020.11.8          |   py36h270354c_0         150 KB  conda-forge
    geos-3.8.0                 |       he6710b0_0         1.1 MB
    imgaug-0.4.0               |             py_1         832 KB  conda-forge
    python_abi-3.6             |          1_cp36m           4 KB  conda-forge
    shapely-1.7.1              |   py36h98ec03d_0         394 KB
    ------------------------------------------------------------
                                           Total:         2.6 MB

```
The following NEW packages will be INSTALLED:

  geos               pkgs/main/linux-ppc64le::geos-3.8.0-he6710b0_0
  imgaug             conda-forge/noarch::imgaug-0.4.0-py_1
  python_abi         conda-forge/linux-ppc64le::python_abi-3.6-1_cp36m
  shapely            pkgs/main/linux-ppc64le::shapely-1.7.1-py36h98ec03d_0

The following packages will be UPDATED:

  ca-certificates    pkgs/main::ca-certificates-2020.10.14~ --> conda-forge::ca-certificates-2020.11.8-h1084571_0
  certifi               pkgs/main::certifi-2020.6.20-py36_0 --> conda-forge::certifi-2020.11.8-py36h270354c_0

Proceed ([y]/n)? y

Downloading and Extracting Packages
```
python_abi-3.6       | 4 KB      | ################################################################################################################################################################ | 100%
shapely-1.7.1        | 394 KB    | ################################################################################################################################################################ | 100%
ca-certificates-2020 | 145 KB    | ################################################################################################################################################################ | 100%
imgaug-0.4.0         | 832 KB    | ################################################################################################################################################################ | 100%
[...]

PR #757, opened 2020-11-15
Edit domain_adaptation for skimage support (#752)

hhawary:
1. Add support to histogram_matching in skimage version less than 0.16 related to issue (#752)
1. Following black format

### DIFF
--- a/albumentations/augmentations/domain_adaptation.py
+++ b/albumentations/augmentations/domain_adaptation.py
@@ -7,7 +7,10 @@ import numpy as np
 from ..core.transforms_interface import ImageOnlyTransform, to_tuple
 from .functional import clipped, preserve_shape
 
-from skimage.exposure import match_histograms
+try:
+    from skimage.exposure import match_histograms
+except ImportError:
+    from skimage.transform import histogram_matching
 
 from .utils import read_rgb_image
 


Dipet (approved):

ternaus:
What does this change do?

We do not use `histogram_matching` in the rest of the file.

albumentations/augmentations/domain_adaptation.py
```diff
@@ -7,7 +7,10 @@
 from ..core.transforms_interface import ImageOnlyTransform, to_tuple
 from .functional import clipped, preserve_shape
 
-from skimage.exposure import match_histograms
+try:
+    from skimage.exposure import match_histograms
+except ImportError:
+    from skimage.transform import histogram_matching
```
Dipet, line 13: Functions has different names

Dipet (changes requested):

albumentations/augmentations/domain_adaptation.py
hhawary: Yes you are right it supposed to be from skimage.transform import match_histograms . my mistake.

hhawary:
> What does this change do?
> 
> We do not use `histogram_matching` in the rest of the file.

please check https://github.com/scikit-image/scikit-image/blob/v0.15.x/skimage/transform/histogram_matching.py the function name is match_histograms

Starting from version 0.16.x they move this function match_histograms to this file https://github.com/scikit-image/scikit-image/blob/v0.16.x/skimage/exposure/histogram_matching.py

However they still have the same function in https://github.com/scikit-image/scikit-image/blob/v0.16.x/skimage/transform/histogram_matching.py

so for precautions i put it in a try except if in the future they remove it.

Dipet:
Everything is working fine today, so these changes are not necessary.

Issue #760, opened 2020-11-18
CropNonEmptyMaskIfExists with multiple masks not working


gatordevin:
Hi all,
I ran into an interesting issue. I was originally using RandomCrop to crop my images for instance segmentation with mask rcnn but, I ran into problems where I had images that lacked any instances after cropping. I saw the CropNonEmptyMaskIfExists and it seems to do what I want, correct me if I am wrong but it crops the image and guarantees there will be a mask present. Well, I used this and it gave me this error.
`AssertionError: CropNonEmptyMaskIfExists requires ['mask']`

My current lien of code that runs the composed transform function is as follows.
`transformed = self.__transformations(image=np.array(image, dtype=np.uint8), masks=masks)`
Since I am doing instance segmentation I have multiple masks and have to use the masks parameter instead of a mask. I believe the CropNonEmptyMaskIfExists has a problem with this that won't allow it to run on more than one mask. Or I could be missing something. Please let me know.

Here is my transform
`transformations = A.Compose([
    A.CropNonEmptyMaskIfExists(height=2000, width=2000),
    A.HorizontalFlip(p=0.5),
    A.RandomBrightnessContrast(p=0.2),
])`
Thank you

Dipet:
Thanks for information. We will try to resolve this problem us soon us possible.
At this time you may try to concatenate your mask along last axis to create single mask with shape `[Height, Width, NumMasks]` and provide it with keyword `mask`

Dipet:
Try latest version from repository `pip install git+https://github.com/albumentations-team/albumentations`. I think we fixed this problem

PR #761, opened 2020-11-19
Add masks keyword support to `CropNonEmptyMaskIfExists`

Dipet:
Added support of keyword `masks` to `CropNonEmptyMaskIfExists`. Now we can provide masks using both arguments: `mask` and `masks`.
In old version `mask` processed in `get_params_dependent_on_targets` function. Because using this approach we can't set condition to select argument now we update parameters in function `update_params`

Also I changed function `sum` to `any`, when we check that mask is not empty. `numpy.any` much more faster.

More info there #760

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -1116,12 +1116,7 @@ class CropNonEmptyMaskIfExists(DualTransform):
             cols=params["cols"],
         )
 
-    @property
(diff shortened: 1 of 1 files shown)


creafz (approved):

BloodAxe (approved):

BloodAxe:
Looks good to me. Approved.
@Dipet thanks for this fix!

PR #770, opened 2020-11-30
fix convert_bbox_to_albumentations when x_max > rows

iMeleon:
convert_bbox_to_albumentations([0.53125, 0.551471, 0.9375, 0.736928] , 'yolo, 608, 608, check_validity=True) == ValueError: Expected x_max for bbox (0.06414473684210527, 0.18421052631578946, 1.0016447368421053, 0.9210526315789473, 0) to be in the range [0.0, 1.0], got 1.0016447368421053.
Happens because in "convert_bbox_to_albumentations" after x_max = int(x_min + width), x_max == 609, but rows == 608.
 x_max = int(x_min + width) - 1 fix it.


Dipet:
Thank you for example when our code fails.
I think, that problem that some libraries adds to `x_min` and `y_min` 1 pixel, but another libraries don't do this.
I do not know how resolve this problem beautifully.
In my opinion this code will fix this problem:
```python
x, y, width, height = denormalize_bbox(bbox, rows, cols)

x_min = x - width / 2 + 1
x_max = x_min + width
y_min = y - height / 2 + 1
y_max = y_min + height

if x_max > rows:
    x_max -= 1
    x_min -= 1
    y_max -= 1
    y_min -= 1
```

Also would be great if you will add this case to test in `tests/test_bbox.py`

Dipet:

PR #775, opened 2020-12-03
Add support for float16 images

BloodAxe:


Dipet:
Are you sure that OpenCV supports float16 data type? If not, many transforms will fail with this data type.

BloodAxe:
Good point. I was too quick to bring this to table. Closing PR

PR #781, opened 2020-12-11
Fallback to mask shape if image is not passed

leotac:
It is sometimes useful to apply an augmentation only to a mask, and avoid wasting CPU time modifying the image.
In my experience, this seems to be the only piece of code that needs to be modified to support a mask-only augmentation.
Not sure if there's a better way to do it.

### DIFF
--- a/albumentations/core/transforms_interface.py
+++ b/albumentations/core/transforms_interface.py
@@ -143,7 +143,13 @@ class BasicTransform:
             params["fill_value"] = self.fill_value
         if hasattr(self, "mask_fill_value"):
             params["mask_fill_value"] = self.mask_fill_value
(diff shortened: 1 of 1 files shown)


Dipet:
Why you do not provide mask as image? I do not think that these changes are a good idea

sbebo:
Might be that this is not the correct way to go.
But transformations should not be applied in the same way to an image and to a mask, I can't pass the mask as an image.

Simple example: I have a ReplayCompose. I apply a bunch of augmentations to an image.
Later, I generate a mask for this image in some way (the mask was not available when I first augmented the image).
Then I want to replay the same augmentations on that mask. I can't simply pass the mask as an image.

Dipet:
We always need image as parameter for transform

PR #785, opened 2020-12-14
Add superpixels support

sjeffry-o:
in imgaug documentation they say, that they use skimage "slic" for function, but i can't reproduce result from imgaug function
maybe something is missed

there is things that arise with rewriting:
1. should we call new functions from imgaug-dependent ones(with "deprecated" warning) or should we completely destroy imgaug-dependent augs?
2. should we completely reproduce augmentation results of imgaug or, if this will be working faster and better, should we enhance options of usage(for example: "slic" function of skimage contains about 10 parameters, we can use them)

here picture with result for now
https://cdn1.savepice.ru/uploads/2020/12/14/eb600411f0e73d6531167d9b490483e7-full.png

passed tests and black formatting locally

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -11,6 +11,8 @@ from typing import Optional
 import cv2
 import numpy as np
 from skimage.measure import label
+from skimage.segmentation import slic
+from skimage.measure import regionprops
 
 from . import functional as F
 from .bbox_utils import denormalize_bbox, normalize_bbox, union_of_bboxes
@@ -83,6 +85,7 @@ __all__ = [
     "MaskDropout",
     "GridDropout",
     "ColorJitter",
+    "Superpixels",
 ]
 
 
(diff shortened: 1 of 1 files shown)


BloodAxe:
Hi! Thanks for your effort on helping us with imgaug.

I can give my 5cents with your first question:
> should we call new functions from imgaug-dependent ones(with "deprecated" warning) or should we completely destroy imgaug-dependent augs?

I'd say it's better to keep original names in the code & imports (But not in docs, he-he) and raise a warning:

```python
class Sharpen(A.ImageOnlyTransform):
   def __init__(...):
    ....
   ....

class IAASharpen(Sharpen):
   def __init__(...):
      warnings.warn("This augmentation is deprecated. Please use A.Sharpen instead").
```


Regarding second question - whenever possible we should keep the default behavior so the sake of backward compatibility. But exposing more parameters for more granular control of the augmentation is also highly appreciated!

albumentations/augmentations/transforms.py
```diff
@@ -3550,3 +3553,61 @@ def apply(self, img, transforms=(), **params):
...
+    Args:
+        p_replace (float): defines the probability of any superpixel area being replaced by the superpixel, i.e. by
+            the average pixel color within its area. Default: 0.1.
+        n_segments (int): target number of superpixels to generate. Default: 100.
+        p (float): probability of applying the transform. Default: 0.5.
+    Targets:
+        image
+    """
+
+    def __init__(self, p_replace=0.1, n_segments=100, always_apply=False, p=0.5):
```
Dipet, line 3570: Please give opportunity to set `n_segments` and `p_replace` as ranges.
Also add `max_size` argument support.

albumentations/augmentations/transforms.py
```diff
@@ -3550,3 +3553,61 @@ def apply(self, img, transforms=(), **params):
...
+        super(Superpixels, self).__init__(always_apply, p)
+        self.p_replace = p_replace
+        self.n_segments = n_segments
+
+    def apply(self, image, cols, rows):
+        segments = slic(image, n_segments=self.n_segments, compactness=5)
```
Dipet, line 3576: `n_segments` must be selected from the range.

albumentations/augmentations/transforms.py
```diff
@@ -3550,3 +3553,61 @@ def apply(self, img, transforms=(), **params):
...
+        nb_channels = image.shape[2]
```
Dipet, line 3577: This will not work with grayscale image

albumentations/augmentations/transforms.py
Dipet: Also maybe would be great to provide `compactness` as a transform argument.
In the latest imgaug version default value 10.

Dipet:
Thank you for your help.
What version of imgaug your implementation based on?
It looks like in latest imgaug this transform works a little bit different.

sjeffry-o:
Yes, in my enviroment it works also different
I'm trying to figure out why

All remarks have been notified, thanks!

albumentations/augmentations/transforms.py
sjeffry-o: why do we need n_segments and p_replace as ranges?

Dipet: Because if we set them as ranges we can randomly sample them. In imgaug we can set them as ranges, and in our wrapper we also can set them as ranges as well.

Dipet:
Implemented #849

Issue #783, opened 2020-12-13
get rid off imgaug dependency


vvv-tech:
There is page with transforms that albu don't do herself, need to fix that
https://albumentations.ai/docs/api_reference/imgaug/transforms/

Dipet:
As I see we need to rewrite only this transforms:
```
    "IAAEmboss",
    "IAASuperpixels",
    "IAASharpen",
    "IAAAffine",
    "IAAPiecewiseAffine",
    "IAAPerspective",
```

sjeffry-o:
#785 
gotcha superpixels augmentation

KiriLev:
I'd take Emboss if you don't mind

sjeffry-o:
sure

PR #786, opened 2020-12-14
Implemented Emboss transform from img_aug

KiriLev:
Implemented Emboss transform from img_aug as a part of #783

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -77,6 +77,7 @@ __all__ = [
     "GridDropout",
     "ColorJitter",
     "Sharpen",
+    "Emboss",
 ]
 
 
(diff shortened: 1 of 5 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -3107,3 +3108,46 @@ def apply(self, img, transforms=(), **params):
...
+        p (float): probability of applying the transform. Default: 0.5.
+
+    Targets:
+        image
+    """
+
+    def __init__(self, alpha=(0.2, 0.5), strength=(0.2, 0.7), always_apply=False, p=0.5):
+        super(Emboss, self).__init__(always_apply, p)
+        self.alpha = to_tuple(alpha, 0.0)
+        self.strength = to_tuple(strength, 0.0)
```
Dipet, line 3129: Invalid usage of `to_tuple`. Valid values for `alpha` in range `[0, 1]`, but if we will provide single value, for example `alpha=0.5` we will get `alpha=(-0.5, 0.5)`.
Correct usage: `to_tuple(alpha, low=0)`

Dipet: The same thing for `strength`. `strength` must be in range `[0, inf)`

albumentations/augmentations/transforms.py
```diff
@@ -3107,3 +3108,46 @@ def apply(self, img, transforms=(), **params):
...
+        self._emboss_matrix = self._generate_emboss_matrix()
+
+    def _generate_emboss_matrix(self):
+        alpha_sample = random.uniform(*self.alpha)
+        if not 0 <= alpha_sample <= 1.0:
+            raise ValueError(f"Expected 'alpha' to be in the interval [0.0, 1.0], got {alpha_sample}")
```
Dipet, line 3135: I think, that better to check `alpha` range after `to_tuple` method, when we save it's range

albumentations/augmentations/transforms.py
```diff
@@ -3107,3 +3108,46 @@ def apply(self, img, transforms=(), **params):
...
+                [0 - strength_sample, 1, 0 + strength_sample],
+                [0, 0 + strength_sample, 1 + strength_sample],
+            ],
+            dtype=np.float32,
+        )
+        matrix = (1 - alpha_sample) * matrix_nochange + alpha_sample * matrix_effect
+        return matrix
+
+    def apply(self, img, **params):
+        return cv2.filter2D(img, -1, self._emboss_matrix)
```
Dipet, line 3150: Better to implement this as function. Look to `motion_blur`, I think we can rename it to `convolution`

albumentations/imgaug/transforms.py
```diff
@@ -133,6 +134,7 @@ class IAAEmboss(ImageOnlyIAATransform):
     """
 
     def __init__(self, alpha=(0.2, 0.5), strength=(0.2, 0.7), always_apply=False, p=0.5):
+        warnings.warn("This augmentation is deprecated. Please use A.Emboss instead")
```
Dipet, line 137: I think better to replace whole class with this strings, because we are want to remove imgaug from dependencies.
```python
class IAAEmboss(Emboss):
    """Emboss the input image and overlays the result with the original image.
    This augmentation is deprecated. Please use Emboss instead

    Args:
        alpha ((float, float)): range to choose the visibility of the embossed image. At 0, only the original image is
            visible,at 1.0 only its embossed version is visible. Default: (0.2, 0.5).
        strength ((float, float)): strength range of the embossing. Default: (0.2, 0.7).
        p (float): probability of applying the transform. Default: 0.5.

    Targets:
        image
    """

    def __init__(self, alpha=(0.2, 0.5), strength=(0.2, 0.7), always_apply=False, p=0.5):
        warnings.warn("This augmentation is deprecated. Please use Emboss instead")
        super().__init__(alpha=alpha, strength=strength, always_apply=always_apply, p=p)
```

Dipet: @BloodAxe @ternaus @creafz What do you think about it?

BloodAxe: @Dipet your snippet is exactly what I'd write by myself. I think it's the most correct way to inform users while keep the compatibility with no added maintenance cost.

albumentations/augmentations/transforms.py
Dipet: I am wrong, it is correct usage because `low` the second argument.

albumentations/augmentations/transforms.py
BloodAxe: I think _emboss_matrix  should be generated for each new call, rather than being re-used. @KiriLev @Dipet

KiriLev: made same as in Sharpen

albumentations/augmentations/transforms.py
KiriLev: done

albumentations/imgaug/transforms.py
KiriLev: done

Dipet (approved):

ternaus (approved):

PR #787, opened 2020-12-14
Implemented Sharpen transform from img_aug

KiriLev:
Implemented Sharpen transform from img_aug as a part of #783

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -76,6 +76,7 @@ __all__ = [
     "MaskDropout",
     "GridDropout",
     "ColorJitter",
+    "Sharpen",
 ]
 
 
@@ -2092,7 +2093,7 @@ class MotionBlur(Blur):
     """
 
     def apply(self, img, kernel=None, **params):
-        return F.motion_blur(img, kernel=kernel)
+        return F.convolve(img, kernel=kernel)
 
     def get_params(self):
         ksize = random.choice(np.arange(self.blur_limit[0], self.blur_limit[1] + 1, 2))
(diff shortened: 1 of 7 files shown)


ternaus:
You also need to add the transforms to other tests like 

https://github.com/albumentations-team/albumentations/blob/master/tests/test_transforms.py

https://github.com/albumentations-team/albumentations/blob/master/tests/test_augmentations.py

https://github.com/albumentations-team/albumentations/blob/master/tests/test_serialization.py

etc

albumentations/augmentations/transforms.py
```diff
@@ -3107,3 +3108,40 @@ def apply(self, img, transforms=(), **params):
...
+        lightness ((float, float)): range to choose the lightness of the sharpened image. Default: (0.5, 1.0).
+        p (float): probability of applying the transform. Default: 0.5.
+
+    Targets:
+        image
+    """
+
+    def __init__(self, alpha=(0.2, 0.5), lightness=(0.5, 1.0), always_apply=False, p=0.5):
+        super(Sharpen, self).__init__(always_apply, p)
+        self.alpha = to_tuple(alpha, 0.0)
```
Dipet, line 3128: Invalid usage of to_tuple. Valid values for `alpha` in range `[0, 1]`, but if we will provide single value, for example `alpha=0.5` we will get `alpha=(-0.5, 0.5)`.
Correct usage: `to_tuple(alpha, low=0)`

albumentations/augmentations/transforms.py
```diff
@@ -3107,3 +3108,40 @@ def apply(self, img, transforms=(), **params):
...
+        self.lightness = to_tuple(lightness, 0.0)
```
Dipet, line 3129: The as `alpha`. `lightness` must be in range `[0, inf)`

albumentations/augmentations/transforms.py
```diff
@@ -3107,3 +3108,40 @@ def apply(self, img, transforms=(), **params):
...
+        self._sharpening_matrix = self._generate_sharpening_matrix()
+
+    def _generate_sharpening_matrix(self):
+        alpha_sample = random.uniform(*self.alpha)
+        if not 0 <= alpha_sample <= 1.0:
+            raise ValueError(f"Expected 'alpha' to be in the interval [0.0, 1.0], got {alpha_sample}")
```
Dipet, line 3135: Better to move checking borders to `__init__`

albumentations/augmentations/transforms.py
```diff
@@ -3107,3 +3108,40 @@ def apply(self, img, transforms=(), **params):
...
+        lightness_sample = random.uniform(*self.lightness)
+        matrix_nochange = np.array([[0, 0, 0], [0, 1, 0], [0, 0, 0]], dtype=np.float32)
+        matrix_effect = np.array([[-1, -1, -1], [-1, 8 + lightness_sample, -1], [-1, -1, -1]], dtype=np.float32)
+
+        matrix = (1 - alpha_sample) * matrix_nochange + alpha_sample * matrix_effect
+        return matrix
+
+    def apply(self, img, **params):
+        return cv2.filter2D(img, -1, self._sharpening_matrix)
```
Dipet, line 3144: Better to implement this as function. Look to `motion_blur`, I think we can rename it to `convolution`

albumentations/augmentations/transforms.py
Dipet: I am wrong, it is correct usage because low the second argument.

albumentations/augmentations/transforms.py
Dipet: I am wrong, it is correct usage because low the second argument.

albumentations/augmentations/transforms.py
BloodAxe: Same as in Emboss, _sharpening_matrix should be generated for each new call (see BrightnessContrastAdjust for instance)

KiriLev: fixed both

albumentations/augmentations/transforms.py
KiriLev: done

Dipet (approved):

albumentations/augmentations/transforms.py
```diff
@@ -3107,3 +3108,44 @@ def apply(self, img, transforms=(), **params):
...
+    def _generate_sharpening_matrix(self):
+        alpha_sample = random.uniform(*self.alpha)
+        lightness_sample = random.uniform(*self.lightness)
+        matrix_nochange = np.array([[0, 0, 0], [0, 1, 0], [0, 0, 0]], dtype=np.float32)
+        matrix_effect = np.array([[-1, -1, -1], [-1, 8 + lightness_sample, -1], [-1, -1, -1]], dtype=np.float32)
+
+        matrix = (1 - alpha_sample) * matrix_nochange + alpha_sample * matrix_effect
+        return matrix
+
+    def apply(self, img, **params):
```
BloodAxe, line 3146: So we are very close here to final version. One last step is to make this augmentation works with ReplayCompose.

To do this, one need to implement method get_params that would do random sampling and return deterministic parameters:

```
def get_params(self):
        return {
            "alpha": random.uniform(*self.alpha)
            "lightness": random.uniform(*self.lightness)
        }
```

apply method would look like:

```
 def apply(self, img, alpha=0., lightness=0., **params):
        sharpening_matrix = self._generate_sharpening_matrix(alpha,  lightness)
        return F.convolve(img, sharpening_matrix)
```

albumentations/augmentations/transforms.py
```diff
@@ -3107,3 +3108,45 @@ def apply(self, img, transforms=(), **params):
...
+
+    def _generate_sharpening_matrix(self, alpha_sample, lightness_sample):
+        matrix_nochange = np.array([[0, 0, 0], [0, 1, 0], [0, 0, 0]], dtype=np.float32)
+        matrix_effect = np.array([[-1, -1, -1], [-1, 8 + lightness_sample, -1], [-1, -1, -1]], dtype=np.float32)
+
+        matrix = (1 - alpha_sample) * matrix_nochange + alpha_sample * matrix_effect
+        return matrix
+
+    def apply(self, img, alpha=0.0, lightness=0.0, **params):
+        sharpening_matrix = self._generate_sharpening_matrix(alpha_sample=alpha, lightness_sample=lightness)
```
Dipet, line 3148: Might be better to create a matrix in `get_params`?

BloodAxe: For sake of speed in case of multi-image & replay processing - computing matrix in `get_params` would be faster. That was my suggestion to compute _generate_sharpening_matrix in `apply` call. @KiriLev would you please move it to `get_params`. The rest looks good to me.

KiriLev: moved!

Dipet (approved):

BloodAxe (approved):

PR #790, opened 2020-12-15
Implemented Perspective transform equal to imgaug

Dipet:
This implementation work correctly with masks that contain values differ to `{0; 1}`
#783

### DIFF
--- a/albumentations/augmentations/geometric/functional.py
+++ b/albumentations/augmentations/geometric/functional.py
@@ -332,14 +332,14 @@ def perspective(img, matrix, max_width, max_height, border_val, border_mode, kee
     return warped
 
 
-def perspective_bbox(bbox, h, w, matrix, max_width, max_height, keep_size):
+def perspective_bbox(bbox, h, w, matrix, max_width, max_height):
     bbox = denormalize_bbox(bbox, h, w)
 
     points = np.array(
         [[bbox[0], bbox[1]], [bbox[2], bbox[1]], [bbox[2], bbox[2]], [bbox[0], bbox[2]]], dtype=np.float32
     )
 
-    points = cv2.perspectiveTransform(points.reshape(1, 4, 2), matrix)
+    points = cv2.perspectiveTransform(points.reshape([1, 4, 2]), matrix)
 
     bbox = [np.min(points[..., 0]), np.min(points[..., 1]), np.max(points[..., 0]), np.max(points[..., 1])]
     return normalize_bbox(bbox, max_height, max_width)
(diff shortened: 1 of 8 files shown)


albumentations/augmentations/geometric/functional.py
```diff
@@ -310,3 +311,52 @@ def longest_max_size(img, max_size, interpolation):
 @preserve_channel_dim
 def smallest_max_size(img, max_size, interpolation):
     return _func_max_size(img, max_size, interpolation, min)
+
+
+@preserve_channel_dim
+def perspective(img, matrix, max_width, max_height, border_val, border_mode, keep_size, interpolation):
```
ternaus, line 317: Let's add types to the functions?

albumentations/augmentations/geometric/functional.py
```diff
@@ -310,3 +311,52 @@ def longest_max_size(img, max_size, interpolation):
...
+    )
+    warped = perspective_func(img)
+
+    if keep_size:
+        return resize(warped, h, w, interpolation=interpolation)
+
+    return warped
+
+
+def perspective_bbox(bbox, h, w, matrix, max_width, max_height):
```
BloodAxe, line 335: How about name h, w -> height, width?

BloodAxe (approved):

albumentations/augmentations/geometric/functional.py
Dipet: Added

albumentations/augmentations/geometric/functional.py
Dipet: Fixed

albumentations/augmentations/geometric/functional.py
Dipet: I don't have much experience with type hints, so if you check it out I will be grateful.

BloodAxe: Something like this would be fine to explain to newscomers what's expected as input
```
def perspective(img: np.ndarray, matrix: np.ndarray, max_width:int, max_height:int, border_val:int, border_mode:int, keep_size:bool, interpolation:int) -> np.ndarray:
```

Dipet: But `border_val` maybe be any type from this list: `Union[int, float, List[int], List[float], np.ndarray]`
You think that better to simplify construction and set only one type  - `int`?

BloodAxe: So you telling me you don't have much experience with type hints? ;) ^_^
Well, yes, for border_val we can use all that. OR. Even better solution - we can define alias BorderTypes = Union[int, float, List[int], List[float], np.ndarray] and use it everywhere for this type of argument.

Dipet: I just read the manual 😄 
I didn't know about types aliases.
I think is better to use the current solution in this PR. And when we refactor other transforms and functions we can add an alias.

ternaus (approved):

Issue #792, opened 2020-12-16
`MultiplicativeNoise(elementwise=True)` fails when image shape is (H,W,1)


mjkvaak:
Hi, 
first of all thanks for one of the most useful ML packages I have found so far! I noticed that `albumentations.MultiplicativeNoise` fails under these conditions:

* Image is unsqueezed grayscale with shape (H,W,1)
* `elementwise` parameter has been set to True

Below is the minimal reproducing example:

```python
h,w = 28,28
img = np.zeros((h,w))
tfm_img = albumentations.MultiplicativeNoise(elementwise=True, p=1)(image=img)
print(tfm_img['image'].shape) #prints (28,28) as it should

img = np.zeros((h,w,1))
tfm_img = albumentations.MultiplicativeNoise(elementwise=True, p=1)(image=img)
print(tfm_img['image'].shape) #prints (28,28,28) which is a bug

img = np.zeros((h,w,3))
tfm_img = albumentations.MultiplicativeNoise(elementwise=True, p=1)(image=img)
print(tfm_img['image'].shape) #prints (28,28,3) as it should
```

Also, if `h` and `w` are unequal, the script will crash with `ValueError`:
```python
h,w = 256,28
img = np.zeros((h,w,1))
tfm_img = albumentations.MultiplicativeNoise(elementwise=True, p=1)(image=img)

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-152-5abd2e633fe7> in <module>
      5 
      6 img = np.zeros((h,w,1))
----> 7 tfm_img = albumentations.MultiplicativeNoise(elementwise=True, p=1)(image=img)

...

ValueError: operands could not be broadcast together with shapes (256,30,1) (256,30) 
```

BloodAxe:
Thanks for your report. We will investigate this.

PR #793, opened 2020-12-16
Fixed MultiplicativeNoise for case with image shape [h, w, 1]

Dipet:
#792


albumentations/augmentations/transforms.py
```diff
@@ -2706,7 +2706,7 @@ def get_params_dependent_on_targets(self, params):
             shape = [c]
```
BloodAxe, line 2706: This comment is not related to the bugfix itself. It points to the possible bug in implementation, when per_channel = True. In this case, the condition check will set shape to `shape = [c]`, which I read as "apply same noise vector of [c] channels to all pixels". Probably, it should be `shape = [h,w,1]` to apply image-level noise to all channels instead.

ternaus (approved):

albumentations/augmentations/transforms.py
Dipet: @BloodAxe
I a little bit confused what you mean.
Do you mean that in your mind condition `per_channel` means that we need to generate noise with shape `[h, w]`?
In my mind condition `per_channel=True` means that we will use `c` equal to image `c` shape when we produce noise tensor, else we use `c=1`.
If `elementwise=True` we set `[h, w]` of noise vector equal to image `[h, w]`. If `elementwise=False` noise vector `h = w = 1`

BloodAxe (approved):

PR #811, opened 2020-12-28
RandomCropFromBorders

IDMIPPM:
Cuts out a rectangular portion of the image by randomly indenting each side. There is no resize of image at the end.


ZFTurbo:
It would be nice if someone approve this. ) I always starting update with adding this functionality into albumentation code...

Dipet:
Oh my god. I forgot about this PR. Sorry.
Need to update it is to add serialization support.
To do this need to implement `get_transform_init_args_names` method and add this transform to tests.
As example you can look at #888

Dipet:
Done in #1240

PR #831, opened 2021-01-31
Update linters

creafz:
- Pin the latest linter versions both in the pre-commit config and CI.
- Delete the `check-yaml` hook because there is no way to exclude certain YAML files from checking, and the linter produces warnings for the `conda.recipe/meta.yaml` file, which is in fact not a YAML file but a Jinja template that generates a YAML file.
- Fix mypy error for the `keypoint` variable inside `perspective_keypoint` (`albumentations/augmentations/geometric/functional.py:382: error: Incompatible types in assignment (expression has type "ndarray", variable has type "Union[List[int], List[float]]")`)


creafz:
Somehow DeepSource suddenly reported 115 new issues. Most of them are not even relevant to the code changed in this PR. After a brief look, some of those issues look very disputable. I think we should disable DeepSource completely until someone is willing to resolve all those issues manually. This will help to reduce the excessive amount of warnings for contributors who create new PRs.

Dipet (approved):

BloodAxe (approved):

PR #834, opened 2021-02-02
Implemented albumentations.core.composition.SomeOf

cm107:
Implemented A.SomeOf.
The intent is to create a class that is identical in functionality to imgaug's iaa.SomeOf.
The functionality of A.SomeOf is very similar to A.OneOf, which is already implemented, but allows for a random sampling of multiple augmentations.
Example:
A.SomeOf(sample_range=(3, 5), transforms) will randomly sample 3, 4, or 5 augmentations from transforms and apply them in a random order. The likelihood of any given transform being chosen is weighted based on transforms[i].p.
Note that A.SomeOf(sample_range=(1,1), transforms) is identical to A.OneOf(transforms).

### DIFF
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -1,5 +1,6 @@
 from __future__ import division
 from collections import defaultdict
+from typing import Tuple
 
 import random
 
@@ -274,28 +275,31 @@ class SomeOf(BaseCompose):
     Transforms probabilities will be normalized to one 1, so in this case transforms probabilities works as weights.
 
     Args:
-        sample_range (int, int): Range of possible sample sizes.
         transforms (list): list of transformations to compose.
+        sample_range (int, int): Range of possible sample sizes.
         p (float): probability of applying selected transform. Default: 0.5.
     """
 
-    def __init__(self, sample_range: (int, int), transforms, p=0.5):
+    def __init__(self, sample_range: Tuple[int], transforms, p=0.5):
         super().__init__(transforms, p)
         if isinstance(sample_range, (list, tuple)):
             self.sample_range = sample_range
         elif isinstance(sample_range, int):
             self.sample_range = (0, sample_range)
         else:
-            raise TypeError(f'Invalid type for sample_range: {type(sample_range).__name__}')
-        self.sample_range = sample_range
-        assert len(self.sample_range) == 2
-        assert self.sample_range[0] <= self.sample_range[1]
-        assert all([isinstance(val, int) for val in self.sample_range])
-        assert self.sample_range[1] <= len(transforms)
-    
+            raise TypeError(f"Invalid type for sample_range: {type(sample_range).__name__}")
+        if len(self.sample_range) != 2:
+            raise ValueError(f"len(sample_range) == {len(sample_range)} != 2")
+        if self.sample_range[0] > self.sample_range[1]:
+            raise ValueError(f"sample_range[0] == {self.sample_range[0]} > {self.sample_range[1]} == sample_range[1]")
+        if not all([isinstance(val, int) for val in self.sample_range]):
+            raise TypeError(f"not all([isinstance(val, int) for val in sample_range])")
+        if self.sample_range[1] > len(transforms):
+            raise ValueError(f"sample_range[1] == {self.sample_range[1]} > {len(transforms)} == len(transforms)")
+
     def _to_dict(self) -> dict:
         item_dict = super()._to_dict()
-        item_dict['sample_range'] = self.sample_range
+        item_dict["sample_range"] = self.sample_range
         return item_dict
 
     def __call__(self, force_apply=False, **data):
@@ -317,6 +321,7 @@ class SomeOf(BaseCompose):
             used_idx_list.append(t_idx)
             del unused_idx_list[unused_idx_list.index(t_idx)]
 
+        random.shuffle(used_idx_list)
         for idx in used_idx_list:
             t = self.transforms.transforms[idx]
             data = t(force_apply=True, **data)
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -13,7 +13,7 @@ from albumentations.core.utils import format_args, Params
 from albumentations.augmentations.bbox_utils import BboxProcessor
 from albumentations.core.serialization import SERIALIZABLE_REGISTRY, instantiate_lambda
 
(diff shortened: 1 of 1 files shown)


BloodAxe:
Greetings! Thanks for your PR. I believe this is good improvement on sampling augmentations. It seems there are couple of failing unit tests, would you mind fixing them following our [guidelines](https://github.com/albumentations-team/albumentations/blob/master/docs/contributing.rst)?

albumentations/core/composition.py
```diff
@@ -269,6 +269,61 @@ def __call__(self, force_apply=False, **data):
...
+    def __init__(self, sample_range: (int, int), transforms, p=0.5):
+        super().__init__(transforms, p)
+        if isinstance(sample_range, (list, tuple)):
+            self.sample_range = sample_range
+        elif isinstance(sample_range, int):
+            self.sample_range = (0, sample_range)
+        else:
+            raise TypeError(f'Invalid type for sample_range: {type(sample_range).__name__}')
+        self.sample_range = sample_range
+        assert len(self.sample_range) == 2
```
BloodAxe, line 291: We use exceptions instead of assertions to check input arguments for validity. We do use assertions in unit tests.

albumentations/core/composition.py
```diff
@@ -269,6 +269,61 @@ def __call__(self, force_apply=False, **data):
...
+        assert all([isinstance(val, int) for val in self.sample_range])
+        assert self.sample_range[1] <= len(transforms)
+    
+    def _to_dict(self) -> dict:
+        item_dict = super()._to_dict()
+        item_dict['sample_range'] = self.sample_range
+        return item_dict
+
+    def __call__(self, force_apply=False, **data):
+        if self.replay_mode:
```
BloodAxe, line 302: As a rule of thumb, generation of all parameters (sampling indexes) should be done in `get_params`. This makes the __call__ function deterministic (you simply get the indexes of transforms to apply) and compatible with A.ReplayCompose. Current implementation will not work in Replay mode.

albumentations/core/composition.py
```diff
@@ -269,6 +269,61 @@ def __call__(self, force_apply=False, **data):
         return data
 
 
+class SomeOf(BaseCompose):
+    """Select some of the transforms to apply. Selected transforms will be called with `force_apply=True`.
+    Transforms probabilities will be normalized to one 1, so in this case transforms probabilities works as weights.
+
+    Args:
+        sample_range (int, int): Range of possible sample sizes.
+        transforms (list): list of transformations to compose.
```
BloodAxe, line 278: Perhaps, it make sense to make `transforms` the first argument, to follow signature of existing transforms.

albumentations/core/composition.py
cm107: Fixed in latest commit

albumentations/core/composition.py
cm107: Fixed in latest commit

albumentations/core/composition.py
cm107: I'm sorry, I'm not quite familiar with how replay mode works or what its purpose is.
I looked at the ReplayCompose code as well as some of the documentation, and I gather that it is used to record information about which augmentations were applied and which parameters were used.
If I understand what you're saying correctly, I would need to make it so that the indexes of the applied transforms are recorded during replay mode. I am a bit confused, though, since the A.OneOf class doesn't seem to do this either.
I copied this if statement from the A.OneOf class.
Would creating a get_params method under A.SomeOf that returns {"used_idx_list": used_idx_list} and using that in __call__ satisfy your request?

albumentations/core/composition.py
```diff
@@ -269,6 +270,65 @@ def __call__(self, force_apply=False, **data):
...
+
+    Args:
+        transforms (list): list of transformations to compose.
+        sample_range (int, int): Range of possible sample sizes.
+        p (float): probability of applying selected transform. Default: 0.5.
+    """
+
+    def __init__(self, sample_range: Tuple[int], transforms, p=0.5):
+        super().__init__(transforms, p)
+        if isinstance(sample_range, (list, tuple)):
```
Dipet, line 285: Wrong `sample_range` type. Below you check that `sample_range` may be list. Better to use type `Sequence`.

Dipet: And better to check is `sample_range` is instance of `Sequence`

albumentations/core/composition.py
```diff
@@ -269,6 +270,65 @@ def __call__(self, force_apply=False, **data):
         return data
 
 
+class SomeOf(BaseCompose):
+    """Select some of the transforms to apply. Selected transforms will be called with `force_apply=True`.
+    Transforms probabilities will be normalized to one 1, so in this case transforms probabilities works as weights.
+
+    Args:
+        transforms (list): list of transformations to compose.
+        sample_range (int, int): Range of possible sample sizes.
```
Dipet, line 279: Please, describe logic for single integer as parameter.

albumentations/core/composition.py
Dipet: @BloodAxe Library uses another logic for `ReplayCompose`. `ReplayCompose` checks is `params` is `None` for each transformation. If `params` is `None` - this transformation is not called.
https://github.com/albumentations-team/albumentations/blob/ba44effb0369ba5eae1e8eb4909105eac9709230/albumentations/core/composition.py#L389

albumentations/core/composition.py
```diff
@@ -269,6 +270,65 @@ def __call__(self, force_apply=False, **data):
...
+        used_idx_list = []
+        for i in range(sample_size):
+            unused_transforms_ps = [self.transforms[idx].p for idx in unused_idx_list]
+            s = sum(unused_transforms_ps)
+            unused_transforms_ps_norm = [p / s for p in unused_transforms_ps]
+
+            random_state = np.random.RandomState(random.randint(0, 2 ** 32 - 1))
+            t_idx = random_state.choice(unused_idx_list, p=unused_transforms_ps_norm)
+            used_idx_list.append(t_idx)
+            del unused_idx_list[unused_idx_list.index(t_idx)]
```
Dipet, line 322: I think better to select transforms using [`np.random.choice`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.choice.html) with probabilities of each transformation.

Dipet:
Implemented in #889

PR #839, opened 2021-02-11
Add RandomToneCurve transform

aaroswings:
This adds a pixel-level RandomToneCurve augmentation to augmentations.transforms and its functional implementation in functional.move_tone_curve. It's based on the idea of the tone curve adjustment that exists in software like Lightroom. I couldn't find anything exactly like this in the library, but the concept is that it may help networks be more robust to different levels of dynamic range in different camera sensors, different image postprocessing, or relative lighting changes between shadow and light areas.

It changes the relationship of light and dark areas of the image by creating a cubic Bezier curve on the interval [0, 1] defined by four points at (0, 0), (.25, <random value>), (.75, <random value>), (1, 1), then evaluating the curve, mapping it to [0, 255] and rescaling pixel values of the image by the curve with a lookup table. This is similar to brightness/contrast adjustment, but has the subtle difference that the change is non-linear, so it changes the relative lighting of different areas. Supports uint8 images. 
Samples:


BloodAxe:
Hi! Many thanks for your first PR. The implementation itself looks very clear 💪 👍 
It seems there are some minor issues still remaining to fix before we can merge it:
- Please ensure that black and flake8 are happy with code formatting
- Re-generate docs of transformation list in README.md (See  `tools/make_transforms_docs.py`)

aaroswings:
Thanks for the warm response 😀 Glad you think this could be a useful contribution.
I ran black and flake8 on my branch, when I run `tools/make_transforms_docs.py make` I noticed that the class I added doesn't show in the outputs. Is there something I'm likely to be missing?

Dipet:
Hello!
Good work! Try to run `python tools/make_transforms_docs.py make` and copy output into README.md. I don't no why, but spatial-level transforms table is changed in your PR.

aaroswings:
Okay, sorted that out. Thanks!

albumentations/augmentations/functional.py
```diff
@@ -450,6 +450,39 @@ def equalize(img, mask=None, mode="cv", by_channels=True):
     return result_img
 
 
+def move_tone_curve(img, low_shift, high_shift):
```
Dipet, line 453: Please wrap this function with `@preserve_shape` because `cv2.LUT` may remove last channel from gray image

albumentations/augmentations/functional.py
```diff
@@ -450,6 +450,39 @@ def equalize(img, mask=None, mode="cv", by_channels=True):
...
+
+    # Defines responze of a four-point bezier curve
+    def evaluate_bez(t):
+        return 3 * (1 - t) ** 2 * t * low_shift + 3 * (1 - t) * t ** 2 * high_shift + t ** 3
+
+    evaluate_bez = np.vectorize(evaluate_bez)
+    remapping = np.rint(evaluate_bez(t) * 255).astype(np.uint8)
+
+    img = cv2.LUT(img, remapping)
+    return img
```
Dipet, line 483: Please wrap cv2.LUT with `_maybe_process_in_chunks` to support multichannel images
```python
lut_fn = _maybe_process_in_chunks(cv2.LUT, lut=remapping)
img = lut_fn(img)
```

albumentations/augmentations/transforms.py
```diff
@@ -1252,7 +1365,44 @@ def get_params_dependent_on_targets(self, params):
...
+            "num_shadows_upper",
+            "shadow_dimension",
+        )
+
+
+class RandomToneCurve(ImageOnlyTransform):
+    """Randomly change the relationship between bright and dark areas of the image by manipulating its tone curve.
+
+    Args:
+        scale (float): Amount of random deviation from the original image tone.
```
Dipet, line 1380: Please describe valid value range for this argument.

albumentations/augmentations/transforms.py
```diff
@@ -1252,7 +1365,44 @@ def get_params_dependent_on_targets(self, params):
...
+
+    Targets:
+        image
+
+    Image types:
+        uint8
+    """
+
+    def __init__(
```
Dipet, line 1389: Add add function `get_transform_init_args_names` and describe `scale` as init argument name. It is necessary because all functions must be serializable.
Also add this transform into these tests:
- `tests/test_serialization.py`:
    - `test_augmentations_serialization`
    - `test_augmentations_serialization_with_custom_parameters`
    - `test_additional_targets_for_image_only_serialization`
 - `tests/test_augmentations.py`:
     - `test_image_only_augmentations`
     - `test_augmentations_wont_change_input`
     - `test_augmentations_wont_change_shape_grayscale`
     - `test_augmentations_wont_change_shape_rgb`
     - `test_multichannel_image_augmentations`
     - `test_multichannel_image_augmentations_diff_channels`

Dipet (changes requested):

Dipet (approved):

aaroswings:
[I made a small notebook](https://nbviewer.jupyter.org/github/aaroswings/RandomToneCurveTests/blob/main/RandomToneCurveTests.ipynb) to explore what the outliers look like for a handful of random samples and different values of "scale", fwiw

creafz (approved):

ternaus (approved):

PR #843, opened 2021-02-15
Implementation of IAACropAndPad

Dipet:
Implemented `IAACropAndPad`. I didn't rewrite `IAACropAndPad`, I just added `DeprecationWarning`, because this implementation is exactly the same - not all padding modes are supported and bboxes and keypoints results may differ.


creafz (approved):

albumentations/augmentations/crops/functional.py
```diff
@@ -201,3 +204,81 @@ def clamping_crop(img: np.ndarray, x_min: int, y_min: int, x_max: int, y_max: in
...
         x_max = w - 1
     return img[int(y_min) : int(y_max), int(x_min) : int(x_max)]
+
+
+@preserve_channel_dim
+def crop_and_pad(
+    img: np.ndarray,
+    crop_params: Sequence[int],
+    pad_params: Sequence[int],
+    pad_value: float,
```
BloodAxe, line 214: Maybe `pad_value: Union[int, float]`, ?

BloodAxe (approved):

albumentations/augmentations/crops/functional.py
Dipet: I thought about this. Mypy can not work with `Number` super class. @creafz said to me that for mypy is common approach to use `float` where we work both on `float` and `int` types. But I agree that `Union[int, float]` more clearer.

Dipet: changed

PR #844, opened 2021-02-16
Deprecated IAAFliplr, IAAFlipud, IAAAdditiveGaussianNoise

Dipet:
- Added deprecation warnings to:
    - IAAFliplr
    - IAAFlipud
    - IAAAdditiveGaussianNoise
- Added logic to check whole links and table to docs. It is needed to remove deprecated transforms from README.
- Added `per_channel` flag to `GaussNoise` because it is exists in IAAAdditiveGaussianNoise
- Fixed unserialized `mean`


creafz:
I propose two changes that could be either implemented in this PR or in a new PR:
1. Use [`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning) instead of `DeprecationWarning` for imgaug augmentations. [`DeprecationWarning` is suppressed by default](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) so most users won't see messages about deprecated augmentations.
2. Explicitly name the augmentation in a warning. So use `IAAFliplr is deprecated...` instead of `This augmentation is deprecated...`. In that case, users will know which augmentation is causing problems without a need to dig into the source code.

creafz (approved):

creafz (approved):

BloodAxe (approved):

PR #849, opened 2021-03-03
Superpixels from imgaug

Dipet:
Implementation of Superpixels transform from imgaug


ternaus (approved):

albumentations/augmentations/functional.py
```diff
@@ -6,7 +6,9 @@
 from itertools import product
 import cv2
 import numpy as np
+import skimage
```
creafz, line 9: I think it is better to explicitly define `scikit-image` in `install_requires` in `setup.py` since we plan to use it after we remove imgaug as a dependency. - https://github.com/albumentations-team/albumentations/blob/master/setup.py#L8.

Dipet: But it is already in the dependencies, as you can see from your link. Or what you mean?

creafz: Sorry, my bad. I somehow managed to miss it.

creafz (approved):

BloodAxe (approved):

Issue #856, opened 2021-03-16
Perspective transforms keypoints inconsistently


sporadicCactus:
## 🐛 Bug

The perspective transform applied to the image differs from the transform applied to the keypoints.
Example:
<img src="https://user-images.githubusercontent.com/51837750/111279308-9ecf7f80-864b-11eb-9f1e-f16eb4b0ff50.jpg" width="400" height="200">

## To Reproduce

```python
def show_image_with_keypoints(image, keypoints):
    image = image.copy()
    for pt in keypoints:
        pt = tuple(int(c) for c in pt)
        cv2.circle(image, pt, 1, (0, 255, 0), 2)
    plt.imshow(image)

height, width = 100, 100

image = np.zeros((height, width, 3), dtype=np.uint8)

x_coords = np.linspace(0, width, 10)[1:-1].astype(np.int)
y_coords = np.linspace(0, height, 10)[1:-1].astype(np.int)

for x in x_coords:
    cv2.line(image, (x, 0), (x, height), (255, 255, 255), 1)
    
for y in y_coords:
    cv2.line(image, (0, y), (width, y), (255, 255, 255), 1)

keypoints = np.stack(
    [
        np.tile(x_coords[:, None], (1, len(y_coords))),
        np.tile(y_coords[None, :], (len(x_coords), 1))
    ],
    axis=-1
).reshape(-1, 2)

transform = A.Compose([A.Perspective()], keypoint_params=A.KeypointParams(format='xy'))
tr = transform(image=image, keypoints=keypoints)

plt.figure(figsize=(16, 8))
plt.subplot(121)
show_image_with_keypoints(image, keypoints)
plt.subplot(122)
show_image_with_keypoints(tr["image"], tr["keypoints"])
plt.show()
```

## Expected behavior
In the given example the keypoints should match line intersections after the transform.

## Environment

 - Albumentations 0.5.2
 - Python 3.7.9
 - Ubuntu 18.04
 - OpenCV 4.5.1
 - Albumentations installed from git

Dipet:
Hello!
Thank you for information. Looks like a problem with the transform functions of OpenCV and skimage.
OpenCV and skimage give the same artifacts.
Unfortunately I don't know how to solve this problem because it depends on the behavior of the backend functions.

Dipet:
Oh, no. Look like problem inside our code. I need more time to check it is.

PR #859, opened 2021-03-22
Fix Perspective keypoints resize

Dipet:
- Fixed forgotten keypoints resize for `Perspective` transform when `keep_size=True`.
- Fixed incorrect rescaling of keypoints
Issue: #856


BloodAxe (approved):

BloodAxe:
Should we also add a unit-test to this fix to help in future?

creafz (approved):

Dipet:
Update:
- Fixed bboxes transformation. Previous work incorrectly
- Added test when `keep_size=True`

creafz (approved):

ternaus (approved):

PR #870, opened 2021-04-08
Fix collapsed Perspective transform

Dipet:
Sometimes the perspective transformation showed incorrect results because the points were not ordered correctly, and sometimes the two vertices had the same coordinates.

Code to reproduce, original image, wrong transform and fixed transform result:
```python
import numpy as np
import albumentations as A
import random
import cv2 as cv

seed = 0
random.seed(seed)
np.random.seed(seed)

img = cv.imread("D:\\img.png")
img = cv.cvtColor(img, cv.COLOR_BGR2RGB)

scale = 0.08
res = A.Compose([A.Perspective(scale=(scale, scale), p=1)])(image=img)["image"]

plt.suptitle("Fixed branch")
plt.subplot(211, title="original")
plt.imshow(img, vmin=0, vmax=255)
plt.subplot(212, title="res")
plt.imshow(res, vmin=0, vmax=255)

plt.show()
```
img
image
image

### DIFF
--- a/albumentations/augmentations/geometric/transforms.py
+++ b/albumentations/augmentations/geometric/transforms.py
@@ -351,7 +351,7 @@ class Perspective(DualTransform):
 
     @classmethod
     def _order_points(cls, pts: np.ndarray) -> np.ndarray:
-        pts = sorted(pts, key=lambda x: x[0])
+        pts = np.sort(pts, axis=0)
         left = pts[:2]  # points with smallest x coordinate - left points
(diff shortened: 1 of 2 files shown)


ternaus:
Looks good, but we need some test.

BloodAxe (approved):

Dipet:
@ternaus Tests added

BloodAxe (approved):

creafz (approved):

PR #885, opened 2021-04-28
Implementation of Affine transform

Dipet:
Implementation of `Affine` transformation. Works the same as `IAAAffine`. Implementation does not support the `skimage` backend because in my case it works the same, but requires complex changes to support it is.

Example with default values
image


creafz (approved):

BloodAxe (approved):

albumentations/augmentations/geometric/functional.py
```diff
@@ -410,3 +417,74 @@ def perspective_keypoint(
         return keypoint_scale((x, y, angle, scale), scale_x, scale_y)
 
     return x, y, angle, scale
+
+
+def _is_identity_matrix(matrix: skimage.transform.ProjectiveTransform, eps: float = 1e-4) -> bool:
+    identity = np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]], dtype=np.float32)
```
BloodAxe, line 423: np.eye(3)?

Issue #884, opened 2021-04-27
Cannot instantiate augmentations with tuple parameters from Hydra config


senarvi:
## 🐛 Bug

I'd like to use [Hydra](https://hydra.cc/) to configure and instantiate augmentations. It works nicely, unless I try to use a tuple as a parameter value. Hydra translates lists and tuples to an internal type that can be casted to a tuple, but Albumentations checks the parameter types and won't try to cast it if it's not a list.

## To Reproduce

It's convenient to define augmentations in a Hydra configuration like this:

```yaml
augmentations:
- _target_: albumentations.HorizontalFlip
  p: 0.5
- _target_: albumentations.HueSaturationValue
  sat_shift_limit:
  - -10
  - 20
```

One can then instantiate the augmentations like this:

```python
from omegaconf import DictConfig
import hydra
from hydra.utils import instantiate

@hydra.main(config_path='conf', config_name='config')
def my_app(cfg: DictConfig) -> None:
    augmentations = [instantiate(a) for a in cfg.augmentations]

if __name__ == "__main__":
    my_app()
```

If you run the above code, it will correctly instantiate the `HorizontalFlip` augmentation, but fails to instantiate the `HueSaturationValue`. The reason is that `instantiate()` translates the `sat_shift_limit` value [-10, 20] to an `omegaconf.ListConfig`. It would be possible to cast `omegaconf.ListConfig` to a tuple, but [to_tuple()](https://github.com/albumentations-team/albumentations/blob/95a048d342cbe748e5acf15cb1a28611a6303885/albumentations/core/transforms_interface.py#L36) checks that the type is either a list or a tuple before casting. It's not, so instead of casting it throws an exception:

    ValueError: Argument param must be either scalar (int, float) or tuple

## Expected behavior

`to_tuple()` should cast the `omegaconf.ListConfig` to a tuple, so that `instantiate()` would instantiate the `HueSaturationValue` transform.

## Environment

 - Albumentations version (e.g., 0.1.8): 0.5.2
 - Python version (e.g., 3.7): 3.8.3
 - OS (e.g., Linux): Linux
 - How you installed albumentations (`conda`, `pip`, source): conda

## Additional context

I understand that it's impossible to check for `omegaconf.ListConfig` and all other possible cases that can be casted to a tuple. Actually it's common to use "duck typing" in Python, meaning that anything that acts like a list can be provided in place of a list, so it's problematic to check the exact type of parameters.

BloodAxe:
[...]

PR #886, opened 2021-04-28
Use the more general type `Sequence` inside `to_tuple` checks

Dipet:
In some cases, users or libraries, may provide types different to `list` or `tuple` for the `to_tuple` method, but these arguments remain valid.
To prevent `ValueError` in these cases, I replaced check `isinstance(param, (list, tuple))` with `isinstance(param, Sequence)`.

These changes will fix issue #884

### DIFF
--- a/albumentations/core/transforms_interface.py
+++ b/albumentations/core/transforms_interface.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 
 import random
 from warnings import warn
+from typing import Sequence
 
 import cv2
 from copy import deepcopy
@@ -33,7 +34,7 @@ def to_tuple(param, low=None, bias=None):
             param = -param, +param
         else:
             param = (low, param) if low < param else (param, low)
-    elif isinstance(param, (list, tuple)):
+    elif isinstance(param, Sequence):
         param = tuple(param)
     else:
         raise ValueError("Argument param must be either scalar (int, float) or tuple")


albumentations/core/transforms_interface.py
```diff
@@ -33,7 +34,7 @@ def to_tuple(param, low=None, bias=None):
             param = -param, +param
         else:
             param = (low, param) if low < param else (param, low)
-    elif isinstance(param, (list, tuple)):
+    elif isinstance(param, Sequence):
```
BloodAxe, line 37: Both, `tuple` and `omegaconf.ListConfig` pass this check, yet I wonder if `isinstance(x, collections.abc.Sequence)` is more correct way to express the contract. I used to think of typing package just as of hints, rather real types.

BloodAxe (approved):

albumentations/core/transforms_interface.py
Dipet: In many cases them give the same results. In this case `typing.Sequence` is alias to `collections.abc.Sequence`

creafz (approved):

ternaus (approved):

albumentations/core/transforms_interface.py
omry: I believe string is a Sequence as well. I suggest collections.MutableSequence.

PR #889, opened 2021-05-01
NOf generalizing OneOf

henrique:


Dipet:
Good PR, thank you!
Because you added 2 new arguments into pipeline, please, update `_to_dict` method and add add new arguments into it.
It is needed for serialization support. To check that it is work please add new transform into these tests:
- file `test_serialization.py`
    - `test_transform_pipeline_serialization`
    - `test_transform_pipeline_serialization_with_bboxes`
    - `test_transform_pipeline_serialization_with_keypoints`

henrique:
Thanks @Dipet good point!
I couldn't find a neat way to get OneOf to serialise been a subclass of NOf, so reverted it to the original.

Dipet (approved):

BloodAxe:
A small suggestion regarding name of this transform - instead `NOf`, can we use `SomeOf`. To me, `SomeOf`' is more appealing in just how it looks in the code.

henrique:
> A small suggestion regarding name of this transform - instead `NOf`, can we use `SomeOf`. To me, `SomeOf`' is more appealing in just how it looks in the code.

Thanks @BloodAxe 
done

creafz (approved):

PR #899, opened 2021-05-19
Alias ToTensor to ToTensorV2.

Dipet:
Old `ToTensor` renamed to `ToTensorOld`

### DIFF
--- a/albumentations/pytorch/transforms.py
+++ b/albumentations/pytorch/transforms.py
@@ -38,7 +38,7 @@ def mask_to_tensor(mask, num_classes, sigmoid):
     return torch.from_numpy(mask)
 
 
-class ToTensor(BasicTransform):
+class ToTensorOld(BasicTransform):
     """Convert image and mask to `torch.Tensor` and divide by 255 if image or mask are `uint8` type.
     WARNING! Please use this with care and look into sources before usage.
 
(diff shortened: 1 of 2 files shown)


albumentations/pytorch/transforms.py
```diff
@@ -116,3 +114,6 @@ def get_transform_init_args_names(self):
 
     def get_params_dependent_on_targets(self, params):
         return {}
+
+
+ToTensor = ToTensorV2
```
creafz, line 119: I propose to remove this alias. I think that implicit behavior change for ToTensor is dangerous and can break users' pipelines. 

In my opinion, a better way is to have two classes:
- `ToTensorOld` (which is previously was known as `ToTensor`).
- `ToTensorV2`.

So users who have `ToTensor` will receive an exception and explicitly change class for the `ToTensor[Old, V2]` transform.

Dipet: In your case I don't know why we need to change current state.
We will remove `ToTensor`, but `ToTensorV2` will remain, which is a little confusing.
We now have a warning that `ToTensor` is deprecated. Maybe it's better to just remove `ToTensor` from `__all__`?

creafz: In my opinion, it is better to remove the functionality of `ToTensor` completely but do not alter the behavior of `ToTensorV2` or ToTensor (by aliasing `ToTensorV2` to `ToTensor`). 

Here is my proposal - raise a RuntimeError when someone tries to use `ToTensor`. The error will contain instructions on how to fix the problem. - https://github.com/albumentations-team/albumentations/compare/remove_to_tensor?expand=1

In that case, we won't silently change the behavior for users that still have `ToTensor` in their pipelines - they will receive an explicit exception with instructions on how to fix it. And because `ToTensor` will not contain any functionality all new users will have to import `ToTensorV2` into their pipelines. 

There is still an open question about naming, but I think it is ok to have an actual transform named `ToTensorV2` and not `ToTensor.`

Dipet: Ok, I think your implementation with `RuntimeError` is the best compromise.

PR #912, opened 2021-05-27
Proposal for yolo bbox conversion

Dimfred:
I noticed a heavy discrepancy between the original yolo labels and labels returned after an augmentation. 
E.g. when doing a resize along an axis like `LongestMaxAxis` the input labels of yolo should be kept! But due to the conversion  to `int` and back to `float` pixels are shifted. 

This is a proposal and I haven't fixed the tests yet and probably won't, but opening this to drive a bit more attention to that. 
Cheers


Dipet:

Issue #883, opened 2021-04-25
Question about YOLO format in `convert_bbox_to_albumentations`


christian-cahig:
Hi, I was looking through this part of `convert_bbox_to_albumentations` :
https://github.com/albumentations-team/albumentations/blob/95a048d342cbe748e5acf15cb1a28611a6303885/albumentations/augmentations/bbox_utils.py#L232-L244

Suppose the call `x, y, width, height = denormalize_bbox(bbox, rows, cols)` yields
- `x` = `y` = 4.0
- `width` = `height` = 3.0

for `rows` = `cols` = 5. This would give `x_min` a value of 3.0 and `x_max` a value of 6.0, *i.e.*, `x_max` is greater than `cols`. The current implementation seems to shift the bbox 1 pixel to the right (and 1 pixel downwards).

With that said, would it be better to replace the definitions of `x_min` and `x_max` (and similarly for `y_min` and `y_max`) as follows?
```
x_min = max(int(x - width / 2), 0)
x_max = min(int(x + width / 2 + 1), cols)
```
Having the `+ 1` in the calculation of `x_max` instead of `x_min` enlarges the bbox by *at most* 2 pixels to the left and to the right but ensures that the object is still enclosed by the bbox. The `max(..., 0)` and `min(..., cols)` ensure that `x_min` and `x_max` are within acceptable values.

christian-cahig:
> With that said, would it be better to replace the definitions of `x_min` and `x_max` (and similarly for `y_min` and `y_max`) as follows?
> 
> ```
> x_min = max(int(x - width / 2), 0)
> x_max = min(int(x + width / 2 + 1), cols)
> ```
> 
> Having the `+ 1` in the calculation of `x_max` instead of `x_min` enlarges the bbox by _at most_ 2 pixels to the left and to the right but ensures that the object is still enclosed by the bbox. The `max(..., 0)` and `min(..., cols)` ensure that `x_min` and `x_max` are within acceptable values.

There's a flaw in this idea: the bounding boxes will only get larger for every transformation. This might not be a good idea especially when using a lot of transformations.

Dimfred:
Why do the `denormalize` and the `normalize` in the first step? 
yolo => aulbumentations can just be transformed by:

    x, y, w, h = bbox # from yolo
    w_half, h_half = w / 2, h / 2
   
    x_min, x_max = x - w_half, x + w_half
    y_min, y_max = y - h_half, y + h_half
    
    return

[...]

PR #916, opened 2021-05-28
yolo bbox format

Gorogorov:
As mentioned several times in Pull Requests, conversion from ```yolo``` format to ```albumentations``` format and back loses accuracy. This is due to the conversion of coordinates from float to integer and back. Since the ```yolo``` format is already normalized, there is no need to convert anything to integer. I removed the unnecessary conversions and edited one of the tests.

I was not able to reproduce this bug on an image of size 100 by 100, so I added tests with images of other sizes that the current version of the library does not pass.


Dipet:
More information there #883

Dipet:

Issue #922, opened 2021-06-15
ValueError: Expected x_max for bbox to be in the range [0.0, 1.0], got 1.0025.


atikapane:
## 🐛 Bug

I tried to transform images using RandomResizedCrop with its bounding box (yolo format). But, there are images that when the program tried to transform the bounding box, I got error saying that the bounding box value (after transformation) is not in the range of 0-1. I already checked the original bounding box, and there is nothing wrong with the original bounding box value.

This is my pipeline
```
transform = A.Compose([
    A.RandomResizedCrop(608, 608),
], bbox_params = A.BboxParams(format='yolo'))
```

This is the code to place the bounding box to an array and transform it
```
    bboxes = []
    for dt in data:
        ann = []

        # Split string to float
        c, x, y, w, h = map(float, dt.split(' '))
        ann.append(x)
        ann.append(y)
        ann.append(w)
        ann.append(h)
        ann.append(c)

        bboxes.append(ann)

    transformed = transform(image=img, bboxes=bboxes)
```
And this is the error
```
  File "C:\Users\Atika Pane\AppData\Roaming\Python\Python38\site-packages\albumentations\core\composition.py", line 180, in __call__
    p.preprocess(data)
  File "C:\Users\Atika Pane\AppData\Roaming\Python\Python38\site-packages\albumentations\core\utils.py", line 62, in preprocess
    data[data_name] = self.check_and_convert(data[data_name], rows, cols, direction="to")
  File "C:\Users\Atika Pane\AppData\Roaming\Python\Python38\site-packages\albumentations\core\utils.py", line 70, in check_and_convert
    return self.convert_to_albumentations(data, rows, cols)
  File "C:\Users\Atika Pane\AppData\Roaming\Python\Python38\site-packages\albumentations\augmentations\bbox_utils.py", line 51, in convert_to_albumentations
    return convert_bboxes_to_albumentations(data, self.params.format, rows, cols, check_validity=True)
  File "C:\Users\Atika Pane\AppData\Roaming\Python\Python38\site-packages\albumentations\augmentations\bbox_utils.py", line 302, in convert_bboxes_to_albumentations
    return [convert_bbox_to_albumentations(bbox, source_format, rows, cols, check_validity) for bbox in bboxes]
  File "C:\Users\Atika Pane\AppData\Roaming\Python\Python38\site-packages\albumentations\augmentations\bbox_utils.py", line 302, in <listcomp>
[...]

Issue #862, opened 2021-03-31
Yolo  Bboxes will generate tags larger than 1.


sang-yc:
Bboxes will generate tags larger than 1 when using 'albumentations' for data augmentation. Why? What should be done?

The error is as follows:
ValueError: Expected x_max for bbox (0.284, 0.15315315315315314, 1.002, 0.993993993993994, 6.0) to be in the range [0.0, 1.0], got 1.002.

sang-yc:
I have found a solution to this problem.
Before data augmentation, the length and width of GT bboxes are subtracted by one pixel, which has little effect on the label bounding box, but it can prevent the overflow of values and prevent the generation of labels greater than 1.


image.shape[0]:  width of image
image.shape[1]:  length of image
bboxes: format yolo 
              [[x, y, w, h, class], [x, y, w, h, class], ...]

`for i in range(len(bboxes)):
            bboxes[i][2] = np.abs(bboxes[i][2] - 0.5 / image.shape[0])
            bboxes[i][3] = np.abs(bboxes[i][3] - 0.5 / image.shape[1])`

To further reduce the impact of processing, I only subtracted 0.5 pixels from the code.

sang-yc:
I have found a solution to this problem.
Before data augmentation, the length and width of GT bboxes are subtracted by one pixel, which has little effect on the label bounding box, but it can prevent the overflow of values and prevent the generation of labels greater than 1.


image.shape[0]:  width of image
image.shape[1]:  length of image
bboxes: format yolo 
              [[x, y, w, h, class], [x, y, w, h, class], ...]
```python
for i in range(len(bboxes)):
            bboxes[i][2] = np.abs(bboxes[i][2] - 0.5 / image.shape[0])
            bboxes[i][3] = np.abs(bboxes[i][3] - 0.5 / image.shape[1])
```
To further reduce the impact of processing, I only subtracted 0.5 pixels from the code.

sang-yc:
I have found a solution to this problem.
Before data augmentation, the length and width of GT bboxes are subtracted by one pixel, which has little effect on the label bounding box, but it can prevent the overflow of values and prevent the generation of labels greater than 1.

labels format: yolo
image.shape[0]:  width of image
image.shape[1]:  length of image
bboxes: [[x, y, w, h, class], [x, y, w, h, class], ...]

```python
        for i in range(len(bboxes)):
            bboxes[i][2] = np.abs(bboxes[i][2] - 0.5 / image.shape[0])
[...]

Issue #848, opened 2021-03-03
Yolo boxes format overflowing image and thowing an exception


Trud09:
## 🐛 Bug

I have some bounding boxes in yolo format that are off the image.

## To Reproduce

1. Have boxes outside image in yolo format and try to augment

\lib\site-packages\albumentations\augmentations\bbox_utils.py", line 330, in check_bbox
    "to be in the range [0.0, 1.0], got {value}.".format(bbox=bbox, name=name, value=value)
ValueError: Expected x_max for bbox (0.9967206790123457, 0.5379050925925926, 1.0001929012345678, 0.5529513888888888, 0) to be in the range [0.0, 1.0], got 1.0001929012345678.

## Expected behavior

I would prefer it to clip the boxes, or at least give me the option to do so.

## Environment

 - Albumentations version (e.g., 0.1.8): 0.5.2
 - Python version (e.g., 3.7): 3.6
 - OS (e.g., Linux): Windows
 - How you installed albumentations (`conda`, `pip`, source): pip
 - Any other relevant information:

ternaus:
It is not a bug. It is a feature.

It is much easier to pre-process labels once before model training, rather than catching the challenging corner cases.

Feel free to add one line that clips the bounding boxes to your dataloader :)

Trud09:
Actually upon further investigation, there is a bug somewhere.
When I convert the boxes on my own from 'yolo' to 'albumentations' format, it caps at 1, and does not throw an exception. So somewhere it is adding a small amount, maybe in rounding somewhere, causing it to be greater than 1.

ternaus:
Could you please provide an example that reproduces the behavior?

Trud09:
The image might be proprietary, but here is one box causing the issue:

Yolo Format:
[0.99662423, 0.7520255 , 0.00675154, 0.01446759]

My albumentations conversion:
[0.99324846, 0.7447917, 1.0, 0.7592593]

Exception for this box:
Exception has occurred: ValueError
Expected x_max for bbox (0.9934413580246914, 0.7450810185185185, 1.0001929012345678, 0.7592592592592593, 0) to be in the range [0.0, 1.0], got 1.0001929012345678.

My conversion code:
def yolo_to_norm_voc(bboxes):
    x_mean, y_mean, w, h = bboxes[:, 0], bboxes[:, 1], bboxes[:, 2], bboxes[:, 3]

    x_min = x_mean - (w/2)
    y_min = y_mean - (h/2)

[...]

PR #924, opened 2021-06-17
YOLO format without normalization and denormalization

Dipet:
Since `yolo` and `albumentations` are normalized formats, we don't need to normalize and denormalize the values in the conversion step.
The previous approach gave round-off errors.

These changes should fix the following issues:
- #922
- #903
- #862
- #883
- #848
- #679

### DIFF
--- a/albumentations/augmentations/bbox_utils.py
+++ b/albumentations/augmentations/bbox_utils.py
@@ -233,20 +233,23 @@ def convert_bbox_to_albumentations(bbox, source_format, rows, cols, check_validi
         # https://github.com/pjreddie/darknet/blob/f6d861736038da22c9eb0739dca84003c5a5e275/scripts/voc_label.py#L12
         bbox, tail = bbox[:4], tuple(bbox[4:])
         _bbox = np.array(bbox[:4])
-        if np.any((_bbox <= 0) | (_bbox > 1)):
-            raise ValueError("In YOLO format all labels must be float and in range (0, 1]")
+        if check_validity and np.any((_bbox <= 0) | (_bbox > 1)):
+            raise ValueError("In YOLO format all coordinates must be float and in range (0, 1]")
 
-        x, y, width, height = denormalize_bbox(bbox, rows, cols)
+        x, y, w, h = bbox
 
-        x_min = int(x - width / 2 + 1)
-        x_max = int(x_min + width)
-        y_min = int(y - height / 2 + 1)
-        y_max = int(y_min + height)
+        w_half, h_half = w / 2, h / 2
+        x_min = x - w_half
+        y_min = y - h_half
+        x_max = x_min + w
+        y_max = y_min + h
     else:
         (x_min, y_min, x_max, y_max), tail = bbox[:4], tuple(bbox[4:])
 
     bbox = (x_min, y_min, x_max, y_max) + tail
-    bbox = normalize_bbox(bbox, rows, cols)
+
+    if source_format != "yolo":
+        bbox = normalize_bbox(bbox, rows, cols)
     if check_validity:
         check_bbox(bbox)
     return bbox
@@ -280,20 +283,21 @@ def convert_bbox_from_albumentations(bbox, target_format, rows, cols, check_vali
         )
     if check_validity:
         check_bbox(bbox)
-    bbox = denormalize_bbox(bbox, rows, cols)
+
+    if target_format != "yolo":
+        bbox = denormalize_bbox(bbox, rows, cols)
     if target_format == "coco":
         (x_min, y_min, x_max, y_max), tail = bbox[:4], tuple(bbox[4:])
         width = x_max - x_min
         height = y_max - y_min
         bbox = (x_min, y_min, width, height) + tail
     elif target_format == "yolo":
-        # https://github.com/pjreddie/darknet/blob/f6d861736038da22c9eb0739dca84003c5a5e275/scripts/voc_label.py#L12
         (x_min, y_min, x_max, y_max), tail = bbox[:4], bbox[4:]
-        x = int((x_min + x_max) / 2 - 1)
-        y = int((y_min + y_max) / 2 - 1)
-        width = x_max - x_min
-        height = y_max - y_min
-        bbox = normalize_bbox((x, y, width, height) + tail, rows, cols)
+        x = (x_min + x_max) / 2.0
+        y = (y_min + y_max) / 2.0
+        w = x_max - x_min
+        h = y_max - y_min
+        bbox = (x, y, w, h) + tail
     return bbox
 
 
--- a/tests/test_bbox.py
+++ b/tests/test_bbox.py
@@ -77,8 +77,11 @@ def test_calculate_bbox_area(bbox, rows, cols, expected):
         ((20, 30, 40, 50, 99), "coco", (0.2, 0.3, 0.6, 0.8, 99)),
         ((20, 30, 60, 80), "pascal_voc", (0.2, 0.3, 0.6, 0.8)),
         ((20, 30, 60, 80, 99), "pascal_voc", (0.2, 0.3, 0.6, 0.8, 99)),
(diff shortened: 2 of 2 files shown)


glenn-jocher:
@Dipet @ternaus what's the current status of the YOLO normalization bug with Albumentations?

I have a YOLOv5 Albumentations PR in progress in https://github.com/ultralytics/yolov5/pull/3882, but scanning the YOLO-related issues raises some concerns currently.

Can you explain the problem and the solution roadmap if this has not been resolved already? Thanks!

creafz (approved):

ternaus (approved):

glenn-jocher:
@Dipet @ternaus is this PR present in 1.0.2? I'm getting this error today when training with the new YOLOv5 integration, and I'm 100% certain that none of the YOLOv5 labels are out of bounds, as we strictly check for this on dataloader init.

```
ValueError: Expected y_max for bbox (0.6265270002186298, 0.9481956716626883, 0.700299333781004, 1.000000024214387, 0.0) to be in the range [0.0, 1.0], got 1.000000024214387.
```

Screenshot 2021-07-09 at 17 25 12

glenn-jocher:
@Dipet @ternaus I modified all probabilities to 0.0, but somehow I am still seeing the same error from albumentations. Now I'm doubly confused about why the boxes are being examined when 1) all probabilities are 0.0, and 2 ) there are no spatial transforms present that would affect the bboxes.
```python
            self.transform = A.Compose([
                A.CLAHE(p=0.0),
                A.RandomBrightnessContrast(p=0.0),
                A.RandomGamma(p=0.0),
                A.Blur(p=0.0),
                A.MedianBlur(p=0.0),
                A.ToGray(p=0.0),
                A.ImageCompression(quality_lower=75, p=0.0),
            ],
```

glenn-jocher:
@Dipet @ternaus to double check that YOLO is not creating out of bounds labels somehow I inserted this code right before the albumentations call:

```python
        if self.augment:
            if np.any(labels[:, 1:5] > 1.0):  # bboxes > 1.0
                print(labels[:, 1:5])
            
            # Albumentations
            img, labels = self.albumentations(img, labels)
```

And during my training nothing is printed before the albumentations error. So I suspect that albumentations is performing some operations on the labels that are causing them to lose numerical precision, even though no augmentations are asked for (probabilities are all zeros), and even though no spatial transforms are called for (labels should never change).

Since I see this with v1.0.2, which appears to include this PR, I would say that this PR has not resolved the issue.

Screenshot 2021-07-09 at 17 41 48

Dipet:
This looks like float numbers rounding error, because `np.isclose(1, 1.000000024214387)` returns `True`.

This check is incorrect because it is doesn't check real bbox coordinates. Albumentations uses internal bbox format - some kind of normalized pascal_voc.
To correctly check bbox you need do this steps:
```python
x, y, h, w = bbox
x1 = x - w / 2
x2 = x1 + w
y1 = y - h / 2
y2  = y1 + h

if x1 < 0 or x2 > 1 or y1 < 0 or y2 > 1:
    print(f"Bbbox greater than image: {[x1, y1, x2, y2]}")
```

Dipet:
> I modified all probabilities to 0.0, but somehow I am still seeing the same error from albumentations. Now I'm doubly confused about why the boxes are being examined when 1) all probabilities are 0.0, and 2 ) there are no spatial transforms present that would affect the boxes.

We have simplified our code and so bboxes and keypoints preprocessors will now be called regardless of whether dual transforms exist in the pipeline or not.

glenn-jocher:
@Dipet it seems like float32 loses precision on xyxy to xywh and back conversions resulting in the error.

I didn't find a good solution so I simply implemented a clip with epsilon in 'Numerical stability fix for Albumentations' PR ultralytics/yolov5#3958. This adds about 50 us of overhead per image unfortunately in the dataloader.

An alternative might be to implement an epsilon value in your Albumentation checks, i.e. for normalized data this might be something like:
```python
eps = 1E-6
assert -eps < x < (1 + eps)
```

rather than the current hard bounds:
```python
assert 0.0 < x < 1.0
```

creafz:
Hey @glenn-jocher 

We have released the [`1.0.3` version](https://github.com/albumentations-team/albumentations/releases/tag/1.0.3) that should deal with the numerical stability issue on our side.

glenn-jocher:
@creafz thanks! I've updated our requirements to 1.0.3 in https://github.com/ultralytics/yolov5/pull/4068.

zhangzhihaozi:
> Hey @glenn-jocher
> 
> We have released the [`1.0.3` version](https://github.com/albumentations-team/albumentations/releases/tag/1.0.3) that should deal with the numerical stability issue on our side.
version1.0.3  still raise the same error:   (-__-) Expected x_max for bbox (0.7486702127659575, 0.18541666666666667, 1.0053191489361701, 0.65, 0.0) to be in the range [0.0, 1.0], got 1.0053191489361701

aengusng8:
Someone and me still faced this problem #955 when using `.BboxParams(format="coco"` with `RandomCrop, CenterCrop`.

mikel-brostrom:
Still facing this issue in 1.3.0

mikel-brostrom:
My issue was that I:

`img = img.transpose((2,0,1))  # H X W C -> C X H X W `

before the transformations. Albumentations expect the image in `H X W X C` which messed up the bbox transformation for me

hiskuDN:
Check if your bbox is inside the image. Here's a short piece of code I wrote to check that. 
`# check if bbox is inside the image
def check_bbox_in_image(fileName, bbox_tl, bbox_br):
    # open the image
    image = Image.open(os.path.join('./images', fileName))

    # bounding box to be checked
    bbox = box(bbox_tl[0], bbox_tl[1], bbox_br[0],bbox_br[1])

    # image dimensions
    width, height = image.size

    # check if the bounding box is inside the image
    if bbox[0] >= 0 and bbox[1] >= 0 and bbox[2] <= width and bbox[3] <= height:
        # the bounding box is inside the image
        print("Bounding box is inside the image.")
    else:
        # the bounding box is outside the image
        print("Bounding box is outside the image.")`

sainisatish:
> I am facing below error -> Version - 1.3.1 Bounding Box = [[0.56195, 0.738133, 0.125, 0.181567], [0.046181, 0.64675, 0.092363, 0.131267], [0.271012, 0.6556, 0.375, 0.457233]] Format - Yolo Error - Expected x_min for bbox (-5.000000000005e-07, 0.5811165, 0.0923625, 0.7123835000000001, 5) to be in the range [0.0, 1.0], got -5.000000000005e-07.

i have used below function to resolve this issue-
```
def modified_check_bbox(bbox):
    for name, value in zip(["x_min", "y_min", "x_max", "y_max"], bbox[:4]):
        if not (0 - 1e-6 <= value <= 1 + 1e-6):
            raise ValueError(f"Expected {name} for bbox {bbox} to be in the range [0.0, 1.0], got {value}.")
```

PawelKinczyk:
@sainisatish me too....

sainisatish:
> @sainisatish me too....
> 
> ```
> # Define transform pipeline
> transform = A.Compose([
>     A.RandomResizedCrop(width=1280, height=1280, scale=(0.2, 2)),
>     A.HorizontalFlip(p=0.5)
> ], bbox_params=A.BboxParams(format='yolo', min_visibility=0.8))
> ```
@PawelKinczyk As of now I have dropped the images that display the mention error.

PawelKinczyk:
@sainisatish yeah I did the same but in 500 data I dropped 60 images...

martinurbieta:
This [comment ](https://stackoverflow.com/a/68588272/21373632) was helpful for me.
I had to modified my bbox_params from` format='coco'` to `format='pascal_voc'`
```
        bbox_params=dict(
            type='BboxParams',
            format='pascal_voc',
            label_fields=['gt_bboxes_labels', 'gt_ignore_flags'],
            min_visibility=0.0,
            filter_lost_elements=True),
```

PR #930, opened 2021-06-23
Add support for max size sampling in SmallestMaxSize, LongestMaxSize

kmistry-wx:
Allow a list for `max_size` and randomly sample the `max_size` to use for `SmallestMaxSize` and `LongestMaxSize`

### DIFF
--- a/albumentations/augmentations/geometric/resize.py
+++ b/albumentations/augmentations/geometric/resize.py
@@ -51,7 +51,7 @@ class LongestMaxSize(DualTransform):
     """Rescale an image so that maximum side is equal to max_size, keeping the aspect ratio of the initial image.
 
     Args:
-        max_size (int): maximum size of the image after the transformation.
+        max_size (int, list of int): maximum size of the image after the transformation.
         interpolation (OpenCV flag): interpolation method. Default: cv2.INTER_LINEAR.
         p (float): probability of applying the transform. Default: 1.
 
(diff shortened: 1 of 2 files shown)


tests/test_transforms.py
```diff
@@ -478,9 +478,9 @@ def test_longest_max_size_keypoints():
     result = aug(image=img, keypoints=keypoints)
     assert result["keypoints"] == [(18, 10, 0, 0)]
 
-    aug = A.LongestMaxSize(max_size=5, p=1)
```
Dipet, line 481: Please add independent tests for the case when `max_size` is list

kmistry-wx: Updated tests

albumentations/augmentations/geometric/resize.py
```diff
@@ -67,20 +67,23 @@ def __init__(self, max_size=1024, interpolation=cv2.INTER_LINEAR, always_apply=F
         self.interpolation = interpolation
         self.max_size = max_size
 
-    def apply(self, img, interpolation=cv2.INTER_LINEAR, **params):
-        return F.longest_max_size(img, max_size=self.max_size, interpolation=interpolation)
+    def apply(self, img, max_size=1024, interpolation=cv2.INTER_LINEAR, **params):
```
ternaus, line 70: Could you please add typing to all function parameters?

I know that it was not there before, but we are slowly moving toward having them everywhere.

kmistry-wx: Added typing

albumentations/augmentations/geometric/resize.py
```diff
@@ -51,7 +54,7 @@ class LongestMaxSize(DualTransform):
     """Rescale an image so that maximum side is equal to max_size, keeping the aspect ratio of the initial image.
 
     Args:
-        max_size (int): maximum size of the image after the transformation.
+        max_size (int, list of int): maximum size of the image after the transformation.
```
Dipet, line 57: Could you please add descriptions about how transform would workd with `list` parameter. Add information that the values will be selected from a list, and not randomly sampled from interval.

kmistry-wx: Updated docstring

Dipet (approved):

albumentations/augmentations/geometric/resize.py
kmistry-wx: Hi @ternaus, can you take another look? Thanks

HasnainRaz:
Would be cool to have this merged soon, is it possible to get it reviewed @ternaus @BloodAxe @creafz sometime soon possibly? Thanks

creafz (approved):

BloodAxe (approved):

creafz:
@HasnainRaz Could you please change the return type for `get_transform_init_args_names` (`resize.py` lines 99 and 148) from `Sequence[str]` to `Tuple[str, ...]`, so it will be compatible with the return type for the base class `BasicTransform` and mypy won't raise errors such as this - https://github.com/albumentations-team/albumentations/pull/930/checks?check_run_id=3389742084

Issue #927, opened 2021-06-18
[Feature Request] Add `position` argument to PadIfNeeded


ternaus:
Right now we put image to the center and pad it to the desired size.

It would be great to add an option to put it to the `top left`, `top right`, `bottom left`, `bottom right` positions.

ternaus:
Done.

PR #933, opened 2021-06-26
Add position parameter to the PadIfNeeded transform

yisaienkov:
Problem described in this issue #927 

I add an option to put the image to the top left, top right, bottom left, bottom right, or center positions.

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -4,7 +4,7 @@ import math
 import random
 import numbers
 import warnings
-from enum import IntEnum
+from enum import IntEnum, Enum
 from types import LambdaType
 from typing import Optional, Union, Sequence, Tuple
 
@@ -87,8 +87,9 @@ class PadIfNeeded(DualTransform):
         min_width (int): minimal result image width.
         pad_height_divisor (int): if not None, ensures image height is dividable by value of this argument.
         pad_width_divisor (int): if not None, ensures image width is dividable by value of this argument.
-        position (str): {'center', 'top_left', 'top_right', 'bottom_left', 'bottom_right'}.
-            Position of the image. Default: 'center'.
+        position (Union[str, PositionType]): Position of the image. should be PositionType.CENTER or
+            PositionType.TOP_LEFT or PositionType.TOP_RIGHT or PositionType.BOTTOM_LEFT or PositionType.BOTTOM_RIGHT.
+            Default: PositionType.CENTER.
         border_mode (OpenCV flag): OpenCV border mode.
         value (int, float, list of int, list of float): padding value if border_mode is cv2.BORDER_CONSTANT.
         mask_value (int, float,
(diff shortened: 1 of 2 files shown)


yisaienkov:
@creafz I think your ci pipeline updated numpy version and all your code is not mypy correct now
The previous pipeline has this output in "Install PyTorch on Linux and Windows" step:
`Successfully installed numpy-1.20.3 pillow-8.2.0 torch-1.8.1+cpu torchvision-0.9.1+cpu typing-extensions-3.10.0.0`
But now:
`Successfully installed numpy-1.21.0 pillow-8.2.0 torch-1.8.1+cpu torchvision-0.9.1+cpu typing-extensions-3.10.0.0`

albumentations/augmentations/transforms.py
```diff
@@ -107,6 +109,7 @@ def __init__(
         min_width: Optional[int] = 1024,
         pad_height_divisor: Optional[int] = None,
         pad_width_divisor: Optional[int] = None,
+        position: str = "center",
```
Dipet, line 112: Maybe would be better to use `Enum` for position type? You could describe it as a class parameter.

albumentations/augmentations/transforms.py
```diff
@@ -214,6 +226,35 @@ def get_transform_init_args_names(self):
...
+            h_top += h_bottom
+            w_right += w_left
+            h_bottom = 0
+            w_left = 0
+
+        if self.position == "bottom_right":
+            h_top += h_bottom
+            w_left += w_right
+            h_bottom = 0
+            w_right = 0
```
Dipet, line 254: I think it better to use `elif` constraction there.

tests/test_augmentations.py
```diff
@@ -714,6 +714,46 @@ def test_pad_if_needed(augmentation_cls: Type[PadIfNeeded], params: Dict, image_
...
+        true_result[: image_shape[0], -image_shape[1] :] = 0
+        assert (image_padded == true_result).all()
+
+    if params["position"] == "bottom_left":
+        true_result[-image_shape[0] :, : image_shape[1]] = 0
+        assert (image_padded == true_result).all()
+
+    if params["position"] == "bottom_right":
+        true_result[-image_shape[0] :, -image_shape[1] :] = 0
+        assert (image_padded == true_result).all()
```
Dipet, line 754: I think it better to use `elif` construction there.

albumentations/augmentations/transforms.py
yisaienkov: @Dipet added, can you check, please?

albumentations/augmentations/transforms.py
yisaienkov: @Dipet fixed, can you check, please?

tests/test_augmentations.py
yisaienkov: @Dipet I fixed it, can you check, please?

creafz (approved):

ternaus (approved):

Issue #850, opened 2021-03-04
Resize mask should not use nearest method of opencv


shuyangyue:
I'm using this project to train my segmentation model. I find that the mask has a right-down offset to the image. Because the opencv resize_nearest is wrong. Please refer the opencv project issue:  
[https://github.com/opencv/opencv/issues/9096](url)
[https://github.com/opencv/opencv/issues/10146](url)

The code of opencv is:
`    for( x = 0; x < dsize.width; x++ )
    {
        int sx = cvFloor(x*ifx);
        x_ofs[x] = std::min(sx, ssize.width-1)*pix_size;
    }`
It's not the nearest, so it will be a offset between image and mask.

Mask uses the resize nearest in Pillow has no offset to the image. 
Using the resize nearest of opencv to train segmentation model will have a bad result.

BloodAxe:
That's a good point which is worth investigation. From what I know, our library were key component in winning many image segmentation competitions and getting SOTA results on some segmentation benchmarks. Therefore even if there is an issue with NN interpolation of masks, the impact on trained model accuracy is not so obvious. Since one pixel offset is rarely the biggest problem. 

I can think of having `nearest-pillow` interpolation mode, but the effort in implementing this feature is not clear at this point. There are lot of usages of OpenCV interpolation under the hood in many augmentations. Would it be possible to you to run ablation study and demonstrate how big the change in accuracy of the segmentation mask for OpenCV and Pillow interpolation modes are?

shuyangyue:
Thank you for your reply. There is no effect on the iou accuracy of segmentation model. Because mask using opencv resize_nearest causes the ground truth has an offset. If I use the model to predict a image, the result will have about one pixel offset in my model. Examples are below. Left is using the origin albumentations library. Right is pillow resize nearest mode. 
![tmp](https://user-images.githubusercontent.com/13464354/110070815-bee28180-7db5-11eb-98aa-3bfbeee7626e.jpg) ![tmp2](https://user-images.githubusercontent.com/13464354/110070819-c1dd7200-7db5-11eb-94ac-2aedcdfeb0c8.jpg)
It's more obvious when resize mask to a very small size. If input is large, it will be not obvious.  Resize-nearest in pillow or opencv will have no effect on the iou accuracy, because it affects the ground truth.

BloodAxe:
[...]

PR #945, opened 2021-07-04
Allow to set a custom mask interpolation method

creafz:
This is POC to fix an issue in #850. It supports two styles for setting mask interpolation.

Option 1. Set a global value for mask interpolation through `Compose`:

```
t = A.Compose([
    A.Resize(128, 128),
], mask_interpolation=cv2.INTER_NEAREST)
```

Option 2. Override a value for mask interpolation for a single transform:

```
t = A.Compose([
    A.Resize(128, 128).set_mask_interpolation(cv2.INTER_NEAREST_EXACT),
])
```

### DIFF
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -134,9 +134,8 @@ class BaseCompose:
                 f"cv2.INTER_LINEAR_EXACT, cv2.INTER_AREA, cv2.INTER_CUBIC, cv2.INTER_LANCZOS4, cv2.INTER_MAX"
             )
         for t in self.transforms:
-            if not isinstance(t, (DualTransform, BaseCompose)):
-                continue
-            t.set_mask_interpolation(mask_interpolation)
+            if isinstance(t, BaseCompose) or (isinstance(t, DualTransform) and t.mask_interpolation is None):
+                t.set_mask_interpolation(mask_interpolation)
 
 
 class Compose(BaseCompose):
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 
 import random
 
+import cv2
 import numpy as np
 
 from albumentations.augmentations.keypoints_utils import KeypointsProcessor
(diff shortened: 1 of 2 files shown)


creafz:
Or as an alternative for Option 2, maybe it is better to an explicit argument for all `DualTransform` subclasses, such as 

```
t = A.Compose([
    A.Resize(128, 128, mask_interpolation=cv2.INTER_NEAREST_EXACT)
])
```

What do you think?

albumentations/core/composition.py
```diff
@@ -116,6 +117,26 @@ def set_deterministic(self, flag, save_key="replay"):
...
+            cv2.INTER_AREA,
+            cv2.INTER_CUBIC,
+            cv2.INTER_LANCZOS4,
+            cv2.INTER_MAX,
+        }:
+            raise ValueError(
+                f"Value {mask_interpolation} is not supported. "
+                f"Choose one of the following methods: cv2.INTER_NEAREST, cv2.INTER_NEAREST_EXACT, cv2.INTER_LINEAR, "
+                f"cv2.INTER_LINEAR_EXACT, cv2.INTER_AREA, cv2.INTER_CUBIC, cv2.INTER_LANCZOS4, cv2.INTER_MAX"
+            )
```
Dipet, line 135: Maybe better to create an Enum as class field for this flag?
With Enum error message would be much more clearly.

albumentations/core/composition.py
```diff
@@ -116,6 +117,26 @@ def set_deterministic(self, flag, save_key="replay"):
...
+        for t in self.transforms:
+            if isinstance(t, BaseCompose) or (isinstance(t, DualTransform) and t.mask_interpolation is None):
+                t.set_mask_interpolation(mask_interpolation)
```
Dipet, line 138: Strange condition, I am not sure that it will be work correctly, because `mask_interpolation` has no default value.

Dipet:
> Or as an alternative for Option 2, maybe it is better to an explicit argument for all `DualTransform` subclasses, such as
> 
> ```
> t = A.Compose([
>     A.Resize(128, 128, mask_interpolation=cv2.INTER_NEAREST_EXACT)
> ])
> ```
> 
> What do you think?

I think approach better, but it is need to change `__init__` method for all transforms. If we will do this, we must to add `**kwargs` as an argument for all transforms to have possibility to add new features simpler

LucaBonfiglioli:
please

ternaus:
@LucaBonfiglioli Will look into this.

zakajd:
Heavy +1 for this PR. It's been 3 years since discussion and PR looks almost done. What's needed to finalise it? @ternaus @Dipet 

The problems with resizing are very difficult to debug and lead to worse model performance. I had bad experience using Albumentations for segmentations and super resolutions tasks due to that and avoid using most of "geometrical" transformations for the same reason.

ternaus:
@zakajd 

We can do it.

As I understand, you are "Power User", meaning use the library for your task + use advanced features.

Sent you invitation to connect on LInkedIn, let's have video chat, your feedback / requests will help with the prioritization.

ternaus:
Already imaplemented.

PR #946, opened 2021-07-05
Remove redundant search of first & last dual transform

BloodAxe:
While working on custom bbox augmentation logic I found that logic of searching of the first & last dual augmentations derived from DualTransform is unclear and redundant to my eye. At least I don't see any other reason for having this other than mere speed improvement (By avoiding keypoints conversion back & forth for each transform). 
Removing this logic makes compose logic a bit more clear and allows to modify bboxes in non-DualTransform based augmentations.

### DIFF
--- a/albumentations/__init__.py
+++ b/albumentations/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-__version__ = "1.0.1"
+__version__ = "1.0.2"
 
 from .core.composition import *
 from .core.transforms_interface import *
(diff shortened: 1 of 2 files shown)


albumentations/__init__.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-__version__ = "1.0.0"
```
Dipet, line 3: We have #948 with version bump

Dipet (approved):

creafz (approved):

PR #959, opened 2021-07-10
[WIP] add PixelDistributionAdaptation transform

arsenyinfo:


Dipet:
Please add also transform to serialization tests: https://github.com/albumentations-team/albumentations/blob/master/tests/test_serialization.py

albumentations/augmentations/domain_adaptation.py
```diff
@@ -200,3 +220,85 @@ def get_transform_init_args_names(self):
...
+    Another naive and quick pixel-level domain adaptation. It fits a simple transform (such as PCA, StandardScaler
+    or MinMaxScaler) on both original and reference image, transforms original image with transform trained on this
+    image and then performs inverse transformation using transform fitted on reference image.
+
+    Args:
+        reference_images (List[str] or List(np.ndarray)): List of file paths for reference images
+            or list of reference images.
+        blend_ratio (float, float): Tuple of min and max blend ratio. Matched image will be blended with original
+            with random blend factor for increased diversity of generated images.
+        read_fn (Callable): Used-defined function to read image. Function should get image path and return numpy
```
Dipet, line 236: I think would be great to describe a function example in docstring, or at least the function arguments.

albumentations/augmentations/domain_adaptation.py
```diff
@@ -200,3 +220,85 @@ def get_transform_init_args_names(self):
...
+        read_fn: Callable[[Union[str, np.ndarray]], np.ndarray] = read_rgb_image,
+        always_apply=False,
+        p=0.5,
+        transform_type: str = "pca",
+    ):
+        super().__init__(always_apply=always_apply, p=p)
+        self.reference_images = reference_images
+        self.read_fn = read_fn
+        self.blend_ratio = blend_ratio
+        assert transform_type in ("pca", "standard", "minmax")
```
Dipet, line 263: Better to user exceptions instead of asserts, because we can disable asserts.

albumentations/augmentations/domain_adaptation.py
```diff
@@ -200,3 +220,85 @@ def get_transform_init_args_names(self):
...
+        self.transform_type = transform_type
+
+    @staticmethod
+    def _validate_shape(img: np.ndarray):
+        if len(img.shape) != 3:
```
Dipet, line 268: To check that image is gray we have function [is_grayscale_image](https://github.com/albumentations-team/albumentations/blob/95a048d342cbe748e5acf15cb1a28611a6303885/albumentations/augmentations/functional.py#L82)

albumentations/augmentations/domain_adaptation.py
```diff
@@ -200,3 +220,85 @@ def get_transform_init_args_names(self):
...
+            raise ValueError(
+                f"Unexpected image shape: expected 3 dimensions, got {len(img.shape)}."
+                f"Is it a grayscale image? It's not supported for now."
+            )
+
+    def ensure_uint8(self, img: np.ndarray) -> Tuple[np.ndarray, bool]:
+        if img.dtype == np.float32:
+            return (img * 255).astype("uint8"), True
```
Dipet, line 276: This conversion is dangerous because the image could be normalized in different way. At least the warning is needed there.

albumentations/augmentations/domain_adaptation.py
arsenyinfo: Extended the comment a bit. However, please keep in mind that I just took this part from the existing HistogramMatching augmentation.

albumentations/augmentations/domain_adaptation.py
arsenyinfo: added more explicit check instead!

creafz:
Hey @arsenyinfo 

Thanks for your contribution. 

Currently, `qudida` has `opencv-python` as a dependency, but albumentations' wheel requires `opencv-python-headless`. So if we merge the current PR albumentations will start to download and install both `opencv-python-headless` (it own dependency) and `opencv-python` (qudida's dependency) into a single Python environment.

The simplest way to fix it, in my opinion, is to slightly modify the logic in qudida's `setup.py` and configure a CI pipeline to upload to PyPI 2 versions of the library:
1. `qudida`, which will be the current version that depends on `opencv-python`.
2. `qudida-headless` which has `opencv-python-headless` as a dependency (and not `opencv-python`).

Then we will add `qudida-headless` as a dependency for albumentations.

If you want, I can contribute a PR with these changes to `qudida`

albumentations/augmentations/domain_adaptation.py
```diff
@@ -200,3 +220,96 @@ def get_transform_init_args_names(self):
...
+            raise ValueError(
+                f"Unexpected image shape: expected 3 dimensions, got {len(img.shape)}."
+                f"Is it a grayscale or multispectral image? It's not supported for now."
+            )
+
+    def ensure_uint8(self, img: np.ndarray) -> Tuple[np.ndarray, bool]:
+        if img.dtype == np.float32:
+            if img.min() < 0 or img.max() > 1:
+                message = (
+                    "PixelDistributionAdaptation uses uint8 under the hood, so float32 shoukd be converted,"
```
Dipet, line 283: `so float32 shoukd be converted` - typo in word `should`

albumentations/augmentations/domain_adaptation.py
```diff
@@ -200,3 +220,96 @@ def get_transform_init_args_names(self):
...
+        img, needs_reconvert = self.ensure_uint8(img)
+
+        adapted = adapt_pixel_distribution(
+            img=img,
+            ref=reference_image,
+            weight=blend_ratio,
+            transform_type=self.transform_type,
+        )
+        if needs_reconvert:
+            adapted = adapted.astype("float32") / 255
```
Dipet, line 302: multiplication `adapted.astype("float32") * (1 / 255)` is much faster

Dipet (approved):

albumentations/augmentations/domain_adaptation.py
```diff
@@ -75,6 +83,18 @@ def apply_histogram(img, reference_image, blend_ratio):
...
 
+@preserve_shape
+def adapt_pixel_distribution(
+    img: np.ndarray, ref: np.ndarray, transform_type: str = "pca", weight: float = 0.5
+) -> np.ndarray:
+    initial_type = img.dtype
+    transformer = {"pca": PCA, "standard": StandardScaler, "minmax": MinMaxScaler}[transform_type]()
+    adapter = DomainAdapter(transformer=transformer, ref_img=ref)
+    result = adapter(img).astype("float32")
+    blended = (img.astype("float32") * (1 - weight) + result * weight).astype(initial_type)
```
Dipet, line 94: `cv2.addWeighted` might be faster there
https://docs.opencv.org/4.5.2/d2/de8/group__core__array.html#gafafb2513349db3bcff51f54ee5592a19

arsenyinfo: For 100*100 images the difference is 7 µs (27 vs 34) on my laptop - I'd stick with `numpy` version for its better readability.

albumentations/augmentations/domain_adaptation.py
arsenyinfo: fixed!

creafz (approved):

PR #963, opened 2021-07-15
Fix wrong shape after ToTensorV2 in _check_data_post_transform

Dipet:


ternaus (approved):

ternaus:
Version bump?

creafz (approved):

albumentations/core/composition.py
```diff
@@ -199,7 +199,12 @@ def __call__(self, *args, force_apply=False, **data):
         return data
 
     def _check_data_post_transform(self, data):
-        rows, cols = data["image"].shape[:2]
+        img = data["image"]
+        if isinstance(img, np.ndarray):
```
BloodAxe, line 203: Maybe we introduce a helper function like this to get a shape from image/tensor/etc for a better clarity

`def infer_shape(img: Any) -> Tuple[int,int]:
    """
    TODO: Add comment
    """
    if isinstance(img, np.ndarray):
        return img.shape[:2]

    try:
        import torch
        if torch.is_tensor(img):
            return img.shape[-2:]
    except ImportError:
        pass

    raise RuntimeError("Unsupported type")        `

Dipet:
I think to merge #958 after this PR and push version into PYPI

albumentations/core/composition.py
Dipet: Done

BloodAxe (approved):

PR #972, opened 2021-07-28
Automized common tests

Dipet:
When a new transformer is added, it will be automatically added to common tests like:
- `test_image_only_augmentations`
- `test_image_only_augmentations_with_float_values`
- `test_dual_augmentations`
- `test_dual_augmentations_with_float_values`
- `test_augmentations_wont_change_input`
- `test_augmentations_wont_change_float_input`
- `test_augmentations_wont_change_shape_grayscale`
- `test_augmentations_wont_change_shape_rgb`
- `test_multichannel_image_augmentations`
- `test_float_multichannel_image_augmentations`
- `test_multichannel_image_augmentations_diff_channels`
- `test_float_multichannel_image_augmentations_diff_channels`
- `test_augmentations_serialization`
- `test_augmentations_for_bboxes_serialization`
- `test_augmentations_for_keypoints_serialization`
- `test_additional_targets_for_image_only_serialization`
- `test_augmentations_serialization_with_custom_parameters`
- `test_augmentations_serialization_to_file_with_custom_parameters`
- `test_binary_mask_interpolation`
- `test_semantic_mask_interpolation`
- `test_multiprocessing_support`
- `test_additional_targets_for_image_only`


BloodAxe (approved):

tests/utils.py
```diff
@@ -56,3 +60,125 @@ def save_value(self, value, file):
...
+                continue
+        except AttributeError:
+            pass
+
+        result.append((cls, custom_arguments.get(cls, {})))
+
+    return result
+
+
+def get_transforms(
```
creafz, line 133: It looks like those three functions differ only on filtering logic for augmentation classes. Maybe it is better to implement a single function that will take an additional argument with a filter function, e.g.:

```
def get_transforms(filter_fn, custom_arguments, except_augmentations):
    ....
    if not filter_fn(cls) or cls in except_augmentations:
        continue
```

And a filter function will look something like this:

```
def filter_dual_transforms(cls):
    return issubclass(cls, albumentations.DualTransform) and not cls == albumentations.DualTransform
```

Dipet: Done

creafz (approved):

PR #974, opened 2021-07-30
Reversible Resize

Dipet:
I tried to implement the interface for Test Time Augmentations (TTA) using `ReplayCompose` API.

If some transform could support reverse operation we must implement these functions:
- `get_reverse_args` - to save all needed parameters to do reverse transformation
- `reverse_image` - to reverse images and masks
- `reverse_bbox` - to reverse bboxes
- `reverse_keypoint` - to reverse keypoints.

Example:

```python
import numpy as np
import albumentations as A

t = A.ReversibleCompose(
    [A.Resize(512, 512, p=1)], bbox_params=A.BboxParams("pascal_voc"), keypoint_params=A.KeypointParams("xy")
)
img = np.empty([100, 100, 3], dtype=np.uint8)
bboxes = [(0, 0, 10, 10, 1)]
keypoints = [(10, 20)]

res = t(image=img, bboxes=bboxes, keypoints=keypoints, test="test")
res = t.reverse(**res)

assert np.allclose(np.array(bboxes), np.array(res["bboxes"]))
assert np.allclose(np.array(keypoints), np.array(res["keypoints"]))
```

### DIFF
--- a/albumentations/augmentations/bbox_utils.py
+++ b/albumentations/augmentations/bbox_utils.py
@@ -1,4 +1,7 @@
 from __future__ import division
+
+import typing
+
 from albumentations.core.utils import DataProcessor
 
 import numpy as np
--- a/albumentations/augmentations/geometric/resize.py
+++ b/albumentations/augmentations/geometric/resize.py
@@ -1,5 +1,8 @@
 import cv2
 import random
+import typing
+
+import numpy as np
 
 from . import functional as F
 from ...core.transforms_interface import DualTransform, to_tuple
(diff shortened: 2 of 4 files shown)


glenn-jocher:
@Dipet thanks for the PR! This would be super useful. 

For the TTA use case we do not need to reverse the images, only the boxes/keypoints. What's the use case for reversing image transforms? Aren't many image transforms non-reversible in any case?

Dipet:
At the moment, this is just a blank. 

> For the TTA use case we do not need to reverse the images, only the boxes/keypoints. What's the use case for reversing image transforms? Aren't many image transforms non-reversible in any case?

We usually use the same function for segmentation masks, this function is only needed for them.

glenn-jocher:
@Dipet oh got it, ok! The main thing i was worried about was the extra cost in reversing a full image transform, but if only the keypoints are reversed and not the full images that's perfect.

agporto:
Is this still planned? It would super useful to have this feature.

hsuominen:
+1, this would be great to have

PR #975, opened 2021-08-02
Fix Affine args to accept mask interp. method. 

dskkato:
In a doc string of Affine, there is mask_interpolation option but users can not specify that method through args. It seems to be better to remove this part if it is intended.


Dipet:
Looks like a mistake and better to add this argument to transform

dskkato:
Here it is.

Dipet (approved):

ternaus (approved):

PR #978, opened 2021-08-06
bounding box clipping in normalize_bbox

developer0hye:
The range of values of normalized bbox should be [0, 1].


BloodAxe:
Clipping bbox coordinates is like hiding the real issue under the carpet. I believe this should NOT be merged. 
Usually, user can see related error for two reasons:
1) There is an implementation bug in albumentations, which should be addressed and fixed at it's root cause.
2) There are errors in user-provided data, which should be also fixed, but on the user side.

Currently we  have mechanism for checking validity of bboxes for pascal voc, coco & yolo formats. These checks allows to spot issues quickly. Masking them is a bad design decision.

developer0hye:
@BloodAxe  Thanks for your review!

PR #980, opened 2021-08-08
Fix variable name

sergei3000:


sergei3000:
By the way, this kind of mistakes in code within docs could be avoided by using the standard library's module `doctest`. The code in docs would need some slight stylistic editing to accommodate for `doctest` though.

BloodAxe (approved):

PR #990, opened 2021-08-16
Fix Rain transform image corruption

ErlingLie:
With the current implementation of add_rain image color sometimes gets corrupted.
Original image:
scene00201

Transformed image with current implementation:
scene00201

Some of the white color turns yellow from the HLS transform. I propose to fix this by altering the brightness in HSV space instead of HLS space.
From Wikipedia: "_The difference between HSL and HSV is that a color with maximum lightness in HSL is pure white, but a color with maximum value/brightness in HSV is analogous to shining a white light on a colored object (e.g. shining a bright white light on a red object causes the object to still appear red, just brighter and more intense, while shining a dim light on a red object causes the object to appear darker and less bright)_."

In other words HSV should preserve the original colors better when altering the brightness.

The same image transformed with the edited method in this PR:
test

This PR should mitigate some of the yellow shift seen in white surfaces with the current rain augmentation.


albumentations/augmentations/functional.py
```diff
@@ -765,10 +765,10 @@ def add_rain(
         )
 
     image = cv2.blur(image, (blur_value, blur_value))  # rainy view are blurry
-    image_hls = cv2.cvtColor(image, cv2.COLOR_RGB2HLS).astype(np.float32)
-    image_hls[:, :, 1] *= brightness_coefficient
```
Dipet, line 769: It mught be better to still use HLS, but clip values to `[0, 255]` range. I think, that we can catch the same error if we will use HSV format without clipping.
```python
image_hls[:, :, 1] *= brightness_coefficient
image_hls[:, :, 1] = np.clip(image_hls[:, :, 1], 0, 255)
```

ErlingLie: I tested now and got the same corruptions as before.
test2
```  
    image_hls = cv2.cvtColor(image, cv2.COLOR_RGB2HLS).astype(np.float32)
    image_hls[:, :, 1] *= brightness_coefficient
    image_hls[:, :, 1] = np.clip(image_hls[:, :, 1], 0, 255)

    image_rgb = cv2.cvtColor(image_hls.astype(np.uint8), cv2.COLOR_HLS2RGB)```

ErlingLie: Naturally we should not be able to get neither overflow nor underflow errors, as the brightness coefficient should be between 0 and 1. The value cannot increase above 255 as we scale by a value below 1, and it cannot go below 0 as we scale by a non-negative value.

Therefore I do not believe clipping will solve this problem, and I also do not believe overflow is the issue.

Dipet: Ok, I got it. I see that in HLS colors have changed too, but in HSV they look the same.

> brightness coefficient should be between 0 and 1
Good point, I forgot that the transform doesn't support values greater than 1 for brightness coefficient.

Dipet (approved):

ternaus (approved):

Issue #942, opened 2021-07-02
[Code improvement] Add typization to all function arguments


ternaus:

rayxke:
Does this mean to add type hints/ function annotations?

Dipet:
Yes, needed to add standard python type hints. As example #988

ternaus:
Fixed

PR #994, opened 2021-08-25
Added Type Hints to Core

rayxke:
Added type hints to files in core. #942

### DIFF
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -2,6 +2,7 @@ from __future__ import division
 from collections import defaultdict
 
 import random
+from typing import Dict, List, Any
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


albumentations/core/composition.py
```diff
@@ -30,11 +31,11 @@
 
 
 class Transforms:
-    def __init__(self, transforms):
+    def __init__(self, transforms: List[Any]):
```
Dipet, line 34: `transforms: typing.Union[BasicTransform, "BaseCompose"]`

albumentations/core/composition.py
```diff
@@ -61,31 +62,31 @@ def get_always_apply(self, transforms):
                 new_transforms.append(transform)
         return Transforms(new_transforms)
 
-    def __getitem__(self, item):
+    def __getitem__(self, item: Any):
         return self.transforms[item]
 
 
-def set_always_apply(transforms):
+def set_always_apply(transforms: List[Any]):
```
Dipet, line 69: We could remove this function because it is unused.

Dipet:
In many cases, you use `Any` type, but we can determine the exact type of functions.
Please, debug all the functions and check exactly what types they use.
You could starts from debugging of these code:
```python
import numpy as np
import albumentations as A


img = np.random.randint(0, 256, [100, 100, 3], dtype=np.uint8)
bboxes = [[0, 0, 10, 10, 1]]
keypoints = [[0, 0]]
mask = img

t = A.Compose([A.Resize(256, 256), A.Normalize()], bbox_params=A.BboxParams("pascal_voc"), keypoint_params=A.KeypointParams("xy"))
res = t(image=img, mask=mask, keypoints = keypoints, bboxes=bboxes)

```

Dipet:
Done in #1219

Issue #861, opened 2021-03-30
Need to check that bboxes or keypoints without params


Dipet:
I think that library must to raise exception if bboxes or keypoints provided into transforms without describing their params.
This is important because without this checks transforms will be silently processed well, but result bboxes may be invalid.

Example
```python
import albumentations as A
import numpy as np

img = np.empty([100, 100, 3], dtype=np.uint8)
bboxes = [
    [0, 0, 99, 99, 0]
]

t = A.Compose([
    A.RandomCrop(50, 50),
],
    # bbox_params=A.BboxParams("pascal_voc")
)

res = t(image=img, bboxes=bboxes)
print(res["bboxes"])
# [(-0.1, -0.64, 197.9, 197.36, 0)]
```

VirajBagal:
@Dipet I would like to work on this!

Dipet:
Fixed in #1013

PR #1003, opened 2021-09-06
Raise Exception if `bbox_params` is not specified and bbox transformation is called

VirajBagal:
In response to this issue: https://github.com/albumentations-team/albumentations/issues/861

Made changes in `_check_args` method in `albumentations/core/composition.py`.

### DIFF
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -255,7 +255,7 @@ class Compose(BaseCompose):
                     if not isinstance(data[0], np.ndarray):
                         raise TypeError("{} must be list of numpy arrays".format(data_name))
             if internal_data_name in check_bbox_param and self.processors.get("bboxes") is None:
-                raise Exception("bbox_params must be specified for bbox transformations")
+                raise ValueError("bbox_params must be specified for bbox transformations")
 
 
 class OneOf(BaseCompose):
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -226,6 +226,7 @@ class Compose(BaseCompose):
         return data
 
     def _to_dict(self):
+        self.is_check_args = True
         dictionary = super(Compose, self)._to_dict()
         bbox_processor = self.processors.get("bboxes")
         keypoints_processor = self.processors.get("keypoints")
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -7,7 +7,10 @@ import random
 import numpy as np
 
 from albumentations.augmentations.keypoints_utils import KeypointsProcessor
-from albumentations.core.serialization import SerializableMeta, get_shortest_class_fullname
+from albumentations.core.serialization import (
+    SerializableMeta,
+    get_shortest_class_fullname,
+)
 from albumentations.core.six import add_metaclass
 from albumentations.core.transforms_interface import DualTransform, BasicTransform
 from albumentations.core.utils import format_args, Params, get_shape
@@ -87,7 +90,11 @@ class BaseCompose:
         return self.indented_repr()
 
     def indented_repr(self, indent=REPR_INDENT_STEP):
-        args = {k: v for k, v in self._to_dict().items() if not (k.startswith("__") or k == "transforms")}
+        args = {
+            k: v
+            for k, v in self._to_dict().items()
+            if not (k.startswith("__") or k == "transforms")
+        }
         repr_string = self.__class__.__name__ + "(["
         for t in self.transforms:
             repr_string += "\n"
@@ -96,7 +103,11 @@ class BaseCompose:
             else:
                 t_repr = repr(t)
             repr_string += " " * indent + t_repr + ","
-        repr_string += "\n" + " " * (indent - REPR_INDENT_STEP) + "], {args})".format(args=format_args(args))
+        repr_string += (
+            "\n"
+            + " " * (indent - REPR_INDENT_STEP)
+            + "], {args})".format(args=format_args(args))
+        )
         return repr_string
 
     @classmethod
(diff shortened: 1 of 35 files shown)


albumentations/core/composition.py
```diff
@@ -253,6 +254,8 @@ def _check_args(self, **kwargs):
                 if data:
                     if not isinstance(data[0], np.ndarray):
                         raise TypeError("{} must be list of numpy arrays".format(data_name))
+            if internal_data_name in check_bbox_param and self.processors.get("bboxes") is None:
+                raise Exception("bbox_params must be specified for bbox transformations")
```
Dipet, line 258: Better to change `Exception` to `ValueError`.

Dipet:
Please, also add this test into `tests/test_core`

```python
def test_bbox_params_is_not_set(image, bboxes):
    t = Compose([])
    with pytest.raises(ValueError) as exc_info:
        t(image=image, bboxes=bboxes)
    assert str(exc_info.value) == "bbox_params must be specified for bbox transformations"
```

albumentations/core/composition.py
VirajBagal: `Exception` changed to `ValueError`

VirajBagal:
> Please, also add this test into `tests/test_core`
> 
> ```python
> def test_bbox_params_is_not_set(image, bboxes):
>     t = Compose([])
>     with pytest.raises(ValueError) as exc_info:
>         t(image=image, bboxes=bboxes)
>     assert str(exc_info.value) == "bbox_params must be specified for bbox transformations"
> ```


Added the mentioned test.

Dipet:
Tests found problems when we combine `Compose`, `OneOf`, or another such transform.
For example look to `test_transform_pipeline_serialization_with_bboxes`.

We need to add some logic to disable `_check_args` for all transforms inherited from `BaseCompose`

VirajBagal:
`test_transform_pipeline_serialization_with_bboxes` has this example of augmentation: 
`    aug = A.Compose(
        [
            A.OneOrOther(
                A.Compose([A.RandomRotate90(), A.OneOf([A.HorizontalFlip(p=0.5), A.VerticalFlip(p=0.5)])]),
                A.Compose([A.Rotate(p=0.5), A.OneOf([A.HueSaturationValue(p=0.5), A.RGBShift(p=0.7)], p=1)]),
            ),
            A.SomeOf(
                [
                    A.HorizontalFlip(p=1),
                    A.Transpose(p=1),
                    A.HueSaturationValue(p=0.5),
                    A.RandomBrightnessContrast(p=0.5),
                ],
                n=5,
            ),
        ],
        bbox_params={"format": bbox_format, "label_fields": ["labels"]},
    )
`

This augmentation raises error because `A.Compose` within `A.OneOrOther` doesn't have the `bbox_params` defined. 

We have two options now:
1.  Either we make the user enter `bbox_params` wherever he/she uses `A.Compose`. We would then need to make these changes in every test where `A.OneOrOther` and bboxes are used. 
2.  Or we come up with a logic to use `bbox_params` defined in the outermost `A.Compose`  and use the same in the inner `A.Compose` of `A.OneOrOther`. 

What do you think? 

I tried the test with this and it works: 

`aug = A.Compose(
    [
        A.OneOrOther(
            A.Compose([A.RandomRotate90(), A.OneOf([A.HorizontalFlip(p=0.5), A.VerticalFlip(p=0.5)])], bbox_params={"format": bbox_format, "label_fields": ["labels"]}),
            A.Compose([A.Rotate(p=0.5), A.OneOf([A.HueSaturationValue(p=0.5), A.RGBShift(p=0.7)], p=1)], bbox_params={"format": bbox_format, "label_fields": ["labels"]}),
        ),
        A.SomeOf(
            [
                A.HorizontalFlip(p=1),
                A.Transpose(p=1),
                A.HueSaturationValue(p=0.5),
                A.RandomBrightnessContrast(p=0.5),
            ],
            n=5,
        ),
    ],
    bbox_params={"format": bbox_format, "label_fields": ["labels"]},
)
`

Dipet:
I think is better to skip `_check_args` in inner transforms.
To do this we must check all transforms and if the transform is inherited from `Compose` to disable this check.
Something like this:
```python
class Compose:
    def __init__(...):
        ....
        self.is_check_args = True
        for transform in self.transforms.transforms:
            if issubclass(transform, Compose):
                transform._disable_check_args()
    
    def _disable_check_args(self):
            self.is_check_args = False

    def __call__(self, *args, force_apply=False, **data):
        if self.is_check_args:
            self._check_args(**data)
```

VirajBagal:
@Dipet thanks for the guidance. Made the changes. Please take a look.

albumentations/core/composition.py
```diff
@@ -170,10 +171,25 @@ def __init__(self, transforms, bbox_params=None, keypoint_params=None, additiona
 
         self.add_targets(additional_targets)
 
+        self.is_check_args = True
```
Dipet, line 174: Please add this var into `_to_dict`.
And reformat code with `black` https://albumentations.ai/docs/contributing/

albumentations/core/composition.py
```diff
@@ -210,12 +263,15 @@ def _check_data_post_transform(self, data):
         return data
 
     def _to_dict(self):
+        self.is_check_args = True
```
Dipet, line 266: Oh my fault. We don't need this var there, because it is not an argument in `__init__`. Sorry.

VirajBagal:
@Dipet why aren't the checks getting completed?

Dipet:
Please pull the recent master branch and resolve conflicts

VirajBagal:
Recreating new PR.

PR #1004, opened 2021-09-06
Speed up elastic transform

b0nce:
image
image

A bit another way to do elastic transform.
It exploits fact that same random shift can be applied to x and y mapping.

### DIFF
--- a/albumentations/augmentations/geometric/transforms.py
+++ b/albumentations/augmentations/geometric/transforms.py
@@ -150,6 +150,7 @@ class ElasticTransform(DualTransform):
         mask_value=None,
         always_apply=False,
         approximate=False,
+        same_dxdy=False,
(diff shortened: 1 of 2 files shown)


albumentations/augmentations/geometric/transforms.py
```diff
@@ -150,6 +150,7 @@ def __init__(
         mask_value=None,
         always_apply=False,
         approximate=False,
+        same_dxdy=False,
```
Dipet, line 153: Please, add this variable into the docstring with a description. And add it into `get_transform_init_args_names`

b0nce: Done!

Dipet (approved):

creafz (approved):

Issue #676, opened 2020-08-06
Cutout and CoarseDropout should be setting height/width based on percentage not absolute pixel value


bfeeny:
In Cutout and CoarseDropout you should be allowing the height/width to be set by percentage of the image not just absolute pixels.  example height=.2

ternaus:
@bfeeny 

`CutOut` was deprecated in favor of `CoarseDropout`

As of today, `CoarseDroput` could pick holes bases on pixel values and percentage of height, width.

Would you consider becoming a sponsor for the project to make work being peformed faster?

https://github.com/sponsors/albumentations-team

PR #1005, opened 2021-09-06
CoarseDropout can now set height and width of holes based on fraction of original image height and width

VirajBagal:
In response to this issue: https://github.com/albumentations-team/albumentations/issues/676

Made changes to the `get_params_dependent_on_targets` method in the `CoarseDropout` class of `albumentations/augmentations/transforms.py`.

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -173,8 +173,11 @@ class PadIfNeeded(DualTransform):
             w_pad_left = pad_cols // 2
             w_pad_right = pad_cols - w_pad_left
 
-        h_pad_top, h_pad_bottom, w_pad_left, w_pad_right = self.__update_position_params(
-            h_top=h_pad_top, h_bottom=h_pad_bottom, w_left=w_pad_left, w_right=w_pad_right
+        (h_pad_top, h_pad_bottom, w_pad_left, w_pad_right,) = self.__update_position_params(
+            h_top=h_pad_top,
+            h_bottom=h_pad_bottom,
+            w_left=w_pad_left,
+            w_right=w_pad_right,
         )
 
         params.update(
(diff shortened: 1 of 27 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -792,8 +795,35 @@ def get_params_dependent_on_targets(self, params):
...
+            elif all(
+                [
+                    isinstance(self.min_height, float),
+                    isinstance(self.min_width, float),
+                    isinstance(self.max_height, float),
+                    isinstance(self.max_width, float),
+                ]
+            ):
+                hole_height = random.randint(round(self.min_height * height), round(self.max_height * height))
+                hole_width = random.randint(round(self.min_width * width), round(self.max_width * width))
```
Dipet, line 817: I think is better to change to these lines:
```python
hole_height = height * random.uniform(self.min_height, self.max_height)
hole_width = width * random.uniform(self.min_width, self.max_width)
```

VirajBagal: Made the changes

Dipet (approved):

creafz:
I think that there should also be a check that verifies that all float values lie in the range `[0.0, 1.0)` to prevent errors when users pass a value in pixels but accidentally use incorrect datatype for it (e.g., 40.0).

VirajBagal:
I added a method `check_range` to `CoarseDropout`. It checks the range if the given values are float. If not satisfied, it raises `ValueError`.

albumentations/augmentations/transforms.py
```diff
@@ -792,8 +820,41 @@ def get_params_dependent_on_targets(self, params):
...
+            elif all(
+                [
+                    isinstance(self.min_height, float),
+                    isinstance(self.min_width, float),
+                    isinstance(self.max_height, float),
+                    isinstance(self.max_width, float),
+                ]
+            ):
+                hole_height = int(height * random.uniform(self.min_height, self.max_height))
+                hole_width = int(width * random.uniform(self.min_width, self.max_width))
```
Dipet, line 842: I think would be great to add a fix prevent situation when height or width equal to 0 using these lines:
```python
hole_height = max(1, hole_height)
hole_width = max(1, hole_width)
```

These need because, value may be equal to `0.9` for example, but it would be clipped to `0`

VirajBagal: Even if sometimes height or width becomes 0 (considering usual image sizes, I think that itself is going to be very rare),  I think that is fine because that would return the image as it is. Part of random augmentation I would say.

VirajBagal:
@Dipet why aren't the checks getting completed?

creafz:
@VirajBagal For some strange reason GitHub doesn’t run checks for your latest commit https://github.com/albumentations-team/albumentations/pull/1005/commits/46d00af47b77ea9ad1ab199794d918f372a914e6

I will look into it. 

By the way this commit looks quite unusual because it suddenly reformatted many files.

VirajBagal:
@creafz I `black` formatted all files, so we see many file changes.  May be, these many file changes is the reason that the checks aren't running. Moreover, my forked repository is a mess right now (sorry, this is my first PR). If you don't mind, shall I close this PR, recreate my fork, make all changes mentioned here and then raise a new PR?

creafz:
@VirajBagal 
>  If you don't mind, shall I close this PR, recreate my fork, make all changes mentioned here and then raise a new PR?

Yes, lets try it.

VirajBagal:
Recreating new PR. Closing this.

PR #1038, opened 2021-10-15
Added sberdevices

ternaus:


BloodAxe:
So how exactly we rank companies in this list? It was previously in historical order, which to me the most clear and obvious one. This PR violates this and I wonder whether it's because of the size of the company or for another reason?

PR #1059, opened 2021-11-17
Fix crop_and_pad_bbox

cannon:
A.CropAndPad doesn't work correctly with bounding boxes when you pass `keep_size=False`. This seems to fix it.

We are using yolo-style boxes if that matters. The issue arises when you pad each of the 4 sides with different positive integers.


Dipet:
Could you please give some examples to reproduce the problem? These changes break existing tests `tests/test_imgaug.py::test_compare_crop_and_pad`

cannon:
> Could you please give some examples to reproduce the problem? These changes break existing tests `tests/test_imgaug.py::test_compare_crop_and_pad`

This shows the problem:

```
import albumentations as A
import numpy as np

for keep_size in [True, False]:
    transform = A.Compose(
        [A.CropAndPad(px=(8, 8, 8, 8), keep_size=keep_size)],
        bbox_params=A.BboxParams(format="yolo"),
    )

    transformed = transform(
        image=np.zeros((16, 16, 3), dtype=np.uint8),
        bboxes=[(0.5, 0.5, 0.5, 0.5, "label")],
    )

    print(keep_size, transformed["bboxes"])
```

output:
```
True [(0.5, 0.5, 0.25, 0.25, 'label')]
False [(0.875, 0.875, 0.25, 0.25, 'label')]
```

These bounding box coordinates are measured on [0,1], so scaling the image with keep_size shouldn't affect them.

Dipet:
Oh, now I understand. Thanks a lot for finding this bug.
We have a bug in the deprecated API for imgaug transforms. [There](https://github.com/albumentations-team/albumentations/blob/master/albumentations/imgaug/transforms.py#L78
) we do not check if the image size is changed.

Please, remove `keep_size` from `crop_and_pad_bbox`, because it is not being used now.
And just remove test cases where `keep_size=False` from [here](https://github.com/albumentations-team/albumentations/blob/master/tests/test_imgaug.py#L259), because they are invalid.

Dipet (approved):

PR #1063, opened 2021-11-19
Implementation of UnsharpMask transform 

### LINKED ISSUE #1062: Feature request: Unsharp Masking (USM) transform.
## Feature description
Hi, it's me again :laughing:

I noticed that current `Sharpen` transform implementation uses simple sharpening kernel, which doesn't work for sharpening noisy images, because it amplifies noise. 

https://github.com/albumentations-team/albumentations/blob/477156d5f14a26f8f0f71c5f2c0d59748bb7de4b/albumentations/augmentations/transforms.py#L3005-L3008

There is an alternative operation called Unsharp Masking (despite it's name, it has nothing to do with masks). You can read more about it here: https://www.idtools.com.au/unsharp-masking-with-python-and-opencv/

## Implementation details

Code for this layer is straightforward:
1. We can use `skimage.filters.unsharp_mask`
2. Implement it using `cv2`
3. Implement on pure `numpy` operations.

zakajd:
Closes #1062 

This is WIP, some questions I'm not sure about:
- [x] Name is OK?
- [x] Do I use correct decorators for `unsharp_mask` function?
- [x] I added typehints to function, but not class implementation. Is it OK?
- [x] Do we need to pass threshold as transform parameter or it's OK to have it fixed in function? 
- [x] I currently have only one very basic test copy-pasted from `GaussianBlur`. Is it enough?

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -1793,3 +1793,22 @@ def superpixels(
 @clipped
 def add_weighted(img1, alpha, img2, beta):
     return img1.astype(float) * alpha + img2.astype(float) * beta
+
+
+@clipped
+@preserve_shape
+def unsharp_mask(image: np.ndarray, ksize: int, sigma: float = 0.0, alpha: float = 0.2, threshold: int = 10):
+    blur_fn = _maybe_process_in_chunks(cv2.GaussianBlur, ksize=(ksize, ksize), sigmaX=sigma)
+    blur = blur_fn(image)
+    residual = image - blur
+
+    # Do not sharpen noise
+    mask = np.abs(residual) * 255 > threshold
+    mask = mask.astype("float32")
+
+    sharp = image + alpha * residual
+    # Avoid color noise artefacts.
+    sharp = np.clip(sharp, 0, 1)
+
+    soft_mask = blur_fn(mask)
+    return soft_mask * sharp + (1 - soft_mask) * image
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -1799,6 +1799,13 @@ def add_weighted(img1, alpha, img2, beta):
 @preserve_shape
 def unsharp_mask(image: np.ndarray, ksize: int, sigma: float = 0.0, alpha: float = 0.2, threshold: int = 10):
     blur_fn = _maybe_process_in_chunks(cv2.GaussianBlur, ksize=(ksize, ksize), sigmaX=sigma)
+
+    input_dtype = image.dtype
+    if input_dtype == np.uint8:
+        image = to_float(image)
+    elif input_dtype not in (np.uint8, np.float32):
+        raise ValueError("Unexpected dtype {} for UnsharpMask augmentation".format(input_dtype))
+
     blur = blur_fn(image)
     residual = image - blur
 
@@ -1811,4 +1818,5 @@ def unsharp_mask(image: np.ndarray, ksize: int, sigma: float = 0.0, alpha: float
     sharp = np.clip(sharp, 0, 1)
 
     soft_mask = blur_fn(mask)
-    return soft_mask * sharp + (1 - soft_mask) * image
+    output = soft_mask * sharp + (1 - soft_mask) * image
+    return from_float(output, dtype=input_dtype)
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -1964,10 +1964,11 @@ class GaussianBlur(ImageOnlyTransform):
         return F.gaussian_blur(image, ksize, sigma=sigma)
 
     def get_params(self):
-        return {
-            "ksize": int(random.choice(np.arange(self.blur_limit[0], self.blur_limit[1] + 1, 2))),
-            "sigma": random.uniform(*self.sigma_limit),
-        }
+        ksize = np.random.randint(self.blur_limit[0], self.blur_limit[1] + 1)
+        if ksize != 0 and ksize % 2 != 1:
+            ksize = (ksize + 1) % (self.blur_limit[1] + 1)
+
+        return {"ksize": ksize, "sigma": random.uniform(*self.sigma_limit)}
 
     def get_transform_init_args_names(self):
         return ("blur_limit", "sigma_limit")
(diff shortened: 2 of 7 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -1963,11 +1964,10 @@ def apply(self, image, ksize=3, sigma=0, **params):
...
 
     def get_params(self):
-        ksize = np.random.randint(self.blur_limit[0], self.blur_limit[1] + 1)
-        if ksize != 0 and ksize % 2 != 1:
-            ksize = (ksize + 1) % (self.blur_limit[1] + 1)
-
-        return {"ksize": ksize, "sigma": random.uniform(*self.sigma_limit)}
+        return {
+            "ksize": int(random.choice(np.arange(self.blur_limit[0], self.blur_limit[1] + 1, 2))),
+            "sigma": random.uniform(*self.sigma_limit),
```
zakajd, line 1969: Here I just unified kernel size selection with other functions.

Dipet:
Could you also provide some examples with different arguments for transformation?

albumentations/augmentations/functional.py
```diff
@@ -1793,3 +1793,22 @@ def superpixels(
...
 def add_weighted(img1, alpha, img2, beta):
     return img1.astype(float) * alpha + img2.astype(float) * beta
+
+
+@clipped
+@preserve_shape
+def unsharp_mask(image: np.ndarray, ksize: int, sigma: float = 0.0, alpha: float = 0.2, threshold: int = 10):
+    blur_fn = _maybe_process_in_chunks(cv2.GaussianBlur, ksize=(ksize, ksize), sigmaX=sigma)
+    blur = blur_fn(image)
+    residual = image - blur
```
Dipet, line 1803: We must do this operation in int32 or float data type, I am not sure that it will be automatically converted to the correct data type

albumentations/augmentations/transforms.py
Dipet: I think it's better to do it in another PR, to be sure that we haven't broken anything.

albumentations/augmentations/transforms.py
```diff
@@ -3240,3 +3240,62 @@ def _to_dict(self):
...
+
+        if (self.blur_limit[0] != 0 and self.blur_limit[0] % 2 != 1) or (
+            self.blur_limit[1] != 0 and self.blur_limit[1] % 2 != 1
+        ):
+            raise ValueError("UnsharpMask supports only odd blur limits.")
+
+    @staticmethod
+    def __check_values(value, name, bounds=(0, float("inf"))):
+        if not bounds[0] <= value[0] <= value[1] <= bounds[1]:
+            raise ValueError("{} values should be between {}".format(name, bounds))
```
Dipet, line 3287: Better to use f-strings

albumentations/augmentations/transforms.py
```diff
@@ -3240,3 +3240,62 @@ def _to_dict(self):
...
+    Args:
+        blur_limit (int, (int, int)): maximum Gaussian kernel size for blurring the input image.
+            Must be zero or odd and in range [0, inf). If set to 0 it will be computed from sigma
+            as `round(sigma * (3 if img.dtype == np.uint8 else 4) * 2 + 1) + 1`.
+            If set single value `blur_limit` will be in range (0, blur_limit).
+            Default: (3, 7).
+        sigma_limit (float, (float, float)): Gaussian kernel standard deviation. Must be greater in range [0, inf).
+            If set single value `sigma_limit` will be in range (0, sigma_limit).
+            If set to 0 sigma will be computed as `sigma = 0.3*((ksize-1)*0.5 - 1) + 0.8`. Default: 0.
+        alpha ((float, float)): range to choose the visibility of the sharpened image. At 0, only the original image is
```
Dipet, line 3258: You use to_tuple for this arg, so single float also supported

albumentations/augmentations/transforms.py
```diff
@@ -3240,3 +3240,62 @@ def _to_dict(self):
...
+        self.blur_limit = to_tuple(blur_limit, 3)
+        self.sigma_limit = self.__check_values(to_tuple(sigma_limit, 0.0), name="sigma_limit")
+        self.alpha = self.__check_values(to_tuple(alpha, 0.0), name="alpha", bounds=(0.0, 1.0))
+
+        if self.blur_limit[0] == 0 and self.sigma_limit[0] == 0:
+            self.blur_limit = 3, max(3, self.blur_limit[1])
+            warnings.warn(
+                "blur_limit and sigma_limit minimum value can not be both equal to 0. "
+                "blur_limit minimum value changed to 3."
+            )
```
Dipet, line 3277: I think better to raise ValueError

albumentations/augmentations/transforms.py
```diff
@@ -3240,3 +3240,62 @@ def _to_dict(self):
...
+
+    @staticmethod
+    def __check_values(value, name, bounds=(0, float("inf"))):
+        if not bounds[0] <= value[0] <= value[1] <= bounds[1]:
+            raise ValueError("{} values should be between {}".format(name, bounds))
+        return value
+
+    def get_params(self):
+        return {
+            "ksize": int(random.choice(np.arange(self.blur_limit[0], self.blur_limit[1] + 1, 2))),
```
Dipet, line 3292: I think better to use `random.randrange(self.blur_limit[0], self.blur_limit[1] + 1, 2)`

albumentations/augmentations/transforms.py
```diff
@@ -3240,3 +3240,62 @@ def _to_dict(self):
...
+            "sigma": random.uniform(*self.sigma_limit),
+            "alpha": random.uniform(*self.alpha),
+        }
+
+    def apply(self, img, ksize=3, sigma=0, alpha=0.2, **params):
+        return F.unsharp_mask(img, ksize, sigma=sigma, alpha=alpha)
+
+    def get_transform_init_args_names(self):
+        return ("blur_limit", "alpha")
```
Dipet, line 3301: You forget `sigma_limit`

Dipet:
I think would be great to set a link to the original article.

- Name is OK? -  I think yes.
- Do I use correct decorators for unsharp_mask function? - Looks good to me
- I added typehints to function, but not class implementation. Is it OK? - It would be great if you also add type hints to the class.
- Do we need to pass threshold as transform parameter or it's OK to have it fixed in function? - I think it's better to add this as a transformation parameter. 
- I currently have only one very basic test copy-pasted from GaussianBlur. Is it enough? - This is fine, but if you come up with more tests it would be great.

albumentations/augmentations/transforms.py
Dipet: We rarely use them because we used to support python2 and python3.5 which didn't support them.

zakajd:
Tested that it works with both `uint8` and `float32` images.

Examples of applying this transform:
```python
transform = A.Compose([
    A.UnsharpMask(blur_limit=(7, 15), alpha=(0.7, 0.9), p=1.0)
])
```

image | image


image | image

Dipet:
Please, add this transform into serialization tests https://github.com/albumentations-team/albumentations/blob/master/tests/test_serialization.py#L66

tests/utils.py
```diff
@@ -120,7 +121,7 @@ def get_transforms(
 
 def check_all_augs_exists(
     augmentations: typing.List[typing.List],
-    except_augmentations: typing.Optional[typing.Set[typing.Type[albumentations.BasicTransform]]] = None,
+    except_augmentations: Optional[Set] = None,
```
zakajd, line 124: I changed type for this function to be more generic, because otherwise `mypy` marked it as wrong and didn't allow to commit.

Dipet (approved):

ternaus (approved):

PR #1064, opened 2021-11-19
Implementation of RingingOvershoot transform

### LINKED ISSUE #1061: Feature request: Add augmentation to mimic ringing and overshoot artefacts.
## Feature description

Ringing artifacts often appear as spurious edges near sharp transitions in an image. They visually look like bands or “ghosts” near edges. Overshoot artifacts are usually combined with ringing artifacts, which manifest themselves as an increased jump at the edge transition. The main cause of these artifacts is that the signal is bandlimited without high frequencies. These artifacts are very common and usually produced by a sharping algorithm, JPEG compression, etc. See examples from [RealESRGAN paper](https://arxiv.org/pdf/2107.10833.pdf). 

image

Idea is to use special _sinc_ filter to mimic such artefacts on clean images. _sinc_ filter kernel is defined as:
image

Then artefacts created on images will look like that:

image

## Implementation
We need to create a special function to create such kernels (probably in `functional.transforms` and then use it together with `convolve` to apply kernel to the image. 

Here is a code snippet that does this:

https://github.com/xinntao/BasicSR/blob/4d34d071218d0e767096eddefa919200d5239936/basicsr/data/degradations.py#L392-L410

## Additional context
There is an existing `Sharpen` transform in Albumentation, which implements an idea proposed here in simplified form.

zakajd:
Closes #1061 

List of TODOs
- [x] Add tests
- [x] Update README and other required places.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,8 @@ Pixel-level transforms will change just an input image and will leave any additi
 - [ToFloat](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.ToFloat)
 - [ToGray](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.ToGray)
 - [ToSepia](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.ToSepia)
+- [RingingOvershoot](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.RingingOvershoot)
+
 
 ### Spatial-level transforms
 Spatial-level transforms will simultaneously change both an input image as well as additional targets such as masks, bounding boxes, and keypoints. The following table shows which additional targets are supported by each transform.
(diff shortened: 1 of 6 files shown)


Dipet:
Could you provide some example images of how this transform works?

zakajd:
```python
transform = A.RingingOvershoot(
    limit=(7, 21), 
    cutoff=(np.pi / 4, np.pi / 2),
    p=1.0
)
```
Examples: 
image | image
Ringing can be clearly seen on this one. 
image | image

README.md
```diff
@@ -165,6 +165,8 @@ Pixel-level transforms will change just an input image and will leave any additi
 - [ToFloat](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.ToFloat)
 - [ToGray](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.ToGray)
 - [ToSepia](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.ToSepia)
+- [RingingOvershoot](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.RingingOvershoot)
```
Dipet, line 168: This list is sorted in alphabetic order, so please add this transform to the right place or generate this table with `tools/make_transforms_docs.py` and just replace the whole list.

zakajd: Done

Dipet (approved):

ternaus (approved):

PR #1066, opened 2021-11-19
Implementation of AdvancedBlur augmentation

### LINKED ISSUE #1060: Feature request: Extend Blur to anisotropic kernels and generalised normal distribution.
## Feature request.
Hi! Thanks for an awesome library.
I have an idea for new augmentation(-s) to make it even better =)

Currently Albumentations has 5 types of blur filtering: 
1. `transforms.Blur` — which applies **box** filter with randomly selected kernel size.
2. `transforms.MedianBlur` — which applies **median** filtering 
3. `transforms.GaussianBlur`  — which applies **isotropic (symmetric) Gaussian** filter. See example below:
photo_2021-11-18_23-27-23
4. `transforms.GlassBlur` — which again does multiple blurring iterations with **isotropic Gaussian**
5. `transforms.MotionBlur` — which blurs by a "line" filter. See example below:
image

Here is the thing — isotropic Gaussian is the **simplest** case of a more general normal distribution. 
1. Kernel can be asymmetric (anisotropic). See example:
photo_2021-11-18_23-27-25
2. It isn't fixed to XY coordinates and can be rotated to any degree `theta`. See example:
photo_2021-11-18_23-27-26
3. By varying distribution generalisation parameter we can make it tails heavier of lighter.  See examples: 
photo_2021-11-18_23-27-27  |  photo_2021-11-18_23-27-28

4. We can also include **noise**  on kernel level, which is shown to better model camera ISP pipelines. See examples: 

photo_2021-11-18_23-27-32 | photo_2021-11-18_23-27-33
 
photo_2021-11-18_23-27-34 | photo_2021-11-18_23-27-35

## Implementation details

Code to generate images above is take from BasicSR repo: https://github.com/xinntao/BasicSR/blob/master/basicsr/data/degradations.py
It can be enhanced, cleaned up and placed into new `AdvancedBlur` transform that does everything at once.

zakajd:
Closes #1060 

TODO:
- [x] Tests
- [ ] Name is OK?
- [x] Is using np.random.uniform safe in this scenario? [Discussion](https://tanelp.github.io/posts/a-bug-that-plagues-thousands-of-open-source-ml-projects/)
- [x] ValueError text messages is OK?

```python
transform = A.AdvancedBlur(
    blur_limit=(15, 21), sigmaX_limit=(0.2, 0.2), sigmaY_limit=(1.0, 1.0), rotate_limit=(45, 46), p=1.0,
)
```

Examples:
image | image | image

image | image | image

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -1931,7 +1931,7 @@ class GaussianBlur(ImageOnlyTransform):
             as `round(sigma * (3 if img.dtype == np.uint8 else 4) * 2 + 1) + 1`.
             If set single value `blur_limit` will be in range (0, blur_limit).
             Default: (3, 7).
-        sigma_limit (float, (float, float)): Gaussian kernel standard deviation. Must in range [0, inf).
+        sigma_limit (float, (float, float)): Gaussian kernel standard deviation. Must be in range [0, inf).
             If set single value `sigma_limit` will be in range (0, sigma_limit).
             If set to 0 sigma will be computed as `sigma = 0.3*((ksize-1)*0.5 - 1) + 0.8`. Default: 0.
         p (float): probability of applying the transform. Default: 0.5.
@@ -3259,8 +3259,11 @@ class AdvancedBlur(ImageOnlyTransform):
         sigmaY_limit: Same as `sigmaY_limit` for another dimension.
         rotate_limit: Range from which a random angle used to rotate Gaussian kernel is picked.
             If limit is a single int an angle is picked from (-rotate_limit, rotate_limit). Default: (-90, 90).
-        beta_limit: Distribution shape parameter, 1 is the normal distribution. Default: (0.5, 8.0).
-        noise_limit: Multiplicative factor that control strength of kernel noise. Default: (0.75, 1.25).
+        beta_limit: Distribution shape parameter, 1 is the normal distribution. Values below 1.0 make distribution
+            tails heavier than normal, values above 1.0 make it lighter than normal. Default: (0.5, 8.0).
+        noise_limit: Multiplicative factor that control strength of kernel noise. Must be positive and preferably
+            centered around 1.0. If set single value `noise_limit` will be in range (0, noise_limit).
+            Default: (0.75, 1.25).
         p (float): probability of applying the transform. Default: 0.5.
 
     Reference:
@@ -3279,7 +3282,7 @@ class AdvancedBlur(ImageOnlyTransform):
         sigmaY_limit: Union[float, Sequence[float]] = (0.2, 1.0),
         rotate_limit: Union[int, Sequence[int]] = 90,
         beta_limit: Union[float, Sequence[float]] = (0.5, 8.0),
-        noise_limit: Union[float, Sequence[float]] = (0.75, 1.25),
+        noise_limit: Union[float, Sequence[float]] = (0.9, 1.1),
         always_apply=False,
         p=0.5,
     ):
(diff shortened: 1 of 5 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -1930,7 +1931,7 @@ class GaussianBlur(ImageOnlyTransform):
             as `round(sigma * (3 if img.dtype == np.uint8 else 4) * 2 + 1) + 1`.
             If set single value `blur_limit` will be in range (0, blur_limit).
             Default: (3, 7).
-        sigma_limit (float, (float, float)): Gaussian kernel standard deviation. Must be greater in range [0, inf).
+        sigma_limit (float, (float, float)): Gaussian kernel standard deviation. Must in range [0, inf).
```
Dipet, line 1934: Typo: `Must be in range`

albumentations/augmentations/transforms.py
```diff
@@ -3240,3 +3241,115 @@ def _to_dict(self):
...
+            If set single value `blur_limit` will be in range (0, blur_limit).
+            Default: (3, 7).
+        sigmaX_limit: Gaussian kernel standard deviation. Must be in range [0, inf).
+            If set single value `sigmaX_limit` will be in range (0, sigma_limit).
+            If set to 0 sigma will be computed as `sigma = 0.3*((ksize-1)*0.5 - 1) + 0.8`. Default: 0.
+        sigmaY_limit: Same as `sigmaY_limit` for another dimension.
+        rotate_limit: Range from which a random angle used to rotate Gaussian kernel is picked.
+            If limit is a single int an angle is picked from (-rotate_limit, rotate_limit). Default: (-90, 90).
+        beta_limit: Distribution shape parameter, 1 is the normal distribution. Default: (0.5, 8.0).
+        noise_limit: Multiplicative factor that control strength of kernel noise. Default: (0.75, 1.25).
```
Dipet, line 3263: Please, describe valid limits and how a single value will be interpreted.

albumentations/augmentations/transforms.py
```diff
@@ -3240,3 +3241,115 @@ def _to_dict(self):
...
+    Image types:
+        uint8, float32
+    """
+
+    def __init__(
+        self,
+        blur_limit: Union[int, Sequence[int]] = (3, 7),
+        sigmaX_limit: Union[float, Sequence[float]] = (0.2, 1.0),
+        sigmaY_limit: Union[float, Sequence[float]] = (0.2, 1.0),
+        rotate_limit: Union[int, Sequence[int]] = 90,
```
Dipet, line 3280: I think `rotate_limit` could be float

Dipet (approved):

Dipet:
- Name is OK? - Maybe better to use Filter2D. But this name ok for me too.
- Is using np.random.uniform safe in this scenario? - It is not safe, but we have a lot of transforms that use np.random and we need to fix all of them in a single PR. But you could fix this problem like [there](https://github.com/albumentations-team/albumentations/blob/master/albumentations/augmentations/transforms.py#L2064) when you need to create `RandomState` first using `random.randint`
- ValueError text messages is OK? - To me they are ok

albumentations/augmentations/transforms.py
zakajd: I used `int` type here for consistency between all other augmentations that do rotations and always take integer values (but they will work for floats too).

ternaus (approved):

BloodAxe (approved):

creafz (approved):

PR #1069, opened 2021-11-22
Fix pad_with_params to work with arbitrary number of channels

BloodAxe:


ternaus:
Could you please add argument types to the function?

BloodAxe:
Good catch. Done)

ternaus (approved):

Dipet (approved):

Issue #873, opened 2021-04-11
RNG library


ozen:
After reading [this blog post](https://tanelp.github.io/posts/a-bug-that-plagues-thousands-of-open-source-ml-projects/) about multiprocess data loading in PyTorch and rng seeds, I was looking at the albumentations source code to see if it uses numpy.random. It seems most of the transforms use Python's random lib, but some use numpy.random. Is there a reason for this choice, and do you think it would be better to use Python's random due to the linked issue?

aliab3d:
@ternaus @BloodAxe I also came across the blog post mentioned by @ozen, highlighting the issue with using PyTorch dataloaders (num_workers > 1) together with numpy RNG, which leads to identical random behaviour by all processes. It seems like a critical issue to check for Albumentations, as an augmentation library.

YassineYousfi:
I also saw the blog post on Twitter and did a quick sanity check: 
https://colab.research.google.com/drive/1ZW_7NslpNVhDt0eRSBECdJN0Gmogjuxo?usp=sharing
I believe Albumentations uses the python RNG (besides IAA transforms I think), my tests didn't show identical samples as in the blog post. But I would love to see a more thorough test!

ozen:
@YassineYousfi as I said albumentations uses both in different places (I think more than IAA transforms). You wouldn't see it if you test for specifically the parts it uses numpy. 
For example, p argument (probability for transform) is compared with Python random, so you should use p=1 when testing.

YassineYousfi:
@ozen thanks for pointing this out. In fact, `GaussianBlur` uses np.random for `ksize` but not for `sigma`.
`GaussNoise` sets the np.random.RandomState using the python PRNG, which avoids this problem. I noticed that this was done with other augmentations. 
As I said my notebook is not a thorough test, so I would expect more done to see which augs/params are affected.

Dipet:
Yes we saw this blog post to. At this time we are thinking how to better write general test to prevent usage of `np.random`
But problem that some transforms uses `np.random` to produce random arrays (for example `GaussNoise`).
We will try to fix these problems as soon as possible

ternaus:
Everything should be fine now.

PR #1070, opened 2021-11-22
Fix usage np.random for multiprocessing case

Dipet:
Fixed all `np.random` use cases to prevent identical values when using multiprocessing. 
More info there: #873


ternaus:
`np.random.randint(a, b)`, `a` inclusive, `b` exclusive

`random.randint(a, b)` `a` and `b` inclusive


=> we cannot just replace functions.

ternaus (approved):

PR #1080, opened 2021-12-10
Corrected off-by-one in random crop. 

### LINKED ISSUE #1075: RandomCrop never samples the last pixel in each row or column
## 🐛 Bug

RandomCrop will never sample the pixels in `image[:, -1]` or `image[-1, :]`. This is because it will only sample this value if you choose `h_start==1.0`, but you can never sample precisely `1.0`. This happens if you use something like `np.linspace(0, 1.0, n_pixels)` to determine the quantisation boundaries. Instead you should use `np.linspace(0, 1.0, n_pixels+1)` to choose quantisation boundaries.

To give a practical example: consider an image of size 4x4 and a crop of 2x2. **We expect**: we may sample a crop with the top left in any of [0, 1, 2], and the boundaries for sampling `2` is `2/3 < h_start=w_start < 1`. **We actually get**: it will actually only sample from [0, 1], and the boundaries for sampling `2` is `1 < h_start=w_start < 3/2`.

## To Reproduce

I have two ways of showing this: 
  1. Showing that sampling 0.9999 doesn't reach the end.
  2. Actually measuring locations sampled.

```python
import numpy as np
import albumentations as A

# `image` is actually a map of the coordinates, so we can see where we sampled
image = np.indices([4, 4]).transpose([1,2,0])
rc = A.RandomCrop(2, 2)

# -- Evidence 1. Showing boundary effect

print('Actual last possible crop location:')
print(image[-2, -2]) # [2 2]

for v in [0.4999, 0.5001, 0.9999, 1.0001, 1.4999, 1.5001]:
    crop = rc.apply(image, h_start=v, w_start=v)
    print(v, crop[0, 0])

# 0.4999 [0 0]
# 0.5001 [1 1]
# 0.9999 [1 1]
# 1.0001 [2 2]
# 1.4999 [2 2]
# 1.5001 [3 3]


# -- Evidence 2. Actually measuring sampled locations in practice
sampled = np.zeros((4, 4))
for _ in range(1000):
    result = rc.apply(image, **rc.get_params())
    y, x = result[0, 0]
    sampled[y, x] += 1

# How many times the top left of the sampled1 region starts
# in each spot in image
print(sampled)

# [[258. 245.   0.   0.]
#  [265. 232.   0.   0.]
#  [  0.   0.   0.   0.]
#  [  0.   0.   0.   0.]]
```

## Expected behavior

It should sample the last possible position if `n/(n+1) < h_start=w_start < 1`. In my example, we would expect it to sample a crop starting at `(2, 2)` when `2/3 < h_start=w_start < 1`.

(Note: in this case, `n = image_size - crop_size = 4 - 2 = 2`, or, more intuitively, "the highest index to sample from").

## Environment

 - Albumentations version (e.g., 0.1.8): 1.1.0
 - Python version (e.g., 3.7): 3.9.7
 - OS (e.g., Linux): Ubuntu 20.04 (using docker)
 - How you installed albumentations (`conda`, `pip`, source): pip
 - Any other relevant information: installed fresh today

Multihuntr:
Fixes #1075 

Previously: 0.9999 would sample the second last position, and 1.0 would sample the last position. But 1.0 is never randomly sampled, so the last position would never be used.

Now: 0.9999 will sample the last position in the image, and 1.0 will select a crop with the right hand edge one pixel outside the image.

**Note**: I had to change `tests/files/output_v0.4.6_without_totensor.json` to make it work. Since the crop index boundaries have potentially changed, it makes sense that we no longer always get exactly the same output. Since the change is so small, and passes the other tests, I copied the output I got with the changes and saved that to the JSON file. Is this the right procedure?


Multihuntr:
With these changes:

```python
import numpy as np
import albumentations as A

# `image` is actually a map of the coordinates, so we can see where we sampled
image = np.indices([4, 4]).transpose([1,2,0])
rc = A.RandomCrop(2, 2)

print('Actual last possible crop location:')
print(image[-2, -2]) # [2 2]

for v in [0.3332, 0.3334, 0.6665, 0.6667, 0.9999, 1.0001]:
    crop = rc.apply(image, h_start=v, w_start=v)
    print(v, crop[0, 0])

# 0.3332 [0 0]
# 0.3334 [1 1]
# 0.6665 [1 1]
# 0.6667 [2 2]
# 0.9999 [2 2]
# 1.0001 [3 3]
```

(See current boundaries in linked issue)

BloodAxe:
Hey! Thanks for your PR and investigation. I'll be looking into this PR and get back to you as soon as possible.
At the first glance it does fix the abovementioned problem, yet I feel that it may be a good opportunity to revisit the design approach we use for sampling.

Multihuntr:
Thanks for the response. Let me know if there's anything I can do that would help.

The current sampling method is very efficient. It implicitly creates quantisation boundaries by just converting to an int. Treating the range [0, 1) to be the normalised positions you can crop from without going outside the image also makes a lot of intuitive sense to me.

I'm not sure what's wrong with the current method. Is it that the [0, 1) does **not** refer to crop positions within the whole image? i.e. are you worried that people would expect `y1 = int(height + 1 * h_start)` instead?

albumentations/augmentations/crops/functional.py
```diff
@@ -12,9 +12,9 @@
 
 
 def get_random_crop_coords(height: int, width: int, crop_height: int, crop_width: int, h_start: float, w_start: float):
-    y1 = int((height - crop_height) * h_start)
+    y1 = int((height - crop_height + 1) * h_start)
```
Dipet, line 15: I think better just change `int` to `round` because this code is a little bit strange for me:
```
y1 = round((height - crop_height) * h_start)
```

Dipet: It would also be great to add a comment here on why we're using this approach so that we don't forget about these changes.

Multihuntr: `int` implicitly creates the correct boundaries. Your provided example doesn't give the same boundaries/answer. e.g. `height=9` `crop_height=4` and `h_start=0.1001`

Let me try to explain why `int` is better via visualisation. I'm having trouble explaining myself in words only. :frowning_face: 

sample-range

The largest index in this theoretical image is 8, there are 9 pixels on the number line, and 10 vertical notches on the number line. Surprisingly tricky, right?

Let's say we have an image with 9 pixels, and we're taking a crop 4 pixels large. It should be clear from the visualisation that the last index we can use is 5 (height - crop_height). Then, when defining what range `h_start` maps onto, I've colour-coded the regions of the number line which map to each index for each solution. And `h_start` is the proportion of the rainbow. You see that the `+1` increases the width on the number line to fit in the last index.

Now, I just need to put this into a small comment that doesn't confuse things further... Before I put in a commit, can I get you to check whether this comment makes sense?

```python
# h_start is [0, 1) and should map to [0, (height - crop_height)] (note inclusive)
# this is conceptually equivalent to mapping onto `range(0, (height - crop_height + 1))`
```

Dipet: It is really tricky, but the current solution also doesn't work correctly, because if you try your code from your issue with these changes it will sometimes produce an empty image.

Dipet: Thanks for explanation, it is really tricky.
I think it would also be great if you provided a link to this PR in the comment.

Dipet (approved):

Multihuntr:
The tests are passing locally. I'm not sure if there's a caching issue with the serialisation files or something?

Multihuntr:
The tests are failing because the stored file doesn't match. Is there anything special I need to do besides pulling from master before trying again? (also; bump, I guess)

Dipet:
Just merge the master branch and change reference files.

Multihuntr:
I figured out what was wrong with my previous commits. I was using a test Dockerfile that didn't have torch installed, and some of the tests are silently not run if you don't have torch installed. These tests weren't failing locally, but would fail in the CI pipeline.

I have installed `torch` and updated both the files that have tests. I have checked `flake8` doesn't have any issues. It should be good for CI now :crossed_fingers:

Issue #1079, opened 2021-12-09
Affine rotation direction inconsistant with Rotate/ShiftScaleRotate and incompatible using xyas keypoints


Multihuntr:
## 🐛 Bug

In a graph with +ve Y pointing up, +ve rotation is conventionally counter-clockwise.
Thus, in an image, with +ve Y pointing down, +ve rotation should then typically be clockwise. 

This convention is respected by `Affine`, but not `Rotate` or `ShiftScaleRotate`. This means that a pipeline with both `Affine` and `Rotate`/`ShiftScaleRotate` when using `xyas` keypoints will rotate those keypoints incorrectly.

## To Reproduce

For my examples it's easier to start with a visualisation. Pretend we have an image shaped (200, 100) (xy) and a single keypoint at (0, 0) with an angle of 45 degrees.

![rotate](https://user-images.githubusercontent.com/10515040/145353778-4a236fed-70a1-44ea-bf40-8f56ff061b88.png)

If +ve is considered clockwise, then (0,0) becomes approximately (64, -56), and we look at the green arrow. If +ve is considered counter-clockwise, then (0,0) becomes approx. (-6, 85) and we look at the blue arrow.

Here I show that :

- `Rotate` and `ShiftScaleRotate` treats +ve rotation as counter-clockwise in an image. 
- `Affine` treats +ve rotation as clockwise in an image. 
- The discrepancy between `Affine` then `Rotate` causes the rotations of the keypoints to be misinterpreted

```python
import math

import skimage.transform
import albumentations as A
import numpy as np

kp_params = A.KeypointParams(format='xya', remove_invisible=False)
img = np.zeros((100, 200))
kp1 = (0, 0, 45, 1)
transform1 = A.Compose(
    [A.Rotate((45, 45), p=1)],
    keypoint_params=kp_params
)
transform2 = A.Compose(
    [A.ShiftScaleRotate(shift_limit=(0, 0), scale_limit=(0, 0), rotate_limit=(45, 45), p=1)],
    keypoint_params=kp_params
)
transform3 = A.Compose(
    [A.Affine(rotate=[45,45], p=1)],
    keypoint_params=kp_params
)
transform4 = A.Compose(
    [
        A.Affine(rotate=[45,45], p=1),
        A.Rotate((45, 45), p=1)
    ],
    keypoint_params=kp_params
)

print(transform1(image=img, keypoints=[kp1])['keypoints'])
# [(-6, 85, 90.0, 1)]
[...]

PR #1091, opened 2021-12-15
Fix Affine wrong rotation angle

Dipet:
OpenCV and skimage works with rotation angle differently, so we need to change angle sign when we creat skimage rotation matrix.
More info there: #1079

**Master branch**
image

**Fixed**
image

Code to reproduce the problem:
```python
import cv2 as cv
import numpy as np
import albumentations as A

kp_params = A.KeypointParams(format='xya', remove_invisible=False)
img = np.zeros((100, 100))
cv.line(img, (25, 50), (75, 50), 255, 3)
cv.line(img, (75, 25), (75, 75), 255, 3)

kp = [(0, 0, 0)]


transform1 = A.Compose(
    [A.Rotate((45, 45), p=1)],
    keypoint_params=kp_params
)
transform2 = A.Compose(
    [A.ShiftScaleRotate(shift_limit=(0, 0), scale_limit=(0, 0), rotate_limit=(45, 45), p=1)],
    keypoint_params=kp_params
)
transform3 = A.Compose(
    [A.Affine(rotate=[45, 45], p=1)],
    keypoint_params=kp_params
)
transform4 = A.Compose(
    [
        A.Affine(rotate=[45, 45], p=1),
        A.Rotate((45, 45), p=1)
    ],
    keypoint_params=kp_params
)


count = 231
def show(res, name):
    global count
    print(res["keypoints"][0][-1])
    img = res["image"]
    plt.subplot(count, title=f"{name} {res['keypoints'][0][-1]}")
    plt.imshow(img, cmap="gray")
    count += 1


show({"image": img, "keypoints": kp}, "original")
show(transform1(image=img, keypoints=kp), "Rotate")
show(transform2(image=img, keypoints=kp), "ShiftScaleRotate")
show(transform3(image=img, keypoints=kp), "Affine")
show(transform4(image=img, keypoints=kp), "Affine + Rotate")
plt.show()
```


albumentations/augmentations/geometric/functional.py
```diff
@@ -461,7 +461,9 @@ def keypoint_affine(
 
     x, y, a, s = keypoint[:4]
     x, y = skimage.transform.matrix_transform(np.array([[x, y]]), matrix.params).ravel()
-    a += rotation2DMatrixToEulerAngles(matrix.params[:2])
+    # Look to issue https://github.com/albumentations-team/albumentations/issues/1079
+    # OpenCV and skimage work differently with angle, so we need change the sign of angle
+    a -= rotation2DMatrixToEulerAngles(matrix.params[:2])
```
Multihuntr, line 466: Ideally, to fix this problem, we should only have to modify the code that creates the matrices; there should be no need to modify code that use the matrices in any way because the discrepancy is just in the definition of "+ve" for a rotation, and everything else should just work.

Thus, I wonder why this is necessary, and I am worried this change is incorrect and will cause problems for other use cases. I will do some experiments tomorrow to try to find some, or convince myself that it is in fact correct (I am going to test shearing and non-square images at 20 degrees; testing square images at 45 degrees can hide all sorts of bugs, as I've found out in my own projects)

In any case, if this change is correct, then it should be applied in the `rotation2DMatrixToEulerAngles` function so that it works for PerspectiveTransform too, right?

Multihuntr: I thought about it more, and mathematically speaking, `rotation2DMatrixToEulerAngles` should definitely return the negative of what it currently does. Currently it operates in mathland, but it should operate in imageland to be compatible with all other rotations. e.g. we would expect (y, x) = (-1, 1) to give 45 degrees in imageland.

```python
math.degrees(math.atan2(1, 1))
# 45.0
math.degrees(math.atan2(-1, 1))
# -45.0
```

Multihuntr:
After a little bit of experimentation, I found that keypoint angles are changing where they are pointing in the image when using either shear or non-square scaling. Personally, I would consider this a bug because I would expect that the angle has a relationship with with other pixels in the image, but it might be debatable.

<details>
<summary> Code </summary>

```python
import math

import cv2
import numpy as np
import albumentations as A

kp_params = A.KeypointParams(format='xyas', remove_invisible=False, angle_in_degrees=True)
img = np.ones((100, 150, 3))*255

kps = [
    (0, 0, 0, 70),
    (0, 100, 0, 70),
    (150, 0, 180, 70),
    (150, 100, 180, 70),
    (25, 50, 0, 50),
    (75, 25, -90, 50)
]

mode = cv2.BORDER_CONSTANT
transform1 = A.Compose(
    [A.Affine(rotate=[10, 10], mode=mode, p=1)],
    keypoint_params=kp_params
)
transform2 = A.Compose(
    [A.Affine(shear=[20, 20], rotate=[10, 10], mode=mode, p=1)],
    keypoint_params=kp_params
)
transform3 = A.Compose(
    [A.Affine(scale={'x': [1, 1], 'y':[2,2]},
              rotate=[10, 10],
              mode=mode,
              p=1)],
    keypoint_params=kp_params
)
transform4 = A.Compose(
    [A.Affine(scale={'x': [1, 1], 'y':[2,2]},
              shear=[20, 20],
              rotate=[10, 10],
              mode=mode,
              p=1)],
    keypoint_params=kp_params
)
transform5 = A.Compose(
    [A.Perspective(scale=(0.1, 0.1), p=1)],
    keypoint_params=kp_params
)


def draw_keypoint(img, kp):
    x, y, a, s = kp
    dx = math.cos(math.radians(a))*s
    dy = math.sin(-math.radians(a))*s

    cv2.line(img, (int(x), int(y)), (int(x+dx), int(y+dy)), (0, 255, 0), 3)

count = 231
def show(res, name):
    global count
    img = res["image"]
    for kp in res['keypoints']:
        draw_keypoint(img, kp)
    plt.subplot(count, title=f"{name} {res['keypoints'][0][-1]}")
    plt.imshow(img)
    count += 1

fig = plt.figure(figsize=(15, 8))
show({"image": img.copy(), "keypoints": kps}, "original")
show(transform1(image=img, keypoints=kps), "Affine rotate")
show(transform2(image=img, keypoints=kps), "Affine rotate shear")
show(transform3(image=img, keypoints=kps), "Affine rotate scale")
show(transform4(image=img, keypoints=kps), "Affine rotate shear scale")
show(transform5(image=img, keypoints=kps), "Perspective")
plt.show()
fig.savefig('out.png')
```
</details>

out

So, I would say that this pull request works for rotating non-square images at non-45 degree angles, so it solves the initial issue for `Affine`. If it is shown to work for Perspective, too, then I'd consider the initial issue solved. I would then make a separate issue for the non-square scale/shear + rotation problems in keypoint angles. Is that acceptable?

albumentations/augmentations/geometric/functional.py
Multihuntr: Oh, and the comment you've put is misleading. The discrepancy between OpenCV and skimage is the discrepancy between mathland and imageland. The discrepancy here is also because of that. But the discrepancy here is not because OpenCV and skimage disagree.

It's an "All A are B, but not all B are A" situation.

Dipet:
Before
image


After
image

Code to reproduce:
```python
import random

import cv2 as cv
import numpy as np
import albumentations as A

kp_params = A.KeypointParams(format='xya', remove_invisible=False)
img = np.zeros((100, 100, 3), dtype=np.uint8)
img[..., 0] = 255
cv.line(img, (25, 50), (75, 50), (255, 255, 255), 3)
cv.line(img, (75, 25), (75, 75), (255, 255, 255), 3)

kp = [(0, 0, 0)]


angle = 30
shift = 0
scale = 0
rotate = A.Rotate((angle, angle), p=1, border_mode=cv.BORDER_CONSTANT, value=0)
ssr = A.ShiftScaleRotate(
    shift_limit=(shift, shift),
    scale_limit=(0, 0),
    rotate_limit=(angle, angle),
    p=1, border_mode=cv.BORDER_CONSTANT, value=0
)
affine = A.Affine(
    rotate=angle,
    scale=1,
    translate_percent=shift,
    shear=None,
    p=1
)
perspective = A.Perspective(scale=[0.2, 0.2], p=1)
transform1 = A.Compose(
    [rotate],
    keypoint_params=kp_params
)
transform2 = A.Compose(
    [ssr],
    keypoint_params=kp_params
)
transform3 = A.Compose(
    [affine],
    keypoint_params=kp_params
)
transform4 = A.Compose(
    [
        affine,
        rotate
    ],
    keypoint_params=kp_params
)
random.seed(1)
transform5 = A.Compose(
    [perspective],
    keypoint_params=kp_params
)
random.seed(1)
transform6 = A.Compose(
    [
        perspective,
        rotate
    ],
    keypoint_params=kp_params
)


count = 331
plt.figure(figsize=(15, 15))
def show(res, name):
    global count
    print(res["keypoints"][0][-1])
    img = res["image"]
    plt.subplot(count, title=f"{name} {res['keypoints'][0][-1]:.2f}")
    plt.imshow(img)
    count += 1


show({"image": img, "keypoints": kp}, "original")
show(transform1(image=img, keypoints=kp), "Rotate")
show(transform2(image=img, keypoints=kp), "ShiftScaleRotate")
show(transform3(image=img, keypoints=kp), "Affine")
show(transform4(image=img, keypoints=kp), "Affine + Rotate")
show(transform5(image=img, keypoints=kp), "Perspective + Rotate")
show(transform6(image=img, keypoints=kp), "Perspective + Rotate")
plt.show()
```

PR #1094, opened 2021-12-16
[Code restructure] Move dropout classes to separate folder

BloodAxe:


albumentations/augmentations/dropout/channel_dropout.py
```diff
@@ -0,0 +1,64 @@
...
+        p (float): probability of applying the transform. Default: 0.5.
+
+    Targets:
+        image
+
+    Image types:
+        uint8, uint16, unit32, float32
+    """
+
+    def __init__(self, channel_drop_range=(1, 1), fill_value=0, always_apply=False, p=0.5):
```
ternaus, line 26: type hints?

README.md
```diff
@@ -177,18 +177,18 @@ Spatial-level transforms will simultaneously change both an input image as well
...
+| [GridDropout](https://albumentations.ai/docs/api_reference/augmentations/dropout/grid_dropout/#albumentations.augmentations.dropout.grid_dropout.GridDropout)                   | ✓     | ✓     |        |           |
 | [HorizontalFlip](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.HorizontalFlip)                                 | ✓     | ✓     | ✓      | ✓         |
 | [Lambda](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.Lambda)                                                 | ✓     | ✓     | ✓      | ✓         |
 | [LongestMaxSize](https://albumentations.ai/docs/api_reference/augmentations/geometric/resize/#albumentations.augmentations.geometric.resize.LongestMaxSize)                     | ✓     | ✓     | ✓      | ✓         |
-| [MaskDropout](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.MaskDropout)                                       | ✓     | ✓     |        |           |
```
Dipet, line 191: @creafz Will the new links be updated automatically on the site?

albumentations/augmentations/__init__.py
```diff
@@ -6,10 +6,19 @@
...
 from .geometric.resize import *
 from .geometric.rotate import *
 from .geometric.transforms import *
+
+from .dropout.channel_dropout import *
+from .dropout.coarse_dropout import *
+from .dropout.cutout import *
+from .dropout.functional import *
+from .dropout.grid_dropout import *
+from .dropout.mask_dropout import *
```
Dipet, line 20: Maybe better to move this imports into `dropout.__init__` and there import them like `from .drop import *`

Dipet:
Are you sure splitting all classes into their unique files is a good idea? It looks Java style and I think it is overkill :). Also, it is a little bit strange that we have a unique file for each transform and single `functional` file,

BloodAxe:
I'm not sure that proposed approach of splitting code is the right and optimal. However I personally find this way of splitting classes is definitely better than having a huge `transforms.py` class. I'm a huge fan of bite-sized content. So from this perspective having a one file per class is justified move (for me at least). 

Regarding `functional.py` - yes. Hovewer, there is a plan :) There will be a second PR to add a support of bboxes to dropout augs, with somewhat generalized approach that is to be reused by many classes.

Dipet:
I also don't like huge files like `transforms.py`, but lots of small independent files are not an option either. This is not a problem at the moment, because we don't have that many files. But if you chose this approach, it might be better to import the transforms into `__init__.py` directly without using `__all__` and `from module import *`?

BloodAxe:
> I also don't like huge files like `transforms.py`, but lots of small independent files are not an option either. This is not a problem at the moment, because we don't have that many files. But if you chose this approach, it might be better to import the transforms into `__init__.py` directly without using `__all__` and `from module import *`?

Ideally, they should be imported as follows: `from .dropout import *` in the main `__init__.py`. 
Yet the issue I'm facing now with that approach is that pytest cannot resolve `F.vflip` for some reason after I introduce this change! Still investigating this issue, but looks like we have something is partially broken with our imports scheme...

Dipet:
> Yet the issue I'm facing now with that approach is that pytest cannot resolve F.vflip for some reason after I introduce this change! Still investigating this issue, but looks like we have something is partially broken with our imports scheme...

Yes, this is a problem with `from module import *`. I think we need to think about changing logic with import using `__all__` to direct import names from modules.

Dipet (approved):

README.md
creafz: This will require site rebuilding against the updated version of Albumentations. I will do it after we merge this PR.

albumentations/augmentations/functional.py
```diff
@@ -19,6 +95,19 @@
...
+    np.uint16: cv2.CV_16U,
+    np.int32: cv2.CV_32S,
+    np.float32: cv2.CV_32F,
+    np.float64: cv2.CV_64F,
+    np.dtype("uint8"): cv2.CV_8U,
+    np.dtype("uint16"): cv2.CV_16U,
+    np.dtype("int32"): cv2.CV_32S,
+    np.dtype("float32"): cv2.CV_32F,
+    np.dtype("float64"): cv2.CV_64F,
+}
```
creafz, line 109: As we split the code into smaller pieces may be it is better to move these constants (such as `MAX_VALUES_BY_DTYPE` and `NPDTYPE_TO_OPENCV_DTYPE`) to a standalone file?

However, I am also ok with the current approach.

creafz (approved):

PR #1095, opened 2021-12-17
update documentation for RandomCropNearBBox

Juphex:
fixes the documentation


Dipet:
This PR doesn't change the meaning of docstring. In this case, `label_fields` are unnecessary.

Juphex:
You may be right, however I guess one need the brackets of the list. Otherwise one gets an not iterable error.

Juphex:
https://github.com/albumentations-team/albumentations/pull/1096

PR #1104, opened 2022-01-09
Add keep ratio flag to Affine transform

### LINKED ISSUE #1011: Affine transformation handles `scale` tuple incorrectly
## 🐛 Bug

I'd like to use the [`Affine` transformation](https://albumentations.ai/docs/api_reference/augmentations/geometric/transforms/#albumentations.augmentations.geometric.transforms.Affine) for a random zoom-in/zoom-out behaviour. According to the documentation, when passing a tuple to the `scale` argument, the scaling factor should be the same across the `x` and `y` axes:

> If a tuple (a, b), then a value will be uniformly sampled per image from the interval [a, b]. That value will be used identically for both x- and y-axis.

However, that's not the case:

Screenshot 2021-09-13 at 08 50 53

## To Reproduce

1. Run an image through the following transformation:
```
albumentations.Affine(scale=(0.5, 2), p=1)
```

## Expected behavior

The sampled scaling factor should be the same for the `x` and `y` axes. As you can see in the image above, that's not the case.

I believe the cause of this bug comes from [this line](https://github.com/albumentations-team/albumentations/blob/162dc0657c065a64cd0cc795583598df88b6441b/albumentations/augmentations/geometric/transforms.py#L548). That function doesn't seem to handle `Sequence[float]` at all.

## Environment

 - Albumentations version (e.g., 0.1.8): 1.0.3
 - Python version (e.g., 3.7): 3.9.7
 - OS (e.g., Linux): macOS
 - How you installed albumentations (`conda`, `pip`, source): pip
 - Any other relevant information:

i-aki-y:
In this PR, I add a feature to keep the original aspect ratio in Affine transform (fixes #1011).

As pointed out in issue #1011, the current Affine transform's random scaling is applied for both axes independently even if the case `scale=(0.5, 1.0)`.
According to the suggestion of @Dipet in #1011, I added a keep_ratio flag to the Affine transform.

I appreciate it if you review this PR.


Dipet (approved):

jamesljlster:
I need this feature as well.
When I'm doing data transformations for rotated objects, I want to keep their original geometrical feature.
Although I set the shear parameter of affine to None, the result of rotated objects still seem to be sheared.

Original data:
image

Transformed one:
image

Issue #1025, opened 2021-10-06
Ensure input image for A.Compose is continuous 


BloodAxe:
## Enhancement

A discussion in #614 revealed a good place for improvement - we should ensure that input image is continuous upon start of the augmentation pipeline. This could be implemented by adding `image = np.ascontiguousarray(image)` to image and mask targets.

A proposed place to add this call - somewhere at the beginning of `A.Compose.__call__`.

Dipet:
I think we need to call `np.ascontiguousarray` at the beginning of `A.Compose.__call__` and after each call of `apply_with_params`, to be sure that we get the valid image after each transform. Because we have non contiguousarray array after channel shuffle.

ternaus:
Fixed

PR #1117, opened 2022-01-31
Fix: RandomShadow input type wrong

### LINKED ISSUE #614: RandomShadow input type wrong
## 🐛 Bug

Weather transformation. For RandomRain, RandomSnow, RandomSunFlare the inputs are just numpy uint8. However, RandomShadow does not allow the same input format.

## To Reproduce

```
albu_shadow = albu.RandomShadow(p=1, num_shadows_lower=1, num_shadows_upper=1, shadow_dimension=5, shadow_roi=(0, 0.5, 1, 1))

x_np = albu_shadow(image=x_np)['image']
```

Error:
```
TypeError: Expected Ptr<cv::UMat> for argument img
```

## Expected behavior
It should take the same uint8 numpy array as input.

## Environment

 - Albumentations version: 0.4.5
 - Python version: 3.7.6
 - OS (e.g., Linux): Linux
 - How you installed albumentations: pip

i-aki-y:
This PR ensures input image and mask array are contiguous in memory space to avoid incompatibility with cv:Mat data type used in OpenCV. (close #614 #1025)

As proposed in #1025, `np.ascontiguousarray` are inserted in the `Compose.__call__()`.
Since some transforms can output a non-contiguous array, `np.ascontiguousarray` is also applied after each transforms.

Note that the error is still raised if users apply a transform to a non-contiguous input individually (applying without Compose). In such a case, ensuring a contiguous input is a user's responsibility.

I think fixes for some base classes like`DualTransform` and `ImageOnlyTransform` (or `BasicTransform`) are needed to cover the individual transform case if needed.

I appreciate any comments and suggestions.


albumentations/core/composition.py
```diff
@@ -201,11 +201,13 @@ def __call__(self, *args, force_apply: bool = False, **data) -> typing.Dict[str,
         for p in self.processors.values():
             p.preprocess(data)
 
+        data = self._make_targets_contiguous(data)  # some transforms require contiguous input
         for idx, t in enumerate(transforms):
             data = t(force_apply=force_apply, **data)
 
             if check_each_transform:
                 data = self._check_data_post_transform(data)
+            data = self._make_targets_contiguous(data)
```
Dipet, line 210: I think we do not need this check there because all transforms must return contiguous arrays. Better to write a test to check that each transform returns contiguous data.
Look at [test_image_only_augmentations_with_float_values](https://github.com/albumentations-team/albumentations/blob/master/tests/test_augmentations.py#L79) for an example

albumentations/core/composition.py
```diff
@@ -271,6 +273,13 @@ def _check_args(self, **kwargs) -> None:
             if internal_data_name in check_bbox_param and self.processors.get("bboxes") is None:
                 raise ValueError("bbox_params must be specified for bbox transformations")
 
+    def _make_targets_contiguous(self, data):
+        targets = ["image", "mask"]
+        for target in targets:
+            if target in data and isinstance(data[target], np.ndarray) and not data[target].flags["C_CONTIGUOUS"]:
+                data[target] = np.ascontiguousarray(data[target])
+        return data
```
Dipet, line 281: Thank you for your contribution.
I think it is better to write a generalized function that will covert all np.ndarrays to a contiguous format.
```python
@staticmethod
def _make_targets_contiguous(data: typing.Dict[str, typing.Any]) -> typing.Dict[str, typing.Any]:
    result = {}

    for key, value in data.items():
        if isinstance(value, np.ndarray):
            value = np.ascontiguousarray(value)
        result[key] = value

    return result
```

Also, we do not need this check `data[target].flags["C_CONTIGUOUS"]` because if the array is contiguous `np.ascontiguousarray` will return it unchanged and will not create a copy.

Dipet (changes requested):

albumentations/core/composition.py
i-aki-y: Thank you for your comment.
OK, I think I got how the test works.
I will examine the violating transforms.

albumentations/core/composition.py
i-aki-y: Yes. you're right! I will fix it.

i-aki-y:
I think I have made a test case to check all transforms. And I found that some transforms return non-contiguous output, the pytest summary is shown below.
Now I will fix these transforms to return contiguous results.

List of transforms that return non-contiguous output:
```
======================================================= short test summary info ========================================================
FAILED tests/test_augmentations.py::test_return_contiguous_ndarray[CenterCrop-params4] - assert False
FAILED tests/test_augmentations.py::test_return_contiguous_ndarray[ChannelShuffle-params6] - assert False
FAILED tests/test_augmentations.py::test_return_contiguous_ndarray[Crop-params9] - assert False
FAILED tests/test_augmentations.py::test_return_contiguous_ndarray[CropNonEmptyMaskIfExists-params11] - assert False
FAILED tests/test_augmentations.py::test_return_contiguous_ndarray[FDA-params16] - assert False
FAILED tests/test_augmentations.py::test_return_contiguous_ndarray[RandomCrop-params48] - assert False
FAILED tests/test_augmentations.py::test_return_contiguous_ndarray[RandomCropNearBBox-params49] - assert False
FAILED tests/test_augmentations.py::test_return_contiguous_ndarray[Transpose-params76] - assert False
```

i-aki-y:
I applied the result of the review comments.

I made a test to check all augmentation and fixed some of them to return contiguous ndarray.

I appreciate it if you review this one.

Dipet (approved):

BloodAxe:
Hi! Thanks for your PR. This is indeed valuable improvement for the library. Yet i feel it may have impact on the performance.
Consider a following pipeline:
```
A.Compose([A.RandomCrop(), A.HorisontalFlip(), A.VerticalFlip()])
```
There are three consecutive augmentations that produce non-contiguous output. With the proposed change we will make 3 copies of the sliced data simply to make another slice. Which is clearly a waste of processing time. 

The initial reason why we want to have contiguous data is that OpenCV fails to handle image slices as input. With this knowledge in mind, I suggest to consider another solution:

1) Extend our unit tests to test all augmentations for non-contiguous inputs
2) For those, who fails, add a decorator function similar to `@preserve_shape` to ensure input for that function is contiguous:
```
def ensure_contiguous(func):

    @wraps(func)
    def wrapped_function(img, *args, **kwargs):
        img = np.require(img, requirements=np.C_CONTIGUOUS)
        result = func(img, *args, **kwargs)
        return result

    return wrapped_function
```
This change would ensure that we make a copy only when necessary and would be optimal from performance point of view.

3) As a final step - call a `_make_targets_contiguous` to ensure that outputs from the whole pipeline are also contiguous.

i-aki-y:
@BloodAxe Thank you for your suggestion. I will try your idea.

i-aki-y:
@BloodAxe @Dipet  I fixed the PR.

I modified the test case to detect transforms that raise an error with non-contiguous input instead of checking contiguous output. 
As a result of the test, I found that the `add_shasow` was the only function with a problem. 
So in this new approach, fixes became simpler than the previous version.

I appreciate it if you review this modification.

BloodAxe:
Thanks @i-aki-y, looking into it.

BloodAxe (approved):

BloodAxe:
Thank your for your contribution. Merged)

PR #1159, opened 2022-04-20
Docs: Affine - improve `fit_output` documentation

mys007:
I believe the documentation of `fit_output` in `Affine` class can be improved:
* "It will however negate translation and scaling." seems to be wrong, the correcting matrix affects only translation: https://github.com/albumentations-team/albumentations/blob/89a675cbfb2b76f6be90e7049cd5211cb08169a5/albumentations/augmentations/geometric/transforms.py#L708 
* I find the description with words like "zoom in" misleading and I miss clearly stating the information that image output shape is affected.
* The description is not in sync with `fit_output` in `Perspective` class.

Therefore, I suggest complete rewording of the description as in this PR.


Dipet:
Please use [pre-commit hooks](https://albumentations.ai/docs/contributing/)
Please fix flake8 issues. You can check flake8 with this commands:
1. `pip install flake8==3.9.2 flake8-docstrings==1.6.0` - install flake8
2. `flake8` - call checker

Dipet (approved):

PR #1161, opened 2022-04-20
Jk/rotation augmentation

joonkAkasha:
implementation for ICCV 2021 paper algorithm 1

https://arxiv.org/pdf/2109.13488.pdf

algorithm.py was used for testing purposes and will be removed after the PR is approved

### DIFF
--- a/albumentations/augmentations/geometric/functional.py
+++ b/albumentations/augmentations/geometric/functional.py
@@ -134,20 +134,35 @@ def bbox_rotate(bbox, angle, rows, cols):
         A bounding box `(x_min, y_min, x_max, y_max)`.
 
     """
+
     x_min, y_min, x_max, y_max = bbox[:4]
-    scale = cols / float(rows)
-    x = np.array([x_min, x_max, x_max, x_min]) - 0.5
-    y = np.array([y_min, y_min, y_max, y_max]) - 0.5
-    angle = np.deg2rad(angle)
-    x_t = (np.cos(angle) * x * scale + np.sin(angle) * y) / scale
-    y_t = -np.sin(angle) * x * scale + np.cos(angle) * y
-    x_t = x_t + 0.5
-    y_t = y_t + 0.5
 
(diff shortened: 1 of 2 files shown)


ternaus:
You can add this type of rotation to the existing rotation algorithms.

For this you need:

1. Add the desired functionality to the existing rotations transforms.
2. Write tests.
3. Make sure that everything works.

victor1cea:
@joonkAkasha do still plan on working on this PR?

victor1cea:
@ternaus I would assume that OP has abandoned this PR, so I could try to implement this rotation method. Could I ask you some implementation details?
Regarding point 1:
I was thinking of adding to `bbox_rotate()` a new argument `method` that can be used to choose between the "largest box" rotation method (current one) and "ellipsoid" rotation method. Would this be ok?
Regarding 2:
There is an [official repo](https://github.com/akasha-imaging/ICCV2021) that I could used to generate some test cases against.
Another thing:
`bbox_rotate()` is basically `bbox_shift_scale_rotate()`, but with no scaling and shifting. However there are 2 explicit implementations. Why not call `bbox_shift_scale_rotate()` with no scaling and shifting inside of `bbox_rotate()`?

https://github.com/albumentations-team/albumentations/blob/0d98b4f46f04db271203381742c6abe113a8f269/albumentations/augmentations/geometric/functional.py#L123-L149

https://github.com/albumentations-team/albumentations/blob/0d98b4f46f04db271203381742c6abe113a8f269/albumentations/augmentations/geometric/functional.py#L209-L229

There is even a this test case

https://github.com/albumentations-team/albumentations/blob/0d98b4f46f04db271203381742c6abe113a8f269/tests/test_functional.py#L113-L116

Dipet:
1. Yes, looks ok to have `method` argument.
2. Its's a good idea to reuse `bbox_shift_scale_rotate` inside `bbox_rotate` :+1:

Dipet:
Done in #1203

Issue #1165, opened 2022-05-02
numpy randn is used in random_utils.rand


hir0sa:
## 🐛 Bug
In the recently introduced `random_utils` module (#1070),  `rand` function uses numpy `randn` ( _N(0, 1)_ ) internally which results in unexpected behavior of some functions. For example,   [elastic_transform](https://github.com/albumentations-team/albumentations/blob/89a675cbfb2b76f6be90e7049cd5211cb08169a5/albumentations/augmentations/geometric/functional.py#L234) uses `random_utils.rand` as _U(0, 1)_. 

https://github.com/albumentations-team/albumentations/blob/880c1aaed10ab74cfe851496a476c1c34cadcd0f/albumentations/random_utils.py#L29-L32

 https://github.com/albumentations-team/albumentations/blob/89a675cbfb2b76f6be90e7049cd5211cb08169a5/albumentations/augmentations/geometric/functional.py#L295

## Expected behavior
`random_utils.rand` should use numpy `rand` ( _U(0, 1)_ )  instead of `randn` ( _N(0, 1)_ ).

## Environment

 - Albumentations version (e.g., 0.1.8): master [89a675c](https://github.com/albumentations-team/albumentations/commit/89a675cbfb2b76f6be90e7049cd5211cb08169a5)
 - Python version (e.g., 3.7): 3.8
 - OS (e.g., Linux): Linux
 - How you installed albumentations (`conda`, `pip`, source): pip

Dipet:
Fixed in #1166

PR #1166, opened 2022-05-02
Fix rand, add randn

hir0sa:
This PR fixed rand issue described in #1165 
- Fix `ranom_utils.rand` to use numpy `rand` 
- Add `random_utils.randn` to use numpy `randn`

### DIFF
--- a/albumentations/random_utils.py
+++ b/albumentations/random_utils.py
@@ -29,13 +29,13 @@ def uniform(
 def rand(d0: NumType, d1: NumType, *more, random_state: Optional[np.random.RandomState] = None, **kwargs) -> Any:
     if random_state is None:
         random_state = get_random_state()
-    return random_state.rand(d0, d1, *more, **kwargs)
+    return random_state.rand(d0, d1, *more, **kwargs)  # type: ignore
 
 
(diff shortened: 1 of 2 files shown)


albumentations/random_utils.py
```diff
@@ -29,7 +29,13 @@ def uniform(
...
     if random_state is None:
         random_state = get_random_state()
-    return random_state.randn(d0, d1, *more, **kwargs)  # type: ignore
+    return random_state.rand(d0, d1, *more, **kwargs)
+
+
+def randn(d0: NumType, d1: NumType, *more, random_state: Optional[np.random.RandomState] = None, **kwargs) -> Any:
+    if random_state is None:
+        random_state = get_random_state()
+    return random_state.randn(d0, d1, *more, **kwargs)
```
victor1cea, line 38: add `# type: ignore` after each return statement (as before) to make an exception for type checking in mypy

victor1cea (changes requested):

albumentations/random_utils.py
hir0sa: Thanks, I added the statement.

Dipet (approved):

Issue #1175, opened 2022-06-01
slant_upper and slant_lower has no effect in RandomRain


mucsacso03:
## 🐛 Bug

I'm using the RandomRain transformation and selecting any slant_upper and slant_lower value won't have any effect on the outcome of the rain. The rain looks always the same.

Code:
`x = RandomRain(slant_lower=10, slant_upper=20, blur_value=3, p=1, rain_type='drizzle')(image=x)["image"]`

## Environment

 - Albumentations version: 1.1.0
 - Python version: 3.8.10
 - OS: Linux
 - How you installed albumentations (`conda`, `pip`, source): pip

Dipet:
Fixed in #1179

PR #1179, opened 2022-06-07
added slant param to params dependent on target

victor1cea:
This PR fixes a bug mentioned in #1175. 
The slant param has now an effect in RandomRain.


Dipet (approved):

victor1cea:
It seems that you need to update `Black` to a newer version in order to work with  newest `click==8.1.3`.
Or downgrade `click` to a previous version.

see: https://stackoverflow.com/a/71674345/11457061

Issue #852, opened 2021-03-09
ReplayCompose ColorJitter params not correct


Trud09:
## To Reproduce

1. Run replay compose with color jitter
2. Try to access the params

Example Outputs:
Wrong:
ColorJitter:{'transforms': [<function ColorJitter.get_params.<locals>.<lambda> at 0x000001E029835D90>, <function ColorJitter.get_params.<locals>.<lambda> at 0x000001E029835C80>, <function ColorJitter.get_params.<locals>.<lambda> at 0x000001DF92FF6378>, <function ColorJitter.get_params.<locals>.<lambda> at 0x000001E029835E18>]}

Right:
GaussianBlur:{'ksize': 3, 'sigma': 0.0}

## Environment

Albumentations version (e.g., 0.1.8): 0.5.2
Python version (e.g., 3.7): 3.6
OS (e.g., Linux): Windows
How you installed albumentations (conda, pip, source): pip

Dipet:
I don't understand what's wrong. Replay compose does not work in your case or what?
For me it is works ok. And yes, parameters are list of lambdas https://github.com/albumentations-team/albumentations/blob/ae50578a46e5981eccaafee311a3ee0b4b7992b2/albumentations/augmentations/transforms.py#L2834

Dipet:
Fixed in #1199

PR #1199, opened 2022-06-22
Fix ColorJitter to show applied parameters in ReplayCompose

zakajd:
Hi, in continuation of discussion #852 

Did a small bug fix. Feel free to merge if it looks OK.

p.s. Increased black version, because otherwise checks failed with 
`cannot import name '_unicodefun' from 'click' ` error. Solution from [StackOverflow](https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click)


Dipet:
Please, resolve branch conflict

Dipet (approved):

PR #1203, opened 2022-06-27
Implement Ellipse Method For Bounding Box Rotation

victor1cea:
This PR implements the rotation method proposed in  #1161 by [@joonkAkasha](https://github.com/joonkAkasha).
Used [this script](https://gist.github.com/victor1cea/7d1bfe9f2affba9d534a7fa00ed5a70b) for showcasing the new method.
Red - original bounding boxes
Blue - rotated with largest box method
Green - rotated with ellipse method
New method is also available for ShiftScaleRotate
Second commit is just a black formatting.
rotated_bbs

### DIFF
--- a/albumentations/augmentations/geometric/transforms.py
+++ b/albumentations/augmentations/geometric/transforms.py
@@ -63,6 +63,7 @@ class ShiftScaleRotate(DualTransform):
         mask_value=None,
         shift_limit_x=None,
         shift_limit_y=None,
+        rotate_method="largest_box",
         always_apply=False,
         p=0.5,
     ):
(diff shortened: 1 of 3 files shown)


albumentations/augmentations/geometric/transforms.py
```diff
@@ -93,8 +100,8 @@ def get_params(self):
             "dy": random.uniform(self.shift_limit_y[0], self.shift_limit_y[1]),
         }
 
-    def apply_to_bbox(self, bbox, angle, scale, dx, dy, **params):
-        return F.bbox_shift_scale_rotate(bbox, angle, scale, dx, dy, **params)
+    def apply_to_bbox(self, bbox, angle, scale, dx, dy, rotate_method="largest_box", **params):
+        return F.bbox_shift_scale_rotate(bbox, angle, scale, dx, dy, rotate_method, **params)
```
Dipet, line 104: Don't add a `method` as an argument, just specify it using `method=self.method`.

Dipet (changes requested):

PR #1214, opened 2022-07-07
add `crop_border` option to Rotate

### LINKED ISSUE #1212: Rotate and Crop within borders
**Problem:** 
Currently all Rotation augmentations introduce boundary where pixel values are not well-defined. I would like to have an option to make a largest possible crop within rotated image. This issue has been discussed before, but still wasn't implemented: #266 #1154 
There is a code [from StackOverflow](https://stackoverflow.com/questions/16702966/rotate-image-and-crop-out-black-borders) which calculates such crop, and gives following results:
Rotate 5°
image
Rotate 5° and take largest crop
image
Rotate 10°
image
Rotate 10° and take largest crop
image

**Solution:**
I'm willing to make a PR with this change but want to discuss implementation first. 

In my understanding the most simple way would be to add a new flag to `SafeRotate` augmentation. I propose the name `crop_border`. And docstring saying `crop_border (bool): If True would make a largest possible crop within rotated image`. 
It would by implemented by adding such import `from ..crops import functional as FCrops` and using crop functions inside `SafeRotate` class.

bonlime:
Closes #1212 
I ended up adding this to Rotate instead of SafeRotate. No breaking changes. Added simple test, but can add more if needed. 

Extreme angles: 45°. False / True
image
image

90°. False / True
image

image

### DIFF
--- a/albumentations/augmentations/geometric/rotate.py
+++ b/albumentations/augmentations/geometric/rotate.py
@@ -6,6 +6,7 @@ import cv2
 import numpy as np
 
 from ...core.transforms_interface import DualTransform, to_tuple
+from ..crops import functional as FCrops
 from . import functional as F
 
 __all__ = ["Rotate", "RandomRotate90", "SafeRotate"]
@@ -63,6 +64,7 @@ class Rotate(DualTransform):
                     list of float): padding value if border_mode is cv2.BORDER_CONSTANT applied for masks.
         method (str): rotation method used for the bounding boxes. Should be one of "largest_box" or "ellipse".
             Default: "largest_box"
+        crop_border (bool): If True would make a largest possible crop within rotated image
         p (float): probability of applying the transform. Default: 0.5.
 
     Targets:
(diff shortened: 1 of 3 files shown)


albumentations/augmentations/geometric/rotate.py
```diff
@@ -90,27 +93,86 @@ def __init__(
...
+        if self.crop_border:
+            bbox_out = FCrops.bbox_crop(bbox_out, x_min, y_min, x_max, y_max, rows, cols)
+        return bbox_out
+
+    def apply_to_keypoint(
+        self, keypoint, angle=0, x_min=None, x_max=None, y_min=None, y_max=None, cols=0, rows=0, **params
+    ):
+        keypoint_out = F.keypoint_rotate(keypoint, angle, rows, cols, **params)
+        if self.crop_border:
+            keypoint_out = FCrops.crop_keypoint_by_coords(keypoint_out, (x_min, x_max, y_min, y_max))
```
jonkoi, line 126: Hi,
I think the order should be changed to (x_min, y_min, x_max, y_max), which is the order in crops.functional

bonlime: Good point, my bad, didn't notice

PR #1258, opened 2022-08-18
Rename method to rotate_method in Rotate

Dipet:
Fixes #1239
We added the ability to select `rotate_method` for `Rotate` and `ShiftScaleRotate` after #1203 
This changes were released in the previous version of the library. Because previous version was minor and I do not think many people used this feature, I suggest using the same name for this param in both transforms to keep consistency,


creafz (approved):

AntonioVispi:
Greetings,
I wanted to ask for confirmation. I have previously used the rotate function with the old 'method' on some images. After some time, I rerun the rotate function with 'rotate_method' on the same photos. In both cases 'crop_border' was True. However, I noticed that the resulting crops are different. Is it possible that in the change of the function you also have changed how much the images are cropped? Thanks in advance.

Dipet:
Rotate sample random rotation angle, so yes, they are might be different. Try to fix random seed and check result:
```python
import random
random.seed(0)
```

Dipet:
This PR do not change the logic of rotations. It is just change name of argument.
Also, this param used only for bbox, image is not affected by this argument.
Coult you provide reproducible example and library versions, on which you apply augmentation.

Issue #1263, opened 2022-08-26
Support of float32 dtype for `InvertImg`


Dipet:
Add correct work with float32 dtype for [`InvertImg`](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.InvertImg)

dmitrie-ai:
Hey, I would like to take this on.
My guess is that the purpose of `float p` is to be the probability of returning the inverted image. Is that correct?
And if so, in what circumstances is there a need for this feature?
Thank you

dmitrie-ai:
After implementing it as such, one of the tests fail. I think this is because the test for InvertImg doesn't account for this probability `p`.

ternaus:
Works.

PR #1276, opened 2022-09-02
Issue #1263 Support for float p in InvertImg

dmitrie-ai:
Added support for `float p` in InvertImg in albumentations/augmentations/transforms.py. It defines the probability that InvertImg performs the inversion operation. By default `p = 0.5`.
The testing function `test_additional_targets_for_image_only()` had to be changed accordingly in order to pass. So it is tested by passing `p=0` and checking that the image is unchanged.
May be worth adding a specific testing function for InvertImg that tests for example that inverting twice returns the original image.

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -1549,8 +1549,15 @@ class InvertImg(ImageOnlyTransform):
         uint8
     """
 
+    def __init__(self, always_apply=False, p=0.1):
+        super(InvertImg, self).__init__(always_apply, p)
+        self.p: float = p
+
     def apply(self, img, **params):
-        return F.invert(img)
+        if random.random() < self.p:
+            return F.invert(img)
+        else:
+            return img
 
     def get_transform_init_args_names(self):
         return ()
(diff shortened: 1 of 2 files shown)


Dipet:
These changes are incorrect. They are only overwrites `BaseTransform` logic and do not affect support of float32.
Original issues says about support `np.float32` dtype images.

dmitrie-ai:
Apologies, it seems I have misunderstood the task.
I would like to try again if that's alright.
To confirm, the task is to add the ability to invert `float32` dtype images. What would be the range of values? 0-1?
Thank you

Dipet:
Now this transform only supports images with a `np.uint8` dtype. Need to add support for any dtype and chose a max value based on the image dtype.

dmitrie-ai:
So if the dtype is `np.float32` then the max value is 1.0 and if it's any `int` dtype then the max is 255? And it should detect the type automatically as oppose to having a parameter that the user can specify?

Dipet:
Yes, like `clipped` function https://github.com/albumentations-team/albumentations/blob/master/albumentations/augmentations/utils.py#L67

PR #1277, opened 2022-09-02
Issue #1263 Support for various dtypes in InvertImg.

dmitrie-ai:
- Added support for dtypes other than `np.uint8` in InvertImg. The changes were made in `invert()` in /augmentations/functional.py.
- Additional function `test_image_invert()` in `test_transforms.py` checking that inverting twice returns the original image.


albumentations/augmentations/functional.py
```diff
@@ -795,8 +795,10 @@ def add_shadow(img, vertices_list):
     return image_rgb
 
 
-def invert(img):
-    return 255 - img
+def invert(img: np.ndarray) -> np.ndarray:
+    # Supports all the valid dtypes
+    # clips the img to avoid unexpected behaviour.
+    return MAX_VALUES_BY_DTYPE[img.dtype] - clip(img, img.dtype, MAX_VALUES_BY_DTYPE[img.dtype])
```
Dipet, line 801: We do not need clip there

albumentations/augmentations/transforms.py
```diff
@@ -1549,6 +1549,9 @@ class InvertImg(ImageOnlyTransform):
         uint8
     """
 
+    def __init__(self, always_apply=False, p=0.5):
+        super(InvertImg, self).__init__(always_apply, p)
```
Dipet, line 1553: Remove this, this do nothing

tests/test_transforms.py
```diff
@@ -312,6 +312,15 @@ def test_additional_targets_for_image_only(augmentation_cls, params):
         assert np.array_equal(aug1, aug2)
 
 
+def test_image_invert():
+    # test for np.uint8 dtype
+    image1 = np.random.randint(low=0, high=256, size=(100, 100, 3), dtype=np.uint8)
+    image2 = image1.copy()
+    r1 = F.invert(image1)
+    r2 = F.invert(r1)
+    assert np.array_equal(image2, r2)
```
Dipet, line 321: Better to compare uint8 and float32 results

Dipet (changes requested):

tests/test_transforms.py
dmitrie-ai: Thank you for the feedback. Do you mean test with both uint8 random array and a random float32 array?

albumentations/augmentations/functional.py
dmitrie-ai: I thought maybe if the user enters an array that is not of valid floats, e.g. floats not in range. But I'm guessing there are some tests in place for that already and it's better to yield an exception. Thank you for the feedback

tests/test_transforms.py
Dipet:
```python
image2  = A.to_float(image1)
r1 = A.to_float(r1)
```

tests/test_transforms.py
```diff
@@ -312,6 +312,17 @@ def test_additional_targets_for_image_only(augmentation_cls, params):
         assert np.array_equal(aug1, aug2)
 
 
+def test_image_invert():
+    for _ in range(10):
+        # test for np.uint8 dtype
+        image1 = np.random.randint(low=0, high=256, size=(100, 100, 3), dtype=np.uint8)
+        image2 = A.to_float(image1)
+        r_int = F.invert(F.invert(image1))
+        r_float = F.invert(F.invert(image2))
```
Dipet, line 321: Why you call `invert` twice?

dmitrie-ai: I figured a good test of inversion is to invert twice and check if it's the same as the original image.

Dipet (approved):

PR #1278, opened 2022-09-06
Fixed the error when bbox dimensions gets less than 0 or greater than 1

POONAM2015:
image

### DIFF
--- a/albumentations/core/bbox_utils.py
+++ b/albumentations/core/bbox_utils.py
@@ -326,6 +326,8 @@ def convert_bbox_to_albumentations(
     else:
         (x_min, y_min, x_max, y_max), tail = bbox[:4], bbox[4:]
 
+    clamp = np.clip(x, 0.0, 1.0)
+    x_min, x_max, y_min, y_max = map(clamp, (x_min, x_max, y_min, y_max))
     bbox = (x_min, y_min, x_max, y_max) + tuple(tail)  # type: ignore
 
     if source_format != "yolo":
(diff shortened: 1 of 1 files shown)


Dipet:
We can not change bboxes silently, because these changes might be incorrect in some cases. Better to add flag `clamp_bbox=False` into this function and into [`BboxParams and BboxProcessor`](https://github.com/albumentations-team/albumentations/blob/master/albumentations/core/bbox_utils.py#L65)

POONAM2015:
I have added the parameter `clamp_bbox=False` to `BBoxParams` which will control whether bboxes will be forcefully clamped to remain between 0 and 1.
Thanks.

albumentations/core/bbox_utils.py
```diff
@@ -118,7 +119,7 @@ def check(self, data: Sequence, rows: int, cols: int) -> None:
         check_bboxes(data)
 
     def convert_from_albumentations(self, data: Sequence, rows: int, cols: int) -> List[BoxType]:
-        return convert_bboxes_from_albumentations(data, self.params.format, rows, cols, check_validity=True)
+        return convert_bboxes_from_albumentations(data, self.params.format, rows, cols, check_validity=True, clamp_bbox=self.params.clamp_bbox)
 
     def convert_to_albumentations(self, data: Sequence[BoxType], rows: int, cols: int) -> List[BoxType]:
         return convert_bboxes_to_albumentations(data, self.params.format, rows, cols, check_validity=True)
```
Dipet, line 125: You added this parameter to wrong function. You changed `to` function, but added param to `from` function. Please add small test to check that this functionality works coreectly

ternaus:
Looks like the problem could be resolved with `clip=True` passed to BboxParams

Issue #1225, opened 2022-07-13
Add CutAndPaste transform


ternaus:
1. Specify the object as PNG with the empty background.
2. Specify background image.
3. Cut out a region in the background image that is larger than the object.
4. Paste the object to the cut.
5. Use Poisson Blending to inpaint space between pasted object and image.

i-aki-y:
@ternaus  Hi, I'm interested in the feature.

How can we specify the pasted object in an albumentation's transform? Do we need to introduce a new target keyword like 'paste_image':

```py
transform(image=image, paste_image=paste_image, ...)
```

, or sample segments from the same target image as yolov5 is doing?

https://github.com/ultralytics/yolov5/blob/15e82d296720d4be344bf42a34d60ffd57b3eb28/utils/dataloaders.py#L706

Dipet:
I think we could support this:
- Add `past_image_key` param for transform - if this key is set we will use provided images
- Add `past_image_dir` parms for transform - is this key is set we will sample random image from this directory
- Otherwise sample random segment from the image.

i-aki-y:
@Dipet thank you for your reply.

Do You mean that the `paste_image_key` is used to set a path to the png file as a target?

```
transform(image=image, paste_image_key="path/to/png", ...)
```

If so, I think using `past_image_dir` is a better choice since the `past_image_dir` can be used as a parameter of the constructor. And we can avoid introducing new targets and follow the standard usage.

Ex.

```py 
transform Compose([
    CopyAndPaste(paste_image_dir=objects_dir, …),
...
])

transform(image=image, …)   # we can still use only the standard targets bboxes, and masks.
```

To get a feel for it, I made a workable example in PR #1297 (still working).

zetyquickly:
> ## Feature description
> Add CopyAndPaste Augmentation from https://arxiv.org/abs/2012.07177
> 
> Used in github.com/ultralytics/yolov5/blob/ac6c4383bc0c7a2a4f7ca18f8733821b49e916bd/utils/augmentations.py#L19

Checked the yolov5 code, [here](https://github.com/ultralytics/yolov5/blob/ac6c4383bc0c7a2a4f7ca18f8733821b49e916bd/utils/augmentations.py#L244). It looks they don't do the method the paper describes

Paper quote:
[...]

PR #1297, opened 2022-09-23
Add CutAndPaste

i-aki-y:
## About PR

This PR tries to implement Cut And Paste Augmentation "[Simple Copy-Paste is a Strong Data Augmentation Method for Instance Segmentation][1]".
It includes multiple image blending methods given in "[Poisson Image Editing][2]".

See also: #1225

[1]: https://openaccess.thecvf.com/content/CVPR2021/papers/Ghiasi_Simple_Copy-Paste_Is_a_Strong_Data_Augmentation_Method_for_Instance_CVPR_2021_paper.pdf
[2]: http://www.irisa.fr/vista/Papers/2003_siggraph_perez.pdf
[3]: http://amroamroamro.github.io/mexopencv/opencv/cloning_demo.html

## Blending Demo

The `A.paste` function supports the following four image blending method.
* **GAUSSIAN**: Used in "[Simple Copy-Paste is a Strong Data Augmentation Method for Instance Segmentation][1]".  
* **NORMAL_CLONE, MIXED_CLONE, MONOCHROME_TRANSFER**: Different types of seamless cloning are explained in "[Poisson Image Editing][2]".

BaseAndPasteImages
BlendComparison

Figure: The result of different methods: GAUSSIAN, NORMAL_CLONE, MIXED_CLONE, MONOCHROME_TRANSFER (Left to right)

## CutAndPaste Transform Demo

This is an example of the CutAndPaste Transform usage.

```py
transform = A.Compose(
    [
        A.CutAndPaste(
            p=1.0, 
            paste_image_dir=object_dir,
            get_label_from_path=get_label_from_path,
            num_object_limit=(1, 3),
            blend_method="GAUSSIAN"
        )
    ],
    bbox_params=A.BboxParams(format='coco', label_fields=["labels"])
)
```

It requires paste_image_dir parameters, which include objects to be pasted. The transform randomly selects multiple files (objects) from the paste_image_dir directory. So in this version, the user needs to prepare images so that each image includes a single object. 

The transform also requires a function parameter `get_label_from_path`, which returns label information from a given object file path. Since the transform could not know the label information of randomly selected objects, the `get_label_from_path` is used to extract label information from the object file path. This means that the user should include label information in the object path. And provide the function to extract the information from the path.

The followings are the results of different trials with a fixed base image.
RandomExamples_no_annots

RandomExamples_with_annots

## Sample Notebook

You can reproduce the same result in this notebook on Colab:
https://colab.research.google.com/drive/1sFCAhS8FTyp7dLIdJgUJwbB5JnGBYoq3

### 1. The user needs to prepare object files as RGBA PNG images.
As described above, the user needs to prepare object files (images to be pasted) in advance.

###  2. The object file path should include label information
Since the transform identifies the label information from the object path, the user needs to include label information in the object path and provide the function `get_label_from_path`as a parameter, which extracts label information from the path.
For example, when the object path is like a `path/to/object/{object_id}_{label_id}.png`, a `get_label_from_path` could be:

```py
get_label_from_path = lambda image_path: int(image_path.stem.split("_")[-1])
```

### 3. Returned masks become binary masks
Even if the input masks are non-binary masks, the transformed masks are binary masks. (The values are 0 or 1)
I think I can remove this limitation with extra work. 

### 1. A new rotation method, `rotate_bound` is added.

A new rotation function, `rotate_bound` is introduced to rotate object images.
While the standard `rotate` function causes unwanted crops, the `rotate_bound` expands the input's shape depending on the rotation. See the following examples.

CompareRotation

### 2. Augmentation to the masks are actually done inside the `get_params_dependent_on_targets`

Augmented masks are needed to calculate the bbox augmentation. But I could not find a better place where both the bboxes and masks are accessible except for `get_params_dependent_on_targets`.
So I implement the masks augmentation in the `get_params_dependent_on_targets`

----

- 2022-10-18:  Update description and examples.


i-aki-y:
Most of the work has been completed.
@Dipet @ternaus I would appreciate it if you reviewed this.

i-aki-y:
Sorry for the additional correction.
The current implementation assumes that the number of all objects in an image is less than 255("the length of target masks" + "the number of pasted objects" < 255).
So, I added some validation code and documentation about the limitation.
I think the number 255 is sufficiently large for a typical usecase, but if needed, it can be increased.

PR #1305, opened 2022-10-07
Color parameter for Spatter

### LINKED ISSUE #1304: Цвет тумана у Spatter()-а
Хорошо бы иметь возможность задавать цвет "тумана" https://github.com/albumentations-team/albumentations/blob/master/albumentations/augmentations/transforms.py#L2482 .

Andredance:
Added color parameter for Spatter augmentation. Related to this issue (resolves #1304 )

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -2391,8 +2391,10 @@ class Spatter(ImageOnlyTransform):
             If tuple of float intensity will be sampled from range `[intensity[0], intensity[1])`. Default: (0.6).
         mode (string, or list of strings): Type of corruption. Currently, supported options are 'rain' and 'mud'.
              If list is provided type of corruption will be sampled list. Default: ("rain").
-        color (list of (r, g, b) or None): Corruption elements color. If None uses default colors (rain: (238, 238, 175)
-             mud: (20, 42, 63)).
+        color (list of (r, g, b) or dict or None): Corruption elements color.
+            If list uses provided list as color for specified mode.
+            If dict uses provided color for specified mode. Color for each specified mode should be provided in dict.
+            If None uses default colors (rain: (238, 238, 175), mud: (20, 42, 63)).
         p (float): probability of applying the transform. Default: 0.5.
 
     Targets:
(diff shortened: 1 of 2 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -2422,10 +2425,13 @@ def __init__(
         self.gauss_sigma = to_tuple(gauss_sigma, gauss_sigma)
         self.intensity = to_tuple(intensity, intensity)
         self.cutout_threshold = to_tuple(cutout_threshold, cutout_threshold)
+        self.color = color
         self.mode = mode if isinstance(mode, (list, tuple)) else [mode]
         for i in self.mode:
             if i not in ["rain", "mud"]:
                 raise ValueError(f"Unsupported color mode: {mode}. Transform supports only `rain` and `mud` mods.")
+        if self.color is not None and len(self.color) < 3:
```
Dipet, line 2433: `len(self.color) != 3`

albumentations/augmentations/transforms.py
```diff
@@ -2412,6 +2414,7 @@ def __init__(
         cutout_threshold: ScaleFloatType = 0.68,
         intensity: ScaleFloatType = 0.6,
         mode: Union[str, Sequence[str]] = "rain",
+        color: Optional[Sequence[int]] = None,
```
Dipet, line 2417: I think if the `mode` is `Sequence`  we need to use `color: Dict[str, Sequence[int]]` where key is mode name. It is needed because we can not use the same color for rain and mud.

Dipet (changes requested):

albumentations/augmentations/transforms.py
Andredance: Makes sense!

albumentations/augmentations/transforms.py
Andredance: Resolved in a new version.

albumentations/augmentations/transforms.py
Andredance: Added support for Dict.

albumentations/augmentations/transforms.py
```diff
@@ -2422,10 +2427,34 @@ def __init__(
...
+            color
+            if color is not None
+            else {
+                "rain": [238, 238, 175],
+                "mud": [20, 42, 63],
+            }
+        )
         self.mode = mode if isinstance(mode, (list, tuple)) else [mode]
+
+        if len(set(self.mode)) > 1 and not isinstance(self.color, dict):
```
Dipet, line 2440: `and len(self.color) != 2` - check that color provided for both modes

Dipet (changes requested):

albumentations/augmentations/transforms.py
Andredance: I think we don't need it since it is covered by lines 2446-2452. What do you think?

Dipet (approved):

Issue #915, opened 2021-05-28
ValueError: Requested crop size is larger than the image size in RandomResizedCrop


claverru:
## 🐛 Bug

`ValueError: Requested crop size is larger than the image size` when including RandomResizedCrop in my transforms pipeline:

```
ValueError: Caught ValueError in DataLoader worker process 1.
Original Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/worker.py", line 202, in _worker_loop
    data = fetcher.fetch(index)
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 44, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/opt/conda/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 44, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "/workspace/ai_damages/parts/src/data_loading.py", line 64, in __getitem__
    transformed = self.transforms(image=img, mask=mask)
  File "/opt/conda/lib/python3.8/site-packages/albumentations/core/composition.py", line 182, in __call__
    data = t(force_apply=force_apply, **data)
  File "/opt/conda/lib/python3.8/site-packages/albumentations/core/transforms_interface.py", line 89, in __call__
    return self.apply_with_params(params, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/albumentations/core/transforms_interface.py", line 102, in apply_with_params
    res[key] = target_function(arg, **dict(params, **target_dependencies))
  File "/opt/conda/lib/python3.8/site-packages/albumentations/core/transforms_interface.py", line 227, in apply_to_mask
    return self.apply(img, **{k: cv2.INTER_NEAREST if k == "interpolation" else v for k, v in params.items()})
  File "/opt/conda/lib/python3.8/site-packages/albumentations/augmentations/transforms.py", line 857, in apply
    crop = F.random_crop(img, crop_height, crop_width, h_start, w_start)
  File "/opt/conda/lib/python3.8/site-packages/albumentations/augmentations/functional.py", line 331, in random_crop
    raise ValueError(
ValueError: Requested crop size (2131, 1636) is larger than the image size (1836, 3264)
```

## To Reproduce

I can't reproduce it deterministically, it just happens sometimes. 

## Environment

[...]

PR #1310, opened 2022-10-11
Check for image and masks shape equality

### LINKED ISSUE #1299: Add check that masks and image have the same height and width
This needed to avoid problems inside spatial-level transforms when params generated based on image shapes, for example pad and crop augs.

Andredance:
Added check inside _check_args function for equality of shapes for image, mask, and masks. (Resolves #1299 )

### DIFF
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -172,6 +172,8 @@ class Compose(BaseCompose):
         self.is_check_args = True
         self._disable_check_args_for_transforms(self.transforms)
 
+        self.is_check_shapes = True
+
     @staticmethod
     def _disable_check_args_for_transforms(transforms: TransformsSeqType) -> None:
         for transform in transforms:
@@ -182,6 +184,9 @@ class Compose(BaseCompose):
 
     def _disable_check_args(self) -> None:
         self.is_check_args = False
+
+    def disable_shapes_check(self) -> None:
+        self.is_check_shapes = False
 
     def __call__(self, *args, force_apply: bool = False, **data) -> typing.Dict[str, typing.Any]:
         if args:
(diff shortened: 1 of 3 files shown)


Andredance:
The reason to check only the first mask from masks is the same as it is for types checking.

albumentations/core/composition.py
```diff
@@ -260,18 +260,24 @@ def _check_args(self, **kwargs) -> None:
...
             if internal_data_name in checked_multi:
                 if data:
                     if not isinstance(data[0], np.ndarray):
                         raise TypeError("{} must be list of numpy arrays".format(data_name))
+                    shapes.append(data[0].shape[:2])
             if internal_data_name in check_bbox_param and self.processors.get("bboxes") is None:
                 raise ValueError("bbox_params must be specified for bbox transformations")
 
+        if shapes and shapes.count(shapes[0]) != len(shapes):
+            raise ValueError("Height and Width of image, mask or masks should be equal.")
```
Dipet, line 279: I think we need to add the param `check_shapes` to`Compose`. It would also be great to add note on how to disable this check if the user thinks that ok.

Dipet (changes requested):

albumentations/core/composition.py
Andredance: Added __disable_shapes_check__ method for Compose class. And mentioned this method inside the Error message.

albumentations/core/composition.py
```diff
@@ -172,6 +172,8 @@ def __init__(
         self.is_check_args = True
         self._disable_check_args_for_transforms(self.transforms)
 
+        self.is_check_shapes = True
```
Dipet, line 175: Please add this argument to `_to_dict` and to `__init__` args this will be useful for users.

albumentations/core/composition.py
```diff
@@ -183,6 +185,9 @@ def _disable_check_args_for_transforms(transforms: TransformsSeqType) -> None:
     def _disable_check_args(self) -> None:
         self.is_check_args = False
 
+    def disable_shapes_check(self) -> None:
+        self.is_check_shapes = False
```
Dipet, line 189: I think we must control this behavior using `is_check_shapes` as common argument inside `__init__`

albumentations/core/composition.py
Dipet: Would be great if you add this parameter to serialization test. But it is not necessary.

Dipet (changes requested):

Dipet:
This PR also fixes #915

albumentations/core/composition.py
Andredance: I think we don't need a separate test for this parameter since tests/test_serialization.py __test_serialization_v2_to_dict__ will cover it. What do you think?

Dipet (approved):

PR #1311, opened 2022-10-12
Fix deprecated imports of scipy.ndimage.gaussian_filter

rbu:
scipy deprecated the use of the `ndimage.filters` submodule in v1.8.0 and plans to remove the compatibility import in v2.0.0 The filters were availalble in the `ndimage` module going back at least to 2012, and I tested the import on scipy==1.1.0.

Fixes the following error:
```
$ python -W always -c "import albumentations.augmentations.transforms" albumentations/albumentations/augmentations/geometric/functional.py:7: DeprecationWarning: Please use `gaussian_filter` from the `scipy.ndimage` namespace, the `scipy.ndimage.filters` namespace is deprecated.
  from scipy.ndimage.filters import gaussian_filter
```

### DIFF
--- a/albumentations/augmentations/geometric/functional.py
+++ b/albumentations/augmentations/geometric/functional.py
@@ -4,7 +4,7 @@ from typing import List, Optional, Sequence, Tuple, Union
 import cv2
 import numpy as np
 import skimage.transform
-from scipy.ndimage.filters import gaussian_filter
+from scipy.ndimage import gaussian_filter
 
 from albumentations.augmentations.utils import (
     _maybe_process_in_chunks,
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -11,7 +11,7 @@ from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 import cv2
 import numpy as np
 from scipy import special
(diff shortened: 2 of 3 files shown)


Dipet:
I think that we also need to update the version requirements for libs for these changes https://github.com/albumentations-team/albumentations/blob/master/setup.py#L8

rbu:
> I think that we also need to update the version requirements for libs for these changes https://github.com/albumentations-team/albumentations/blob/master/setup.py#L8

Which dependency do you have in mind? Specifically for scipy, I tested that the "new" import location (i.e. the one that is not deprecated, and which this PR starts using) has been around in scipy 1.1.0 from 2012. I didn't run the full test suite to find out what the minimal version requirement actually are, but I doubt that this change would change that in practice.
That being said, I'm happy to run more tests or adapt version specs in this PR, if you let me know in what direction.

rbu:
> In that case, specify `scipy>=1.1.0` in `setup.py`. At the moment, no version is specified for `scipy` at all.

I've adapted the branch accordingly and rebased to master.

Dipet (approved):

rbu:
Thank you for the timely review and merge @jangop  @Dipet

PR #1314, opened 2022-10-14
Fix misleading docstrings in transforms.py

oguz-hanoglu:
Jpeg Compression uses cv2.imencode and cv2.imdecode to get a compressed version of the image. It compresses the image and decreases the image quality.

However, the docstrings of the related methods are misleading: "Decrease Jpeg compression of an image."

This implies that an active decrease in quality is recovered by this method.

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -222,7 +222,7 @@ class Normalize(ImageOnlyTransform):
 
 
 class ImageCompression(ImageOnlyTransform):
-    """Decrease Jpeg, WebP compression of an image.
+    """Decreases image quality by Jpeg, WebP compression of an image.
 
     Args:
         quality_lower (float): lower bound on the image quality.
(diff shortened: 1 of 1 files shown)


albumentations/augmentations/transforms.py
```diff
@@ -292,7 +292,7 @@ def get_transform_init_args(self):
 
 
 class JpegCompression(ImageCompression):
-    """Decrease Jpeg compression of an image.
+    """Decreases image quality by Jpeg, WebP compression of an image.
```
Dipet, line 295: This is only Jpeg aug

Dipet (changes requested):

albumentations/augmentations/transforms.py
oguz-hanoglu: Corrected with another commit.

Dipet (approved):

Issue #361, opened 2019-09-19
RandomFog and other augmentations can get stuck in infinite while loops


nsfabina:
## 🐛 Bug

RandomFog (and other augmentations with while loops) may get stuck in infinite while loops under certain conditions. At least one of those conditions is when an image has been augmented prior to RandomFog being applied.

## To Reproduce

I was able to get this bug with an image using the following compose operation:
```
    return albumentations.Compose([
        albumentations.RGBShift(r_shift_limit=0.2, g_shift_limit=0.2, b_shift_limit=0.2, p=1),
        albumentations.RandomFog(fog_coef_lower=0, fog_coef_upper=1, alpha_coef=0.08, p=1),
    ])
```
After ~2 mins of waiting, I interrupted and the code was still here
```
~/miniconda3/envs/asu/lib/python3.7/site-packages/albumentations/augmentations/transforms.py in get_params_dependent_on_targets(self, params)
   1609
   1610         while midx > -hw or midy > - hw:
-> 1611             for i in range(hw // 10 * index):
   1612                 x = random.randint(midx, width - midx - hw)
   1613                 y = random.randint(midy, height - midy - hw)
```

## Expected behavior

A few thoughts after testing the library:
- many augmentations appear to be order-of-operations dependent, might be worth putting a note in the documentation about how results can vary drastically depending on this order
- specifically for RandomFog and other augmentations where results depend on the image properties, perhaps a warning that they should be applied first, or perhaps a warning that they can have suboptimal interactions with other augmentations
- augmentations with while loops may need some sort of time- or iteration-based limit to avoid hangs or excessive pre-processing
- (that's all I've got, sorry, it's still early here!)

## Environment

 - Albumentations version (e.g., 0.1.8):  albumentations==0.3.3
 - Python version (e.g., 3.7):  3.7
 - OS (e.g., Linux):  OSX
 - How you installed albumentations (`conda`, `pip`, source): conda
 - Any other relevant information: nope

nsfabina:
Apologies. Further testing shows that I'm able to do this with RandomFog only, that there's no interactive effect here.

Reproduce:
```
    return albumentations.Compose([
        albumentations.RandomFog(fog_coef_lower=0, fog_coef_upper=1, alpha_coef=0.08, p=0.5),
    ])
```

ternaus:
Fixed in https://github.com/albu/albumentations/pull/362

opeide:
[...]

PR #1321, opened 2022-10-21
fix RandomFog infinite loop

cgkim412:
### Where's the bug?
```
while midx > -hw or midy > -hw:
    for _i in range(hw // 10 * index):
        x = random.randint(midx, width - midx - hw)
        y = random.randint(midy, height - midy - hw)
        haze_list.append((x, y))

    midx -= 3 * hw * width // sum(imshape)
    midy -= 3 * hw * height // sum(imshape)
    index += 1
```

This while loop is supposed to end when _midy_ drops below a certain value. But in some cases _height // sum(imshape)_ can be 0, which means _midy_ won't get any smaller from there, resulting in an infinite loop.
Note that _hw_ is always a positive integer thus the change in _midy_ becomes 0 if and only if _height // sum(imshape)_ equals 0.

The same applies to _midx_, so both lines should be changed.

### How to reproduce?
Run the code below and it will almost surely fall into an infinite loop.
```
from PIL import Image
import albumentations as A
import numpy as np

img = Image.new("RGB", (32, 95), "white")
img = np.array(img)
for _ in range(20):
    A.RandomFog(0.13814, 0.13814, 0.6, p=1.0)(image=img)
```

### Additional comments
As far as I can see, a similar issue was reported and a fix was merged about 3 years ago. (#361 )
However the infinite loop still occurs under certain conditions and requires an additional change; therefore this PR.

### DIFF
--- a/albumentations/augmentations/transforms.py
+++ b/albumentations/augmentations/transforms.py
@@ -566,8 +566,8 @@ class RandomFog(ImageOnlyTransform):
                 y = random.randint(midy, height - midy - hw)
                 haze_list.append((x, y))
 
-            midx -= 3 * hw * width // sum(imshape)
-            midy -= 3 * hw * height // sum(imshape)
+            midx -= 3 * hw * max(1, width // sum(imshape))
+            midy -= 3 * hw * max(1, height // sum(imshape))
             index += 1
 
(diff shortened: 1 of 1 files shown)


cgkim412:
Sorry, I made a stupid mistake. Closing it for now.

Issue #1032, opened 2021-10-08
[FR] Grayscale to RGB


hoangtnm:
Hi, I'm using `albumentations` for my project, but my image is GrayScale with shape `[H, W]` but I need to use RGB image (3 channels or `[H, W, 3]`) for training neural networks. Does the library provides any transformation to `convert grayscale to RGB`, which is similar to ` tf.image.grayscale_to_rgb`? thank you

BloodAxe:
A combination of `[A.Normalize(), A.ToTensorV2()] shoud do the job. Please refer to this [notebook](https://github.com/albumentations-team/albumentations_examples/blob/24f3bf740ffda010dd5ceda743d4cb41899b7d22/notebooks/migrating_from_torchvision_to_albumentations.ipynb) for an example.

hoangtnm:
Hi @BloodAxe, as I understand [RGBShift](https://albumentations.ai/docs/api_reference/augmentations/transforms/#albumentations.augmentations.transforms.RGBShift) is used for RGB image with shape [H, W, 3], does `albumentations` has the equivalent transform for Grayscale image with shape [H, W]?

amin-nejad:
+1 for a `ToRGB` transformation in the same vein as `ToGray`

Beluuuuuuga:
Is this Issue still ongoing? Can I take it?

domef:
A tranformation `A.ToRGB` would be very usefull.

Dipet:
ToRGB added in #1323

PR #1323, opened 2022-10-21
Add ToRGB

kinoooshnik:
Added ToRGB transform, as suggested in Issue https://github.com/albumentations-team/albumentations/issues/1032.


albumentations/augmentations/transforms.py
```diff
@@ -1422,6 +1423,35 @@ def get_transform_init_args_names(self):
...
+
+    Targets:
+        image
+
+    Image types:
+        uint8, float32
+    """
+
+    def __init__(self, always_apply=True, p=1.0):
+        super(ToRGB, self).__init__(always_apply=always_apply, p=p)
```
Dipet, line 1440: Do not need to copy `__init__` if it just calls super method without changes. Please, remove it.

Dipet (approved):

albumentations/augmentations/transforms.py
kinoooshnik: I changed the default argument values from `always_apply=False, p=0.5` to `always_apply=True, p=1.0`, so you can write `ToRGB()` instead of `ToRGB(p=1)`. Do I still need to remove this?

Dipet: Oh, you are right. Sorry.

PR #1325, opened 2022-10-23
Support keypoint on ``ElasticTransform`` (2-D)

minipuding:
Rewrite (Add)  ``apply_to_keypoint`` funtion on class ``ElasticTransform``. Only consider 2-D case.


albumentations/augmentations/geometric/transforms.py
```diff
@@ -251,6 +252,28 @@ def apply_to_bbox(self, bbox, random_state=None, **params):
...
+            self.sigma,
+            self.alpha_affine,
+            self.interpolation,
+            self.border_mode,
+            self.mask_value,
+            np.random.RandomState(random_state),
+            self.approximate,
+        )
+        interp_y, interp_x = np.where(remap_image == np.max(remap_image))
+        return (interp_x[0], interp_y[0], 0.0, 0.0)
```
Dipet, line 276: What if the keypoint outside of the image? Looks like we always will find keypoint inside the image and in the case  when we can not see the keypoint

minipuding: Yes, there is such a problem, but I can't think of a simple and elegant way to solve it. Therefore, you should try to ensure that the keypoints do not go beyond the scope of the image, otherwise it will return incorrect results.

PR #1356, opened 2022-11-30
[bugfix] random.choice error with python3.11 (blur/transforms.py, ver…

whlook:
…sion on 1.3.0)

  err info:ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
  fix with array.tolist()

### DIFF
--- a/albumentations/augmentations/blur/transforms.py
+++ b/albumentations/augmentations/blur/transforms.py
@@ -41,7 +41,7 @@ class Blur(ImageOnlyTransform):
         return F.blur(img, ksize)
 
     def get_params(self) -> Dict[str, Any]:
-        return {"ksize": int(random.choice(np.arange(self.blur_limit[0], self.blur_limit[1] + 1, 2)))}
+        return {"ksize": int(random.choice(np.arange(self.blur_limit[0], self.blur_limit[1] + 1, 2).tolist()))}
 
     def get_transform_init_args_names(self) -> Tuple[str, ...]:
         return ("blur_limit",)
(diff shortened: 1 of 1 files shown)


Dipet:
Could you provide a reproducible example of the error?

Dipet:
Related #1357

Dipet:
Please change it inside `Blur` too.

albumentations/augmentations/blur/transforms.py
```diff
@@ -41,7 +41,7 @@ def apply(self, img: np.ndarray, ksize: int = 3, **params) -> np.ndarray:
         return F.blur(img, ksize)
 
     def get_params(self) -> Dict[str, Any]:
-        return {"ksize": int(random.choice(np.arange(self.blur_limit[0], self.blur_limit[1] + 1, 2)))}
+        return {"ksize": int(random.choice(np.arange(self.blur_limit[0], self.blur_limit[1] + 1, 2).tolist()))}
```
Dipet, line 44: use `range` instead of `np.arange`

whlook:
> It looks like we need to replace all `np.arange` with a `range`

Have tried itSomething wrong with 'random.choice(np.arange...'，

albumentations/augmentations/blur/transforms.py
whlook: Have changed to 'range'. It seems a problem with the combination of 'random.choice' and np.arange,so i just fix 'arange' on these cases.

Issue #1374, opened 2023-01-03
Support different rgb to grayscale methods.


Dipet:
Support different [methods](https://docs.gimp.org/2.6/en/gimp-tool-desaturate.html) for grayscale conversions:
- `Luminosity` with different coefficients.
- `Lightness`
- `Average`
Some info you can find on wiki: https://en.wikipedia.org/wiki/Grayscale

pneycho:
Hi @Dipet, It'd be great if you could take a look at the commit. There is an issue I fixed, about the gray image inversion if the mean pixel value of the resultant value is over 127 (which is mentioned was mentioned in the docstring but not implemented). Is there any specific reason why that is being done? The only thing I could think of is visibility enhancement of specific structures. Am I correct in assuming that?

Dipet:
@pneycho 
About what commit are you talking? I do not see any pull request there. Try to create it please

Dipet:
Ok, I see. It looks like we talking about #1387

ternaus:
@pneycho Desired functionality merged.

PR #1387, opened 2023-01-26
Fixed Issue #1374: Support for different rgb to grayscale methods

pneycho:
1. Added support for 7 different methods .
                       'average': Average value of R,G and B channels.
                                 Formula is, gray = (R + G + B)/3
                      'luminosity': Based on the human perception of brightness and assigns more weight to
                                    the G channel than the R or B channels.
                                    Formula is, gray = 0.21 * R + 0.72 * G + 0.07 * B
                      'lightness': Uses the maximum and minimum values of the R,G and B channels.
                                   Formula is, gray = (max(R, G, B) + min(R, G, B)) / 2
                      'weighted_average' [DEFAULT since it is used by OpenCV]: Assigns different weights to the red, green and blue channels. Formula is, gray =  0.2989 * R + 0.5870 * G + 0.1140 * B
                      'extract_luminance_hsv': Converts the image to the HSV color space and
                                               then extracts the Value channel.
                      'extract_luminance_lab': Used to preserve color information by converting the
                                               image to LAB color space and then extracting the Luminance channel.
                      'single_channel': Selects one of the R,G, or B channels as the grayscale image.
3. Fixed the image inversion issue (invert colors if mean of resultant image > 127). It was present in the documentation but not implemented.
4. Added support for use of randomized method selection, if user does not want to make a specific choice.


albumentations/augmentations/functional.py
```diff
@@ -919,11 +919,46 @@ def iso_noise(image, color_shift=0.05, intensity=0.5, random_state=None, **kwarg
...
+    elif method == "extract_luminance_hsv":
+        gray = _rgb_to_gray_via_hsv(img)
+    elif method == "extract_luminance_lab":
+        gray = _rgb_to_gray_via_lab(img)
+    elif method == "single_channel":
+        gray = np.random.choice([r, g, b])
+
+    # if the mean pixel value is over 127, the image is inverted.
+    if np.mean(gray) > 127:
+        gray = 255 - gray
```
Dipet, line 942: I think this is a strange approach .I don't think we need to do this. Just change the original docstring and remove the mention of invert.

albumentations/augmentations/functional.py
```diff
@@ -919,11 +919,46 @@ def iso_noise(image, color_shift=0.05, intensity=0.5, random_state=None, **kwarg
     return image.astype(np.uint8)
 
 
-def to_gray(img):
-    gray = cv2.cvtColor(img, cv2.COLOR_RGB2GRAY)
+def to_gray(img, method="weighted"):
+    r, g, b = img[:, :, 0], img[:, :, 1], img[:, :, 2]
+
+    if method == "average":
+        gray = (r + g + b) // 3
```
Dipet, line 926: I think better to use:
```python
gray = np.average(img, -1)
```

albumentations/augmentations/functional.py
```diff
@@ -919,11 +919,46 @@ def iso_noise(image, color_shift=0.05, intensity=0.5, random_state=None, **kwarg
...
+    elif method == "lightness":
+        gray = (np.maximum(r, np.maximum(g, b)) + np.minimum(r, np.minimum(g, b))) / 2.0
+    elif method == "weighted_average":
+        gray = cv2.cvtColor(img, cv2.COLOR_RGB2GRAY)
+    elif method == "extract_luminance_hsv":
+        gray = _rgb_to_gray_via_hsv(img)
+    elif method == "extract_luminance_lab":
+        gray = _rgb_to_gray_via_lab(img)
+    elif method == "single_channel":
+        gray = np.random.choice([r, g, b])
```
Dipet, line 938: We must to move this behaviour to indepenedt function and wrap it with `@clipped` to preserve dtype.

albumentations/augmentations/functional.py
```diff
@@ -919,11 +919,46 @@ def iso_noise(image, color_shift=0.05, intensity=0.5, random_state=None, **kwarg
...
-    gray = cv2.cvtColor(img, cv2.COLOR_RGB2GRAY)
+def to_gray(img, method="weighted"):
+    r, g, b = img[:, :, 0], img[:, :, 1], img[:, :, 2]
+
+    if method == "average":
+        gray = (r + g + b) // 3
+    elif method == "luminosity":
+        gray = 0.21 * r + 0.72 * g + 0.07 * b
+    elif method == "lightness":
+        gray = (np.maximum(r, np.maximum(g, b)) + np.minimum(r, np.minimum(g, b))) / 2.0
```
Dipet, line 930:
```python
gray = (np.max(img, -1) - np.min(img, -1)) * 0.5
```

albumentations/augmentations/functional.py
```diff
@@ -919,11 +919,46 @@ def iso_noise(image, color_shift=0.05, intensity=0.5, random_state=None, **kwarg
     return image.astype(np.uint8)
 
 
-def to_gray(img):
-    gray = cv2.cvtColor(img, cv2.COLOR_RGB2GRAY)
+def to_gray(img, method="weighted"):
```
Dipet, line 922: `weighted` is not exists

albumentations/augmentations/functional.py
```diff
@@ -919,11 +919,46 @@ def iso_noise(image, color_shift=0.05, intensity=0.5, random_state=None, **kwarg
...
+    r, g, b = img[:, :, 0], img[:, :, 1], img[:, :, 2]
+
+    if method == "average":
+        gray = (r + g + b) // 3
+    elif method == "luminosity":
+        gray = 0.21 * r + 0.72 * g + 0.07 * b
+    elif method == "lightness":
+        gray = (np.maximum(r, np.maximum(g, b)) + np.minimum(r, np.minimum(g, b))) / 2.0
+    elif method == "weighted_average":
+        gray = cv2.cvtColor(img, cv2.COLOR_RGB2GRAY)
```
Dipet, line 932: Better to call it `opencv` or to have 2 names for this case

albumentations/augmentations/functional.py
Dipet, line 938:
```python
gray = img[..., random_utils.randint(0, 3)]
```

Dipet (changes requested):

PR #1389, opened 2023-01-31
Unproportional mosaicing

DavidAdamczyk:
We would like to contribute to project Albumentations with new augmentation named Unproportional mosaicing.

image

image

image


albumentations/augmentations/transforms.py
```diff
@@ -2664,3 +2667,151 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+                current_num_semgs = current_num_semgs + 1
+
+        tiles = []
+        for i in segms:
+            tiles.append([i.x, i.y, i.width, i.height])
+
+        tiles = np.asarray(tiles)
+
+        shape_y, shape_x = tiles.shape
+        shuffled_ids = np.linspace(0, shape_y - 1, shape_y, dtype=np.int32).tolist()
```
Dipet, line 2813: Do not save them into class member. Return as an argument.

albumentations/augmentations/transforms.py
```diff
@@ -2664,3 +2667,151 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+    def apply(self, img, tiles=None, **params):
+        if tiles is None:
+            tiles = []
+        self.ids = self.shuffled_ids.copy()
+        return F.unprop_swap_tiles_on_image(img, tiles, self.shuffled_ids)
+
+    def apply_to_mask(self, img, tiles=None, **params):
+        if tiles is None:
+            tiles = []
+        return F.unprop_swap_tiles_on_image(img, tiles, self.shuffled_ids)
```
Dipet, line 2697: If it's the same logic as `apply`, remove it - by default for masks `apply` will be called.

albumentations/augmentations/transforms.py
```diff
@@ -2664,3 +2667,151 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+
+    def apply_to_bboxes(self, bboxes, **params):
+        tiles = params["tiles"]
+        rows = params["rows"]
+        cols = params["cols"]
+        return F.unprop_bbox_transpose(bboxes, tiles, self.ids, rows, cols)
+
+    def get_transform_init_args_names(self):
+        return ()
```
Dipet, line 2706: Return `["grid"]`

albumentations/augmentations/transforms.py
```diff
@@ -2664,3 +2667,151 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+
+    def apply_to_keypoint(self, keypoint, **params):
+        return ()
```
Dipet, line 2709: Keypoints wil not be changed. We must raise error if method is not implemented (default behaviour) or implement it. Remove this function or implement it.

albumentations/augmentations/transforms.py
```diff
@@ -2664,3 +2667,151 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+
+    def get_params_dependent_on_targets(self, params):
```
Dipet, line 2711: Need to implement `targets_as_params` to use this method.

albumentations/augmentations/transforms.py
```diff
@@ -2664,3 +2667,151 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+        num_segments = params["num_segments"] if "num_segments" in params else 7
+        ratio = params["ratio"] if "ratio" in params else 4.0 / 3.0
+        # refinement_steps = params["refinement_steps"] if "refinement_steps" in params else 10
```
Dipet, line 2714: Mabe better to use it as an augmentation argument on creation. Also we could rewrite it if it will be provided on transform call.

albumentations/augmentations/transforms.py
```diff
@@ -2664,3 +2667,151 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+
+        def refinement():
+            """
+            Refinement the rectangles to look more like the demanded ratio considering tolerance and prevent stretches.
+            """
+            at_least_one_invalid = False
+            index = 0
+
+            # We must work with the copy because "Split" alters the original list
+            segms_copy = segms.copy()
```
Dipet, line 2742: We do not copy there

albumentations/augmentations/transforms.py
```diff
@@ -2664,3 +2667,151 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+
+            for segm in segms_copy:
```
Dipet, line 2744: Use `enum` for index

albumentations/augmentations/transforms.py
```diff
@@ -2664,3 +2667,151 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+
+                segms[index].width = new_width
+
+            # Add new rect to the list
+            segms.append(new_segm)
+
+            return True
+
+        # Split
+        current_num_semgs = 1
```
Dipet, line 2800: Typo `semgs`

Dipet (changes requested):

albumentations/augmentations/transforms.py
DavidAdamczyk: 121c461d4ecac16166fa34c7a02e6a1d19793631

albumentations/augmentations/transforms.py
DavidAdamczyk: d965c6d0933546e47478d600fd9193b44fbcf648

albumentations/augmentations/transforms.py
DavidAdamczyk: dbb4581811217408772c85a282446e4786b25d33

albumentations/augmentations/transforms.py
DavidAdamczyk: 2ed42382b0e844b47471f4bf509166626f8b4761

albumentations/augmentations/transforms.py
DavidAdamczyk: @Dipet I would like to ask you what do you mean by `enum`. You mean enumerate function?

Dipet: Replace
```python
index = 0

# We must work with the copy because "Split" alters the original list
segms_copy = segms.copy()

for segm in segms_copy:
```

with:
```python
# We must work with the copy because "Split" alters the original list
segms_copy = segms.copy()

for index, segm in enumerate(segms_copy):
```

thiagoribeirodamotta:
Hi there,
Are there any plans on merging this PR?

PR #1439, opened 2023-04-21
Update functional.py

rohitdavas:
There is big issue if the compose fails during training. the skimage warp method tries to access the inverse_map which could be none also. which is an issue directly, leading to failure of the compose call. 

wrapping it inside exception catch to reduce the failure.

Screenshot 2023-04-21 102310

Full exception :


Dipet:
Check your data or give reproducible example. It looks like a problem with your data

haddagart:
Apparently the problem is not related to Albumentation clearly the source of the issue comes from a datapoint that has to be of None type.

PR #1444, opened 2023-05-11
Web-App for Trying out Image Augmentations 

Varun0801:
Feature Addition:
- Created a web-app using Streamlit to try out Albumentations Image (Pixel-level transformation and Spatial-level transformation) Augmentations to any random image (can be uploaded and tested by user) or from sample images.
- Updated Readme to include the web-app's link in-place of broken Heroku web-app link.
- Try at: https://try-albumentations-by-varun.streamlit.app/


Dipet:
Thank you for your work, but we already have a repo with our demo: https://github.com/albumentations-team/albumentations-demo
If you want to add some features, please create a PR in that repository.
It looks like your app has some bugs.

PR #1447, opened 2023-05-14
Call `random.random` at the last in the `if` condition of `BasicTransform`

rabinadk1:
The order of the `if` condition is reordered so that the local variable is accessed first, then the instance variable, and finally call to the `random` module is made. Since or can stop evaluating when the first statement evaluated `True` is found, it makes the statement efficient.

### DIFF
--- a/albumentations/core/transforms_interface.py
+++ b/albumentations/core/transforms_interface.py
@@ -98,7 +98,7 @@ class BasicTransform(Serializable):
 
             return kwargs
 
-        if (random.random() < self.p) or self.always_apply or force_apply:
+        if force_apply or self.always_apply or (random.random() < self.p):
             params = self.get_params()
 
             if self.targets_as_params:


ternaus:
In practice variable `force_apply or self.always_apply` is typically equal to `false and false = false`

=> What is important is the value of `random.random() < self.p`

ternaus:
@Dipet What do you think?

Dipet:
Sounds logical. I am not sure how expensive a random call, but if we want to get high rps it's better to avoid uneccesery calls.
But in general we very rarely use `always_apply` flag and we call random in 90% of cases.
Additionally, this benchmark shows a very low impact of random call - aroud 0.7%. But custom code show 5% impact. I am not sure in what test believe more... But 5% is a huge impact.
I think we have nothing to lose with these changes, and merging it will be ok - in a good case we will get 5% speedup.

```python
import random
import time


def call_with_random(flag: bool = False):
    if flag or random.random() < 0.5:
        return True
    return False


def call_without_random(flag: bool = False):
    if random.random() < 0.5 or flag:
        return True
    return False


def bench(func, n: int = int(1e8)) -> float:
    s = time.time()
    for _ in range(n):
        func()
    return time.time() - s


if __name__ == '__main__':
    import timeit
    print(timeit.timeit("call_with_random()", setup="from __main__ import call_with_random", number=int(1e8)))
    print(timeit.timeit("call_without_random()", setup="from __main__ import call_without_random", number=int(1e8)))
    print(bench(call_with_random, int(1e8)))
    print(bench(call_without_random, int(1e8)))


# 8.588571457
# 8.528335685000002
# 8.731735467910767
# 8.299729824066162
```

ternaus:
If there is small, but nonzero improvement -> why not? Let's merge it.

ternaus:
My numbers:
```
9.243369000032544
9.3151286249049
10.138988971710205
10.394120216369629
```

and

```
9.400188999949023
9.401099791983142
10.169310092926025
9.768703937530518
```

They all look with errorbars from each other

Dipet:
Interesting.
The results are really inconsistent - it looks like the impact is very low, and background tasks have a big impact on this benchmark

Dipet:
ok, i understand - benchmark was wrong... My fail
we need to change default flag - impact with another order of args  we have 2x speedup.
```python
import random
import time


def call_with_random(flag: bool = True):
    if random.random() < 0.5 or flag:
        return True
    return False


def call_without_random(flag: bool = True):
    if flag or random.random() < 0.5:
        return True
    return False


def bench(func, n: int = int(1e8)) -> float:
    s = time.time()
    for _ in range(n):
        func()
    return time.time() - s


if __name__ == '__main__':
    import timeit
    print(timeit.timeit("call_with_random()", setup="from __main__ import call_with_random", number=int(1e8)))
    print(timeit.timeit("call_without_random()", setup="from __main__ import call_without_random", number=int(1e8)))
    print(bench(call_with_random, int(1e8)))
    print(bench(call_without_random, int(1e8)))

# 8.865977536999992
# 3.8430869389999884
# 8.653024435043335
# 3.906151294708252
```

Dipet:
reopen

PR #1450, opened 2023-05-17
add random tile

xbkaishui:


ternaus:
@xbkaishui is not it [RandomGridShuffle](https://albumentations.ai/docs/api_reference/full_reference/?h=randomgrid#albumentations.augmentations.transforms.RandomGridShuffle)?

PR #1489, opened 2023-10-25
Eliminate the possibility of the perspective transform collapsing

### LINKED ISSUE #1001: Undefined behaviour of albumentations.Perspective()
## 🐛 Bug

It happens on rare occasions that `albumentations.Perspective()` returns a completely distorted image.
The bug in the example code is caught because the keypoints get transformed outside the image plane (despite the `fit_output` argument being set), I suspect that that is not the only time such behaviour arises.  

## To Reproduce

Run the following code:
__Edited__: added random seeding as instructed in #1000

```python
import numpy as np
from tqdm import tqdm
import albumentations as A
import cv2
import random

random.seed(42)
np.random.seed(42)


perspective = A.Compose([
    A.Perspective(
        pad_mode=cv2.BORDER_CONSTANT,
        fit_output=True,
        always_apply=True
    )
], keypoint_params=A.KeypointParams(format='xy')
)


def apply(
        image: np.ndarray,
        keypoints: np.ndarray,
):
    augmented = perspective(
        image=image,
        keypoints=keypoints
    )

    return augmented['image'], augmented['keypoints']


def plot(
        title: str,
        image: np.ndarray,
        keypoints: np.ndarray,
):
    plt.title(title)
    plt.imshow(image)
    plt.scatter(*zip(*keypoints))
    plt.xticks([])
    plt.yticks([])
    plt.savefig(f"{title}.png")
    plt.cla()


if __name__ == '__main__':
    image = np.ones((126, 224, 3))
    keypoints = np.array(
        [
            [11, 8],
            [14, 115],
            [196, 114],
            [201, 13]
        ]
    )

    plot(
        title="Input",
        image=image,
        keypoints=keypoints,
    )

    # Run until distortion  
    for _ in tqdm(range(315505)):
        augmented_image, augmented_keypoints = apply(
            image=image,
            keypoints=keypoints
        )

    plot(
        title="Result",
        image=augmented_image,
        keypoints=augmented_keypoints,
    )
```

## Example outputs for different random seeds:
Otherwise, see my [example input](https://drive.google.com/uc?id=1BC62XGGcjXJB1C6uVgmE0zTQ-n42AWj5), [expected](https://drive.google.com/uc?id=1j5qIRGECYBCtgkfe_P8XTbdOmuGiWRGS) (and most common) output and two examples of distorted output: [(1)](https://drive.google.com/uc?id=1Iu_DjpWXFTZuWu1hXeoxr-K3uNCPPNCF), [(2)](https://drive.google.com/uc?id=1dq46LuUcV2u-3o6556EgKRvWJePQOuXr). 

## Expected behaviour:
Something like [this](https://drive.google.com/uc?id=1j5qIRGECYBCtgkfe_P8XTbdOmuGiWRGS). 

## Environment:
Albumentations version: `1.0.3`
Python version: `3.9.6`
OS (e.g., Linux): `macOS Big Sur 11.4`
How you installed albumentations: `pip`

alicangok:
Prior to this change, while the perspective transform could collapse at any selection of the `scale` parameter, it was especially more prevelant for higher values of `scale`. Specifically, the randomly generated transformation polygon had a non-zero chance of being concave, which led to collapsed perspective transformations.

A more refined (and easier to understand) solution requires the implementation of "Concave Polygon Recoverer" like in imgaug. That being said, simply restricting the randomly generated `points` values to the range [0, 1/3] ensures geometrically that the generated polygon will be convex, hence resulting in a valid (uncollapsed) perspective transformation.

In the code, I set the upper limit at 0.32 instead of 1/3, to avoid any potential numerical issues (due to near-concave polygons), with the same motivation behind the piece of code ensuring that `min_height` and `min_width` are over 2.0, shortly following the polygon generation.

Fixes #1001

### DIFF
--- a/albumentations/augmentations/geometric/transforms.py
+++ b/albumentations/augmentations/geometric/transforms.py
@@ -341,7 +341,7 @@ class Perspective(DualTransform):
 
         scale = random_utils.uniform(*self.scale)
         points = random_utils.normal(0, scale, [4, 2])
-        points = np.mod(np.abs(points), 1)
+        points = np.mod(np.abs(points), 0.32)
 
         # top left -- no changes needed, just use jitter
         # top right


alicangok:
image

PR #1545, opened 2024-02-28
Add dithering augmentation

DmitryKorolev:


albumentations/augmentations/functional.py
```diff
@@ -1424,3 +1425,67 @@ def spatter(
         raise ValueError("Unsupported spatter mode: " + str(mode))
 
     return img * 255
+
+@clipped
+@preserve_shape
+def dither(img: np.ndarray, nc: int) -> np.ndarray:
+    img = img.copy()
+    height = np.shape(img)[0]
+    is_rgb = True if len(np.shape(img)) == 3 else False
```
ternaus, line 1434: Better to use is_rgb_image function from albumentations/augmentations/utils

albumentations/augmentations/transforms.py
```diff
@@ -2785,3 +2786,44 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+        https://en.wikipedia.org/wiki/Floyd%E2%80%93Steinberg_dithering
+    """
+
+    def __init__(
+        self,
+        nc: int = 2,
+        always_apply: bool = False,
+        p: float = 0.5,
+    ):
+        super().__init__(always_apply, p)
```
ternaus, line 2819: nv is positive, we sjould have a check here

albumentations/augmentations/transforms.py
```diff
@@ -2785,3 +2786,44 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
+        uint8, float32
+
+    References :
+        https://en.wikipedia.org/wiki/Dither
+        https://en.wikipedia.org/wiki/Floyd%E2%80%93Steinberg_dithering
+    """
+
+    def __init__(
+        self,
+        nc: int = 2,
```
ternaus, line 2815: nc is not really intuituve name.

`num_colors` would be better

albumentations/augmentations/functional.py
```diff
@@ -1424,3 +1425,67 @@ def spatter(
         raise ValueError("Unsupported spatter mode: " + str(mode))
 
     return img * 255
+
+@clipped
+@preserve_shape
+def dither(img: np.ndarray, nc: int) -> np.ndarray:
+    img = img.copy()
+    height = np.shape(img)[0]
```
Dipet, line 1433: I think better to use class method `img.shape[0]`. I think it's more readable

albumentations/augmentations/functional.py
```diff
@@ -1424,3 +1425,67 @@ def spatter(
...
+    for y in range(height):
+        oldrow = img[y].copy()
+        quant_errors = []
+
+        if is_rgb:
+            # Turn into one list of length `width`, per channel (R, G, B)
+            #
+            # Use `tolist()` since operating on individual elements of an ndarray
+            # is very slow compared to a normal list.
+            channels = np.transpose(oldrow).tolist()
```
Dipet, line 1445: Why we need to convert to list? Why do not use np.ndarray `channels = oldrow.transpose()`? Result would be the same, but faster.

albumentations/augmentations/functional.py
```diff
@@ -1424,3 +1425,67 @@ def spatter(
...
+
+            # Transpose back to one list containing all channels
+            # and replace the row
+            img[y] = np.transpose(channels)
+            quant_errors = np.transpose(quant_errors)
+        else:
+            img[y], quant_errors = _apply_dithering_to_channel(img[y].tolist(), nc)
+
+        if y < height - 1:
+            zero_or_zeros = 0 if np.shape(quant_errors[-1]) == () else np.zeros_like(quant_errors[-1])
```
Dipet, line 1459: Just use `np.zeros_like(quant_errors[-1])` this if else is useles

albumentations/augmentations/functional.py
```diff
@@ -1424,3 +1425,67 @@ def spatter(
...
+
+@clipped
+@preserve_shape
+def dither(img: np.ndarray, nc: int) -> np.ndarray:
+    img = img.copy()
+    height = np.shape(img)[0]
+    is_rgb = True if len(np.shape(img)) == 3 else False
+
+    for y in range(height):
+        oldrow = img[y].copy()
```
Dipet, line 1437: we don't need clone there

albumentations/augmentations/functional.py
```diff
@@ -1424,3 +1425,67 @@ def spatter(
         raise ValueError("Unsupported spatter mode: " + str(mode))
 
     return img * 255
+
+@clipped
+@preserve_shape
+def dither(img: np.ndarray, nc: int) -> np.ndarray:
+    img = img.copy()
```
Dipet, line 1432: It looks like beetter to use empty:
```python
result = np.empty_like(img)
result[0] = img[0]
```
It is much faster

albumentations/augmentations/functional.py
```diff
@@ -1424,3 +1425,67 @@ def spatter(
...
+            r1 = np.roll(quant_errors, -1, axis=0) * (3 / 16)
+            r1[-1] = zero_or_zeros
+            r2 = np.roll(quant_errors, 1, axis=0) / 16
+            r2[0] = zero_or_zeros
+            updated_row = r1 + r2 + np.array(quant_errors) * (5 / 16)
+            img[y + 1] = (img[y + 1] + updated_row).astype(img.dtype)
+    return img
+
+
+def _apply_dithering_to_channel(ch, nc):
```
Dipet, line 1469: typing

albumentations/augmentations/functional.py
```diff
@@ -1424,3 +1425,67 @@ def spatter(
...
+    # on the original value instead of the value it
+    # got after being affected by error propagation.
+    quant_error = [0] * width
+
+    for x in range(width - 1):
+        oldval = ch[x]
+        newval = round(oldval * (nc - 1)) / (nc - 1)
+        ch[x] = newval
+        quant_error[x] = oldval - newval
+        ch[x + 1] += quant_error[x] * (7 / 16)
```
Dipet, line 1483: Let's vectorize this
```python
new_val = np.round(quant_error[:-1] * (nc - 1) * (1 / (nc - 1))
quant_error = ch[:-1] - new_val
ch[:-1] = new_val
ch[1:] += quant_error * (7 / 16)
```

Dipet (changes requested):

PR #1549, opened 2024-02-29
[WIP] Mix up

ternaus:


albumentations/augmentations/mixing/transforms.py
```diff
@@ -0,0 +1,116 @@
...
+        - If no reference data is provided, this transform will issue a warning and act as a no-op.
+        - The lambda value used for mixing is sampled once per batch and applied globally across all images
+          and labels within that batch.
+        - The class labels are expected to be numpy arrays, enabling the mixing process when applicable.
+
+    """
+
+    def __init__(
+        self,
+        reference_data: Sequence[ReferenceImage] = [],
```
Dipet, line 70: I think better to use some generator to get images. Some users might want to take provide image and some additional data into this generator to generate reference image.

albumentations/augmentations/mixing/transforms.py
```diff
@@ -0,0 +1,116 @@
...
+        read_fn: Callable[[Any], np.ndarray] = lambda x: x,
+        alpha: float = 0.4,
+        always_apply: bool = False,
+        p: float = 0.5,
+    ):
+        super().__init__(always_apply, p)
+        if alpha < 0:
```
Dipet, line 77: I think alpha must be in range [0, 1]

Dipet (changes requested):

albumentations/augmentations/mixing/transforms.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def apply(self, img: np.ndarray, mix_data: ReferenceImage, mix_coef: float, **params: Any) -> np.ndarray:
+        mix_img = mix_data.get("image")
+        return mix_arrays(img, mix_img, mix_coef) if mix_img is not None else img
+
+    def apply_to_mask(self, mask: np.ndarray, mix_data: ReferenceImage, mix_coef: float, **params: Any) -> np.ndarray:
+        mix_mask = mix_data.get("mask")
+        return mix_arrays(mask, mix_mask, mix_coef) if mix_mask is not None else mask
+
+    def apply_to_global_label(
```
Dipet, line 101: I think we must bun bbox and keypoint transforms or add info into docstring about how transorm process them.

Dipet (changes requested):

PR #1570, opened 2024-03-09
Fix pad if needed serialization

ternaus:


albumentations/augmentations/crops/transforms.py
```diff
@@ -502,13 +502,13 @@ class RandomCropNearBBox(DualTransform):
     def __init__(
         self,
         max_part_shift: ScaleFloatType = (0.3, 0.3),
-        cropping_box_key: str = "cropping_bbox",
+        cropping_bbox_key: str = "cropping_bbox",
```
Dipet, line 505: This will might break backward compatibility. Are you sure that this is a good idea - rename this parameter?

albumentations/core/serialization.py
```diff
@@ -165,6 +167,15 @@ def check_data_format(data_format: str) -> None:
         raise ValueError(f"Unknown data_format {data_format}. Supported formats are: 'json' and 'yaml'")
 
 
+def serialize_enum(obj: Any) -> Any:
+    """Recursively search for Enum objects and convert them to their value."""
+    if isinstance(obj, dict):
+        return {k: serialize_enum(v) for k, v in obj.items()}
+    if isinstance(obj, list):
+        return [serialize_enum(v) for v in obj]
```
Dipet, line 175: I think better to compare with typing.Mapping and  typing.Sequence

Dipet (changes requested):

albumentations/augmentations/crops/transforms.py
ternaus: I do not like that naming is confusing across the class, but you are right.

Let's do it smoother, through the deprecation warning.

Dipet (approved):

PR #1578, opened 2024-03-12
Refactor grid shuffle

ternaus:


albumentations/augmentations/functional.py
```diff
@@ -1395,3 +1395,30 @@ def spatter(
...
+    # Compute split points for the grid
+    height_splits = np.linspace(0, height, n_rows + 1, dtype=int)
+    width_splits = np.linspace(0, width, n_cols + 1, dtype=int)
+
+    # Initialize tiles coordinates
+    tiles = []
+
+    # Calculate tiles coordinates using list comprehension for efficiency
+    tiles = [(height_splits[i], width_splits[j], height_splits[i + 1], width_splits[j + 1]) 
+             for i in range(n_rows) for j in range(n_cols)]
```
Dipet, line 1422: It looks like might be vectorized. Right now I can not figure out how to do this. But latter I will try show some workaround.

ternaus: I do not see how to do it either, and ChatGPT does not propose anything that works.

Also here `n_cols` and `n_rows` are rather small, less than 10 each, and most typical value is 2.

Hence vectorization would be beneficial, but benefits would be limited.

PR #1605, opened 2024-03-22
Add pydantic

ternaus:


conda.recipe/build_upload.sh
Dipet, line 1: empty file

albumentations/augmentations/geometric/transforms.py
```diff
@@ -1342,6 +1343,14 @@ class HorizontalFlip(DualTransform):
 
     _targets = (Targets.IMAGE, Targets.MASK, Targets.BBOXES, Targets.KEYPOINTS)
 
+    def __init__(
+        self,
+        always_apply: bool = False,
+        p: float = 0.5,
+    ):
+        config = BaseTransformConfig(always_apply=always_apply, p=p)
```
Dipet, line 1351: This must be inside `BasicTransform.__init__`

albumentations/augmentations/transforms.py
```diff
@@ -105,14 +140,11 @@ class RandomGridShuffle(DualTransform):
...
     def __init__(self, grid: Tuple[int, int] = (3, 3), always_apply: bool = False, p: float = 0.5):
-        super().__init__(always_apply, p)
-
-        n, m = grid
-
-        if not all(isinstance(dim, int) and dim > 0 for dim in [n, m]):
-            raise ValueError(f"Grid dimensions must be positive integers. Current grid dimensions: [{n}, {m}]")
-
-        self.grid = grid
+        config = RandomGridShuffleConfig(grid=grid, always_apply=always_apply, p=p)
```
Dipet, line 143: Maybe we can do this using metcaclasses...
Look SerializableMeta as an example https://github.com/albumentations-team/albumentations/blob/main/albumentations/core/serialization.py#L44

PR #1622, opened 2024-04-04
read images with torchvision.io

ayasyrev:


ternaus:
We are not planning adding `torchvision` dependency at the moment. But reading images with `opencv` is indeed twice slower than in `torchvision`

Issue #1625, opened 2024-04-05
RandomGridShuffle Does Not Apply Any Transformation


mjorrico:
### Description
I applied the `RandomGridShuffle` transformation to an image but it returns the very same input image. I could be missing something so I put this under "Bug".

OS: Ubuntu 22.04.4 LTS
Python 3.10.12
Albumentations 1.4.3
Numpy 1.26.4

### The Code

```
import cv2
import albumentations as A

transformation = A.Compose(
    [
        A.RandomGridShuffle(grid=(4, 4), always_apply=True),
    ]
)

img = cv2.imread("input.jpg")
img = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
aug = transformation(image=img)
img = aug["image"]

cv2.imwrite("augresult.jpg", img)
```

### Expected behavior

It should always return the input image but with a randomized tiling.

### Actual behavior

It returns the same input image

### Context
Running `file input.jpg` gives me the following output
```
$ file input.jpg
input.jpg: JPEG image data, JFIF standard 1.01, aspect ratio, density 1x1, segment length 16, baseline, precision 8, 1000x1000, components 1
```

ternaus:
Fixed in https://github.com/albumentations-team/albumentations/pull/1626

PR #1626, opened 2024-04-05
Fix random grid shuffle

ayasyrev:
#1625


ternaus:
We also need to add test that shows that tiles are indeed shuffled.

ayasyrev (commented):
Fix tests - change defaults settings at serialization, check image changed

ayasyrev (commented):
Add warning if image size not dividable by grid size.
Add test that image not changed and got warning

ayasyrev (commented):
rename image_shape -> height, weight

Issue #1638, opened 2024-04-08
Augmentation for multiple keypoints is broken


domef:
## Describe the bug

The augmentation for multiple keypoints works only if the `additional_targets` are defined in the constructor whereas if they are added with `add_targets` the angle of the second keypoints are wrong.

First example:
```python
import numpy as np
import albumentations as A

transform = A.Compose(
    transforms=[A.Rotate(limit=(90.0, 90.0), p=1.0)],
    keypoint_params=A.KeypointParams(
        angle_in_degrees=True,
        check_each_transform=True,
        format="xyas",
        label_fields=None,
        remove_invisible=False,
    ),
    additional_targets={"keypoints2": "keypoints"},
)

image = np.zeros((100, 100, 3), dtype=np.uint8)
kps1 = np.array([[10, 10, 1, 0], [20, 20, -100, 0]], dtype=np.float64)
kps2 = np.array([[20, 20, 190, 0], [40, 40, 123, 0]], dtype=np.float64)

out = transform(image=image, keypoints=kps1, keypoints2=kps2)

kps1_t = out["keypoints"]
kps2_t = out["keypoints2"]

print(kps1)
print(kps1_t)
print()
print(kps2)
print(kps2_t)
```
The output is correct:
```
[[  10.   10.    1.    0.]
 [  20.   20. -100.    0.]]
[(9.999999999999993, 89.0, 91.0, 0.0), (19.999999999999993, 79.0, 350.0, 0.0)]

[[ 20.  20. 190.   0.]
 [ 40.  40. 123.   0.]]
[(19.999999999999993, 79.0, 280.0, 0.0), (39.99999999999999, 59.0, 213.0, 0.0)]
```

Second example:
```python
import numpy as np
import albumentations as A

transform = A.Compose(
    transforms=[A.Rotate(limit=(90.0, 90.0), p=1.0)],
    keypoint_params=A.KeypointParams(
        angle_in_degrees=True,
        check_each_transform=True,
        format="xyas",
        label_fields=None,
        remove_invisible=False,
    ),
)
transform.add_targets({"keypoints2": "keypoints"})

image = np.zeros((100, 100, 3), dtype=np.uint8)
kps1 = np.array([[10, 10, 1, 0], [20, 20, -100, 0]], dtype=np.float64)
kps2 = np.array([[20, 20, 190, 0], [40, 40, 123, 0]], dtype=np.float64)

out = transform(image=image, keypoints=kps1, keypoints2=kps2)

[...]

Issue #1372, opened 2022-12-26
The `add_targets` method sets targets instead of adding them


ernest-tg:
Suppose you do 
```python
import albumentations as A
t = A.ToGray()  # Works with all transformations
t.add_targets({"my_image1": "image"})
t.add_targets({"my_image2": "image"})
print(t._additional_targets)
```
You get `{'my_image2': 'image'}`. That seems to be what you want since you (albu) wrote 

```python
class BasicTransform(Serializable):
    ...
    def add_targets(self, additional_targets):
        ...
        self._additional_targets = additional_targets
```

But, given the name of the function, and the docstring "Add targets to transform them the same way as one of existing targets", I expected `{'my_image1': "image", 'my_image2': "image"}`.

My own use case is very uncommon, but I could see other people adding some targets in 2 different places for some other reason. I would suggest either:
* Replace `self._additional_targets = additional_targets` by `self._additional_targets.update(additional_targets)`
* Rename `add_targets` to `set_targets` / `set_additional_targets`

For completeness purpose, my own use case is that I have a child class (named `Tee`) of `BasicTransform` which outputs 2 keys from one key. So my pipeline looks like:

```python
before_tee1 = A.SomeTransform(...)
before_tee2 = A.SomeTransform(...)
tee = Tee(...)
after_tee1 = A.SomeTransform(...)
after_tee2 = A.SomeTransform(...)
for transfo in [after_tee1, after_tee2]:
    transfo.add_targets({"image_copy": "image"})

...

dynamic_composed_transfo = A.Compose(
    [before_tee1, before_tee2, tee, after_tee1, after_tee2], additional_targets=dynamic_targets)
```

Dipet:
This method designed for internal use only. How to use additional targets you could read there: https://albumentations.ai/docs/examples/example_multi_target/

ernest-tg:
I can write the pipeline above using only the `additional_targets` parameter from `Compose`, but I still have the same issue. Or, for a simpler pipeline:

```
import albumentations as A
t1 = A.Flip()
t2 = A.ToGray()
c2 = A.Compose([t2], additional_targets={'image_for_t2': "image"})
c = A.Compose([t1, c2], additional_targets={'image_for_t1_t2': "image"})

print(c.transforms[1].transforms[0]._additional_targets)
```
I would expect the targets of `t1` to be "image_for_t1_t2", and the targets of `t2` to be both "image_for_t1_t2" and "image_for_t2". 
[...]

Issue #422, opened 2019-10-15
Multiple images


theRealSuperMario:
Hi,

I just found this amazing library for image augmentation and wanted to adapt it my usecase.
Is it possible to apply the same augmentation that I get from the compose interface to a list of images.

Kinda like
```
data = {"image_1": im1, "image_2": im2}
augmented = augmentation(**data)
```

Could not get it to work so far. Thanks in advance

theRealSuperMario:
Sorry, I found the solution like 5 minutes later. Just in case anyone else is wondering, here is the minimal example.


```python
from albumentations import (
    HorizontalFlip, IAAPerspective, ShiftScaleRotate, CLAHE, RandomRotate90,
    Transpose, ShiftScaleRotate, Blur, OpticalDistortion, GridDistortion, HueSaturationValue,
    IAAAdditiveGaussianNoise, GaussNoise, MotionBlur, MedianBlur, IAAPiecewiseAffine,
    IAASharpen, IAAEmboss, RandomBrightnessContrast, Flip, OneOf, Compose
)

def strong_aug(p=0.5):
    return Compose([
        RandomRotate90(),
        Flip(),
        Transpose(),
        OneOf([
            IAAAdditiveGaussianNoise(),
            GaussNoise(),
        ], p=0.2),
        OneOf([
            MotionBlur(p=0.2),
            MedianBlur(blur_limit=3, p=0.1),
            Blur(blur_limit=3, p=0.1),
        ], p=0.2),
        ShiftScaleRotate(shift_limit=0.0625, scale_limit=0.2, rotate_limit=45, p=0.2),
        OneOf([
            OpticalDistortion(p=0.3),
            GridDistortion(p=0.1),
            IAAPiecewiseAffine(p=0.3),
        ], p=0.2),
        OneOf([
            CLAHE(clip_limit=2),
            IAASharpen(),
            IAAEmboss(),
            RandomBrightnessContrast(),
        ], p=0.3),
        HueSaturationValue(p=0.3),
    ], p=p,
    additional_targets={"image2" : "image"})

import numpy as np
from skimage import data
from matplotlib import pyplot as plt

image1 = data.astronaut().astype(np.uint8)
image2 = data.astronaut().astype(np.uint8)
aug_input = {"image" : image1, "image2" : image2}
aug = strong_aug(0.9)
augmented_data = aug(**aug_input)
i1 = augmented_data["image"]
i2 = augmented_data["image2"]

plt.subplot(121)
[...]

PR #1641, opened 2024-04-09
fix add_target interface

ayasyrev:
Fix add_target interface.
Related #1372 #1638 #422


albumentations/core/composition.py
```diff
@@ -114,8 +120,11 @@ def get_dict_with_id(self) -> Dict[str, Any]:
 
     def add_targets(self, additional_targets: Optional[Dict[str, str]]) -> None:
         if additional_targets:
+            self._additional_targets = {**self._additional_targets, **additional_targets}
```
Dipet, line 123: I think it would be great to check keys. And if they are exist, check if they are the same

Dipet (changes requested):

albumentations/core/composition.py
ayasyrev: If we use this way - second dict keys overwrite keys from first. Second is after transforms. If no transforms applied to data - it will be same. And here we got references to same objects.

Dipet: Somethink like this:
```python
for k, v in additional_targets.items():
    if k in self._additional_targets and not v == self._additional_targets[k]:
        raise ValueError(f"Trying to owervwrite existed additional targets. Key={k} Exists={self._additional_targets[k]} New value: {v}")
    self._additional_targets[k] = v
```

ayasyrev: Ok, let's do it this way. It names `add`, not change...

Dipet (approved):

Issue #1633, opened 2024-04-05
[Tech debt] Improve interface for CoarseDropout


ternaus:
* `min_holes`, `max_holes` => `num_holes_range`
* `min_height`, `max_height` => `hole_height_range`
* `min_width`, `max_width` => `hole_width_range`
=>

We can update transform to use new signature, keep old as working, but mark as deprecated.
----
PR could be similar to https://github.com/albumentations-team/albumentations/pull/1615

ternaus:
Addressed in https://github.com/albumentations-team/albumentations/pull/1677

PR #1651, opened 2024-04-11
add num_holes_range, holes_width_range, holes_height_range to use instead of {min,max}_(holes,width,height}

dokazhi:
(https://github.com/albumentations-team/albumentations/issues/1633#issue-2228684092)

### DIFF
--- a/albumentations/augmentations/dropout/coarse_dropout.py
+++ b/albumentations/augmentations/dropout/coarse_dropout.py
@@ -1,4 +1,5 @@
 import random
+import warnings
 from typing import Any, Dict, Iterable, List, Optional, Sequence, Tuple, cast
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


albumentations/augmentations/dropout/coarse_dropout.py
```diff
@@ -129,39 +186,26 @@ def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, A
...
-                [
-                    isinstance(self.min_height, float),
-                    isinstance(self.min_width, float),
-                    isinstance(self.max_height, float),
-                    isinstance(self.max_width, float),
-                ]
-            ):
-                hole_height = int(height * random.uniform(self.min_height, self.max_height))
-                hole_width = int(width * random.uniform(self.min_width, self.max_width))
+        num_holes_range: Tuple[int, int] = (
```
ternaus, line 189: No need to compute `num_holes_range` in `apply`.

It is already done in `InitSchema` in model_validator and all parameters code in `__init__` are already validated and preprocessed => `self.num_holes_range` and other ranges could be directly used here.

ternaus (changes requested):

ternaus:
We also need tests that show that both old and new versions of the interface work.

Something like this: https://github.com/albumentations-team/albumentations/blob/main/tests/test_transforms.py#L1286

Or at least as: https://github.com/albumentations-team/albumentations/blob/main/tests/test_transforms.py#L1360

albumentations/augmentations/dropout/coarse_dropout.py
dokazhi: `    def get_params_dependent_on_targets(self, params: Dict[str, Any]) -> Dict[str, Any]:
        img = params["image"]
        height, width = img.shape[:2]

        holes = []

        for _ in range(random.randint(*self.num_holes_range)):
            if all(isinstance(x, int) for x in self.hole_height_range + self.hole_width_range):
                hole_height = random.randint(*self.hole_height_range)
                hole_width = random.randint(*self.hole_width_range)
            elif all(isinstance(x, float) for x in self.hole_height_range + self.hole_width_range):
                hole_height = int(height * random.uniform(*self.hole_height_range))
                hole_width = int(width * random.uniform(*self.hole_width_range))
            else:
                msg = f"Hole width range and hole height range \
                    values should all either be ints or floats. \
                    Got: {[type(x) for x in self.hole_height_range + self.hole_width_range]} respectively"
                raise ValueError(msg)

            y1 = random.randint(0, height - hole_height)
            x1 = random.randint(0, width - hole_width)
            y2 = y1 + hole_height
            x2 = x1 + hole_width
            holes.append((x1, y1, x2, y2))

        return {"holes": holes}`

I did it like that in first time but mypy errors made me refactor

ternaus: Could you please share errors?
---
It is better to fix, even in hackish way mypy errors than code quality lower.
----
I suspect that if you do cast to Tuple in lines:

```python
self.num_holes_range = num_holes_range
self.hole_width_range = hole_width_range
self.hole_height_range = hole_height_range
```

say 
```
self.num_holes_range = cast(Tuple[int, int], num_holes_range)
```
the error will go away

ternaus:
Addressed in https://github.com/albumentations-team/albumentations/pull/1677

Issue #1645, opened 2024-04-10
[Feature Request] Make RandomGridShuffle work when `size % num_splits != 0`


ternaus:
Right now if side is non divisible by the number of splits => we do not do anything.

Better would be:

image shape  `(20, 17)`
Want `(3, 4)` splits

=> 20 = 6 + 7 + 7
=> 17 = 4 + 4 + 5 + 4

After that perform shuffle on every tuple:

Let it become: (7, 6, 7) и (4, 5, 4, 4)

Now we have 12 tiles with different sized => shuffle those that have the same size.

=>

6 tiles 7x4
2 tiles 7x5
3 tiles 6x4

ternaus:
Fixed in https://github.com/albumentations-team/albumentations/pull/1655

PR #1655, opened 2024-04-12
Update random grid shuffle

ternaus:
Addressing https://github.com/albumentations-team/albumentations/issues/1645


ternaus:
Fixed in https://github.com/albumentations-team/albumentations/pull/1655

PR #1670, opened 2024-04-17
check_each_transform - move calculation from`__call__` to init

ayasyrev:
Move calculation `check_each_transform` -  from`__call__` to `__init__`


ayasyrev:
At `_check_data_post_transform` we need to check additional_targets

PR #1679, opened 2024-04-22
Adds error message to the add_label_fields_to_data function.

wilderrodrigues:
When the lengths os bboxes and labels mismatch, raise the ValueError with a proper error message.


ternaus:
FIxed in https://github.com/albumentations-team/albumentations/pull/1687

PR #1691, opened 2024-04-27
Issue/update image compression interface

MarognaLorenzo:
# New interface
The new interface allow users to set `quality_range` field and be ready to go. Implementations that are setting `quality_upper` and `quality_lower` will raise a `DeprecationWarning`. 

## Behaviour

1. Default uses such as `transform_name = A.ImageCompression()` won't raise any warning (default values = (99, 100))
2. `transform_name = A.ImageCompression(lower, upper)` will raise a Warning
3. `transform_name = A.ImageCompression(quality_lower=lower, quality_upper= upper)` will raise a Warning
4. `transform_name = A.ImageCompression(quality_range=[99,100])` won't raise any warning, even if a tuple is passed ( but they need to have a length of 2, both in the [1,100] range and the first one must be lower than the second one
5. `transform_name = A.ImageCompression([99,100])`will raise an error because the first 2 expected arguments are `quality_lower` and `quality_upper`. This is the drawback for allowing code like at point 2 to raise a warning without failing in existing code.


albumentations/augmentations/transforms.py
```diff
@@ -318,8 +318,12 @@ class ImageCompression(ImageOnlyTransform):
...
 
     Args:
-        quality_lower: lower bound on the image quality. Should be in [0, 100] range for jpeg and [1, 100] for webp.
-        quality_upper: upper bound on the image quality. Should be in [0, 100] range for jpeg and [1, 100] for webp.
+        quality_lower: **Deprecated: use 'quality_range' instead** lower bound on the image quality.
+            Should be in [0, 100] range for jpeg and [1, 100] for webp.
+        quality_upper: **Deprecated: use 'quality_range' instead** upper bound on the image quality.
+            Should be in [0, 100] range for jpeg and [1, 100] for webp.
+        quality_range: tuple of bounds on the image quality i.e. (quality_lower, quality_upper).
+            Both values should be in [0, 100] range for jpeg and [1, 100] for webp.
```
ternaus, line 326: Just remove `quality_lower` and `quality_upper` from docstring. 

If someone uses them -> that's fine, but it could be better if all new users would not even know about such option.

albumentations/augmentations/transforms.py
```diff
@@ -332,32 +336,70 @@ class ImageCompression(ImageOnlyTransform):
     """
 
     class InitSchema(BaseTransformInitSchema):
-        quality_lower: int = Field(default=99, description="Lower bound on the image quality", ge=1, le=100)
-        quality_upper: int = Field(default=100, description="Upper bound on the image quality", ge=1, le=100)
+        _size_len = 2
+        quality_range: Optional[Union[Tuple[int, int], List[int]]] = Field(
```
ternaus, line 340: Let's use for quality range: `Optional[Tuple[int, int]]`

Historically we had an option in all transforms to specify a number or a Tuple and number is converted to Tuple before sampling.

But there is so many different ways to do this, say

`x => (-x, x)`
`x => (x, x)`
`x => (1 - x, 1+x)`

etc that it became a mess.

New policy that for all new interfaces, including refactoring of the old one like here, or creating new one - if it should be number => make it a number, it range => it should be tuple.

We have this in Countributor's guide, but I should probably extend it there:

```
Explicit Definitions: Parameters should be explicitly defined as either a single value or a tuple. This change avoids ambiguity and ensures that the intent and behavior of the transform are clear to all users.
```

albumentations/augmentations/transforms.py
```diff
@@ -332,32 +336,70 @@ class ImageCompression(ImageOnlyTransform):
...
         @model_validator(mode="after")
-        def validate_quality(self) -> Self:
-            if self.quality_lower >= self.quality_upper:
-                msg = "quality_lower must be less than quality_upper"
-                raise ValueError(msg)
+        def validate_ranges(self) -> Self:
+            lower = 1
+            upper = 100
+
+            def validate_range(range: Tuple[int, int], name: str) -> None:
```
ternaus, line 366: Pydantic should do this validation for us.

When we say in `InitShema`:

`quality_range: Optional[Tuple[int, int]]`

It will throw an error if anything but None of Tuple[int, int]will be used => You can delete:

```
if len(range) != self._size_len:
                    msg = f"'quality_range' should have lenght of {self._size_len} but lenght of {len(range)} was found"
                    raise ValueError(msg)
```

albumentations/augmentations/transforms.py
```diff
@@ -332,32 +336,70 @@ class ImageCompression(ImageOnlyTransform):
...
+                    raise ValueError(msg)
+                for i, el in enumerate(range):
+                    if el < lower or el > upper:
+                        msg = f"element {i} in {name} has value {el} and not between {lower} and {upper}"
+                        raise ValueError
+                if range[0] > range[1]:
+                    msg = "quality lower bound must be less than quality upper bound"
+                    raise ValueError(msg)
+
+            if self.quality_lower or self.quality_upper:
```
ternaus, line 378: In pydantic 2.7 that we use, you add `deprectated` to the type definition in InitShema and if variable is used => deprecation warning will happen authomatically.

Example:

```python
min_holes: Optional[int] = Field(
            default=None,
            ge=0,
            description="Minimum number of regions to zero out.",
            deprecated="Use num_holes_range instead.",
        )
```
https://github.com/albumentations-team/albumentations/blob/main/albumentations/augmentations/dropout/coarse_dropout.py#L59

ternaus:
Almost there:

1. There are a few changes that leverage Pydantic syntax sugar that will simplify the code as some checks and warnings will happen automatically.
2. You need to install `pre-commit` hook and make sure that all checks pass before the commit. Details: https://albumentations.ai/docs/CONTRIBUTING/#using-pre-commit-hooks

P.S. `Pydantic` and `pre-commit hooks` are very powerful and popular tools in python software development. This commit is a good way to get acquainted to them :)

albumentations/augmentations/transforms.py
ternaus: `transform=A.ImageCompression()`

It is not optimal, but I think we can live with this till figure out better solution for warnings, or deprecate old interface (I think we can do it in fall)

ternaus:
There were some issues with CI config, 

I created another PR on top of your work and merged it.

All credit goes to you :)

Will mention your contribution in the next releaze.

https://github.com/albumentations-team/albumentations/pull/1704

PR #1711, opened 2024-05-07
Add version check

ternaus:


Antinomy20001:
We have noticed that the 'check_for_update' function only handles exceptions for certain cases (network exceptions). It does not fully cover the scenarios in complex network environments. However, this function will be called whenever 'import albumentations' is executed, which indicates its incompleteness.

ternaus:
Thanks. 

Will make exception general

Correct me, if I am wrong, but regardless of the exception type, nothing happens as it is just a `pass` after that.

ternaus:
Made exception more general in https://github.com/albumentations-team/albumentations/pull/1725

Issue #1715, opened 2024-05-10
[tech debt] Deprecate always_apply


ternaus:
I do not remember why did we introduce it in the first place, but looks like this functionality could be easily done by setting `p=1`

ternaus:
Addressed

PR #1759, opened 2024-05-31
deprecate `always_apply`

ayasyrev:
#1715

### DIFF
--- a/albumentations/core/composition.py
+++ b/albumentations/core/composition.py
@@ -43,16 +43,6 @@ TransformsSeqType = List[TransformType]
 AVAILABLE_KEYS = ("image", "mask", "masks", "bboxes", "keypoints", "global_label")
 
 
-def get_always_apply(transforms: Union["BaseCompose", TransformsSeqType]) -> TransformsSeqType:
-    new_transforms: TransformsSeqType = []
-    for transform in transforms:
-        if isinstance(transform, BaseCompose):
-            new_transforms.extend(get_always_apply(transform))
-        elif transform.always_apply:
-            new_transforms.append(transform)
-    return new_transforms
-
-
 def get_transforms_dict(transforms: TransformsSeqType) -> Dict[int, BasicTransform]:
     result = {}
     for transform in transforms:
@@ -236,7 +226,6 @@ class Compose(BaseCompose):
         self._disable_check_args_for_transforms(self.transforms)
 
         self.is_check_shapes = is_check_shapes
-        self._always_apply = get_always_apply(self.transforms)  # transforms list that always apply
         self._check_each_transform = tuple(  # processors that checks after each transform
             proc for proc in self.processors.values() if getattr(proc.params, "check_each_transform", False)
         )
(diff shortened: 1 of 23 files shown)


albumentations/core/transforms_interface.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, downscale: int = cv2.INTER_NEAREST, upscale: int = cv2.INTER_
 
 class BaseTransformInitSchema(BaseModel):
     model_config = ConfigDict(arbitrary_types_allowed=True)
-    always_apply: bool = Field(default=False, description="Always apply the transform")
+    always_apply: bool = Field(default=False, description="Deprecated. Use `p=0` instead to always apply the transform")
```
ternaus, line 37: Pydantic 2.7+ has deprecated field. =>

```python
always_apply: bool = Field(default=False, deprecated="Deprecated. Use `p=0` instead to always apply the transform")
```

albumentations/core/composition.py
```diff
@@ -272,13 +261,12 @@ def __call__(self, *args: Any, force_apply: bool = False, **data: Any) -> Dict[s
             data[self.save_key] = OrderedDict()
 
         need_to_run = force_apply or random.random() < self.p
-        if not need_to_run and not self._always_apply:
```
rakhil575, line 275: I think this should be reverted. I have RandomCrop with always_apply in Compose that has a probability of 0.9. Previously, the crop was always being performed but now it would only be performed 90% of the time.

ternaus: Can you provide an example of the pipeline where `always_apply` would be useful?

rakhil575: I have a dataset with images having different shapes. I use RandomCrop with `always_apply` in `Compose` so that the final images after transformations are always the same shape. Without `always_apply`, images in a batch would have different shapes and the batch formation would fail.

ternaus: Will RandomCrop with p=1 do the job?

rakhil575: Ideally it should but with the current implementation it doesn't. There is no function currently that checks if any transforms inside Compose has p=1.0. This function would be similar to _always_apply but also checks for transforms that has p=1.0.

ternaus: I do not get it.

```python
transform = A.Compose([A.RandomCrop(p=1), other transforms], p=1)
```

will do the job?

If not, could you please provide minimal example of the pipeline that shows the issue

rakhil575: This wouldn't work: `transform = A.Compose([A.RandomCrop(p=1), other transforms], p=0.9)`

ternaus: Let's close this conversation for now and continue the conversation when you could provide code that could not be executed with p=1

rakhil575: The transformation above is the code that wouldn't work now. I am mostly concerned about the behaviour change that this PR silently brings. In older version, `A.Compose([A.RandomCrop(p=1, always_apply=True), other transforms], p=0.9)` would always apply RandomCrop but with this PR, RandomCrop would only be applied with 0.9 probability with the same transform.

PR #1828, opened 2024-07-03
Prevent to get zeros array after transform by chainging order of asty…

rubai1597:
This pull request fixes an issue with getting zeros array after ISONoise transform.
I fixed it by changing the order of astype operations.
It seems that this issue can occur depending on the version of numpy.

## Summary by Sourcery

This pull request addresses a bug in the ISONoise transform function that caused it to produce an array of zeros. The issue was resolved by adjusting the order of the astype operations, ensuring compatibility across different versions of numpy.

- **Bug Fixes**:
    - Fixed an issue where the ISONoise transform resulted in an array of zeros by changing the order of astype operations.

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -801,7 +801,7 @@ def iso_noise(
         raise TypeError(msg)
 
     one_over_255 = float(1.0 / 255.0)
-    image = multiply(image, one_over_255).astype(np.float32)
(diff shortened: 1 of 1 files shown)


ternaus:
Thanks about reporting the issue.

Addressed it in a more extended way in https://github.com/albumentations-team/albumentations/pull/1829

PR #1907, opened 2024-08-23
Add pytorch import

LightAllWorld:
Added the missing pytorch import to the __init__ file.
This addition prevents potential errors during the transform automation process.

Please review the PR, and feel free to provide any feedback. Thank you!

## Summary by Sourcery

Fix import error by adding the missing PyTorch import in the __init__.py file to ensure proper functionality of the transform automation process.

Bug Fixes:
- Add missing import for PyTorch in the __init__.py file to prevent potential errors during the transform automation process.

### DIFF
--- a/albumentations/__init__.py
+++ b/albumentations/__init__.py
@@ -7,6 +7,7 @@ from .augmentations import *
 from .core.composition import *
 from .core.serialization import *
 from .core.transforms_interface import *
+from .pytorch import *
 
 # Perform the version check after all other initializations
 if os.getenv("NO_ALBUMENTATIONS_UPDATE", "").lower() not in {"true", "1"}:


ternaus:
@LightAllWorld 

Could you please be more specific.

1. Why do you think that pytorch dependency is missing?
2. What transform automation process you have in mind?
3. What proper functionality you have in mind, that is not in the library yet?

Your feedback would help to pick the proper priorities for the next step in the library.

LightAllWorld:
When training an AI model, I may wish to utilize the various transformation functions provided by Albumentations. By importing everything under Albumentations with import albumentations as A, you can conveniently use transformations such as A.Resize directly. However, when it comes to converting the image into a tensor for model input, A.ToTensorV2 is not immediately accessible in the same way.

ternaus:
Can you just do

```python
import torch
import albumentations
```
in your code?

LightAllWorld:
yes.
This is a piece of my code.
```python
import torch
import torchvision
import albumentations as A
from albumentations.pytorch import ToTensorV2
...
```

ternaus:
Ok, so the issue is not that torch should be imported in albumentations, bu that import in the way that you show does not work.

```python
 from albumentations.pytorch import ToTensorV2
```
will do the job

PR #1909, opened 2024-08-27
Remove unnecessary call to random.random() in deterministic cases

guillaume-rochette-oxb:
Since `p` is a probability, we should not need to "roll the dice" if p = 1.
This would be beneficial to enforce the reproducibility aspect.
A compelling example is:
- We have a **stochastic** pipeline of augmentations for the training samples.
- We have a **deterministic** pipeline of augmentations for the validation samples.
- Two training runs w/ identical training data would **diverge** w.r.t. their stochastic augmentations on the training data in the event where the number of validation samples would not be different.

## Summary by Sourcery

Optimize the transformation application logic by introducing a method to handle deterministic cases, avoiding unnecessary randomness when the probability is 1 or 0, and update related tests accordingly.

Enhancements:
- Introduce a new method `whether_to_apply` to determine if a transformation should be applied, optimizing for deterministic cases by avoiding unnecessary calls to `random.random()`.

Tests:
- Update test case in `test_pad_if_needed_position` to reflect changes in augmentation logic.

### DIFF
--- a/albumentations/core/transforms_interface.py
+++ b/albumentations/core/transforms_interface.py
@@ -98,7 +98,7 @@ class BasicTransform(Serializable, metaclass=CombinedMeta):
 
             return kwargs
 
-        if force_apply or (random.random() < self.p):
+        if force_apply or self.p >= 1.0 or (random.random() < self.p):
             params = self.get_params()
             params = self.update_params_shape(params=params, data=kwargs)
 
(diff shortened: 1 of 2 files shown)


ternaus:
Interesting use case.

We may assume that self.p is always in the `[0, 1]` range as it it forced by Pydantic in the constructor.

You also need to check that `pre-commit` hook does not raise any errors: https://github.com/albumentations-team/albumentations/blob/main/CONTRIBUTING.md#coding-guidelines

guillaume-rochette-oxb:
@sourcery-ai review

Issue #1941, opened 2024-09-20
Version 1.4.15 removes bounding boxes inconsistently and without permission


gui-miotto:
## Describe the bug

Version 1.4.15 removes bounding boxes with small area when `min_area=0.0` or even `-1.0`.

Moreover, it removes only the bounding box and not it's corresponding label. Which of course brings the data to an inconsistent state.

This behavior is not observed in version 1.4.14.

### To Reproduce

Just run the minimal working example below

```python
import numpy as np
import albumentations as A

in_data = dict(
    image=np.ones((500, 500, 3)),
    bboxes=np.array((0.1, 0.2, 1E-3, 1E-3)).reshape(1, 4),
    classes=np.array([1]),
)

bbox_conf = A.core.bbox_utils.BboxParams(format="yolo", label_fields=["classes"], min_area=-1.)
transf = A.Compose([A.HorizontalFlip(p=1.)], bbox_params=bbox_conf, is_check_shapes=False)

out_data = transf(**in_data)

print(in_data["classes"], in_data["bboxes"])
print(out_data["classes"], out_data["bboxes"])
```

### Expected behavior

Expected output is what is given by version 1.4.14:

```sh
[1] [[0.1   0.2   0.001 0.001]]
[1] [(0.8999999999999999, 0.2, 0.0010000000000000009, 0.0010000000000000009)]
```

### Actual behavior

Inconsistent behavior is given by version 1.4.15. Notice that the the bounding box coordinates is deleted, but the class is kept.

```sh
[1] [[0.1   0.2   0.001 0.001]]
[1] []
```

ternaus:
Thanks, will take a look ASAP.

ternaus:
Good catch. FIxed.

PR #1944, opened 2024-09-21
Fix small box filtering

ternaus:
Fixes: https://github.com/albumentations-team/albumentations/issues/1941

## Summary by Sourcery

Fix the small box filtering issue by adjusting the epsilon value in the bounding box filtering logic and add new test cases to ensure correct behavior across different bounding box formats.

Bug Fixes:
- Fix small box filtering by adjusting the epsilon value used in bounding box filtering logic to ensure that very small boxes are correctly filtered out.

Tests:
- Add new test cases to verify the correct filtering of very small bounding boxes across different formats, ensuring that boxes below a certain size are removed.

### DIFF
--- a/albumentations/core/bbox_utils.py
+++ b/albumentations/core/bbox_utils.py
@@ -475,10 +475,7 @@ def filter_bboxes(
     filtered_bboxes = clipped_bboxes[mask]
 
     # If no bboxes pass the filter, return an empty array with the same number of columns as input
-    if len(filtered_bboxes) == 0:
-        return np.array([], dtype=np.float32)
-
-    return filtered_bboxes
+    return filtered_bboxes if len(filtered_bboxes) > 0 else np.array([], dtype=np.float32)
 
 
 def union_of_bboxes(bboxes: np.ndarray, erosion_rate: float) -> np.ndarray | None:
--- a/albumentations/core/bbox_utils.py
+++ b/albumentations/core/bbox_utils.py
@@ -21,7 +21,6 @@ __all__ = [
 ]
 
 BBOX_WITH_LABEL_SHAPE = 5
-EPSILON = 1e-5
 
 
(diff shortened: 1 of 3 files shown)


albumentations/core/bbox_utils.py
```diff
@@ -475,10 +475,7 @@ def filter_bboxes(
     filtered_bboxes = clipped_bboxes[mask]
 
     # If no bboxes pass the filter, return an empty array with the same number of columns as input
-    if len(filtered_bboxes) == 0:
-        return np.array([], dtype=np.float32)
-
-    return filtered_bboxes
+    return filtered_bboxes if len(filtered_bboxes) > 0 else np.array([], dtype=np.float32)
```
gui-miotto, line 478: Hey @ternaus , thanks for fixing the issue.
Regarding this line of code, notice that it creates the issue of returning arrays with different number of dimensions depending on the number of bounding boxes that remains.

- If there is at least one box in `filtered_boxes` the function will return an array of shape (n, 4), i.e. dim=2.
- If there is no box in `filtered_boxes` the function will return an array of shape (0,), i.e. dim=1.

**Ideally, if there are no boxes left, we should return an array of shape (0, 4).** Which ensures data shape consistency.
So far, whenever I use albumentations, I have to call `bboxes.reshape(-1, 4)`  to work around this issue. Otherwise I can't , for instance, concatenated two arrays of bounding boxes.

Regarding this line of code, can't we just return `filtered_boxes` straight away, i.e. remove this if clause? (Disclaimer: I haven't checked the shape of `filtered_boxes` when the number of boxes is zero)

ternaus: @gui-miotto 

Thanks. Fixed in https://github.com/albumentations-team/albumentations/pull/1949

PR #1955, opened 2024-09-30
use the same interpolation for image and mask for Rotate

chen-xin-94:
Change to use the same interpolation for image and mask in Rotate augmentation. Originally Rotate always forces cv2.INTER_NEAREST for masks which causes unexpected artifacts, especially for heatmap masks

## Summary by Sourcery

Update the Rotate augmentation to use the same interpolation method for both images and masks, addressing issues with artifacts caused by the previous forced use of cv2.INTER_NEAREST for masks.

Bug Fixes:
- Ensure consistent interpolation for images and masks in the Rotate augmentation to prevent artifacts, especially in heatmap masks.

### DIFF
--- a/albumentations/augmentations/geometric/rotate.py
+++ b/albumentations/augmentations/geometric/rotate.py
@@ -151,7 +151,7 @@ class Rotate(DualTransform):
         y_max: int,
         **params: Any,
     ) -> np.ndarray:
-        img_out = fgeometric.rotate(mask, angle, cv2.INTER_NEAREST, self.border_mode, self.mask_value)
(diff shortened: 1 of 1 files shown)


ternaus:
Having NEAREST as interpolation for mask does not work for heatmap, but replacing it by the same interpolation as image will create artifacts for Semantic segmentation tasks.

I will implement a better solution, but right now, a workaround will be to pass both image and mask as images, like

```python
transform = A.Compose([A.Rotate()])
transformed = transform(images = [image, mask])
```

Issue #1874, opened 2024-08-12
Enhance augmentation objects with references to a random state.


Erotemic:
## Suggested Improvement

Looking at the current code, to draw random samples augmentation objects are using the global random state in the `random` module. This is ideal for maximally random pipelines that are impacted by any other use of the global random state outside of albumentations itself. 

This is not ideal for cases where the subcomponents of a system want their random generators to be seeded and not impacted by other components. For instance, right now there is no way for me to define a seeded augmentation pipeline that does not interfere with any other usage of the global random state.

I suggest adding a parameter to each augmentation class called: `seed`, `random_state`, or `rng` that defaults to None. When it is `None`, the it gets resolved to the global random state, which keeps the current behavior.

If it is an integer, then it would create a new `random.Random` object, and if `rng` is already a `random.Random` object, then it keeps it as-is, which allows augmentation pipelines to be independent of the global random state, but use an internally consistent random state.

## Potential Benefits

* Default behavior is unchanged
* Makes it easy to test augmentation pipelines without modifying the global state
* Makes it possible to set up a highly random, but consistent augmentation pipeline independent of any global random usage. 

## Additional Information

This is how the (now defuct) [imgaug](https://github.com/aleju/imgaug/blob/master/imgaug/augmenters/convolutional.py#L407) library handled randomness, where random states are explicitly passed and maintained.

I see there is a [random_utils](https://github.com/albumentations-team/albumentations/blob/7eda70e01e7f3bb31c1085e4fa473089b5a468be/albumentations/random_utils.py) module which somewhat handles this, but only for numpy random states, but as documented in CONTRIBUTING, it is only to ensure that any numpy.random usage is impacting the global Python random state.

I've written a function that I widely use called [ensure_rng](https://kwarray.readthedocs.io/en/release/kwarray.util_random.html#kwarray.util_random.ensure_rng) that handles the resolution of an argument to a valid random state object. In fact, it can also convert between the stdlib random.Random and np.random.RandomState objects. This might be useful here, although it doesn't exactly handle what is done in `random_utils.get_random_state`, but it is compatible with it.

I also see that [ReplayCompose](https://albumentations.ai/docs/examples/replay/) is a good solution to the problem of creating reproducible pipelines, but I believe maintaining a random state in each augmentation instance is complementary, especially in the realm of testing.

ternaus:
Thanks. Makes sense. Let me think about it.

ternaus:
@Erotemic 

[...]

PR #2031, opened 2024-10-25
Clean random generator

ternaus:
Fixes: https://github.com/albumentations-team/albumentations/issues/1874

## Summary by Sourcery

Refactor the random generator usage across the codebase to use numpy's default_rng, ensuring consistent and reproducible results. Enhance test coverage by adding new test cases for various augmentation functions, and fix issues related to random generator usage in tests.

Bug Fixes:
- Fix random generator usage to ensure consistent and reproducible results across various functions and tests.

Enhancements:
- Refactor random generator usage across the codebase to use numpy's default_rng for better performance and reproducibility.
- Improve test coverage by adding new test cases for various augmentation functions, ensuring they handle different scenarios and edge cases.

Tests:
- Add extensive test cases for cutout, filter_bboxes_by_holes, and other augmentation functions to validate their behavior with different parameters and inputs.
- Ensure all tests use numpy's default_rng for random number generation to maintain consistency in test results.


albumentations/core/transforms_interface.py
```diff
@@ -85,6 +85,34 @@ def __init__(self, p: float = 0.5, always_apply: bool | None = None):
...
+        self.py_random = random.Random(self.seed)  # Create instance instead of using global
+
+    def set_random_state(self, seed: int | None) -> None:
+        """Set random state for this transform and all nested transforms.
+
+        Args:
+            seed: Random seed to use
+        """
+        if seed is not None:
+            random.seed(seed)  # Set standard library random seed
```
Erotemic, line 99: It looks like this implementation impacting the global random state? Is that necessary? I would have expected that calling the `set_random_state` on an object would only impact the object's internal state, and not the global state.

albumentations/core/composition.py
```diff
@@ -120,6 +124,26 @@ def __init__(self, transforms: TransformsSeqType, p: float, mask_interpolation:
...
+        self.seed = seed
+        self.random_generator = np.random.default_rng(seed)
+
+        if seed is not None:
+            random.seed(seed)  # Set standard library random seed
+
+        # Propagate to all transforms
+        for transform in self.transforms:
+            if isinstance(transform, (BasicTransform, BaseCompose)):
+                transform.set_random_state(seed)
```
Erotemic, line 146: It looks like this is passing around the seed rather than the random generator. I think that might lead to undesirable behavior. For instance, say you compose two random affine transforms (a and b), and then you seed them with `seed=12345`. Each This would create two independent random states:

```
a = np.random.default_rng(12345)
b = np.random.default_rng(12345)
```

And then if you query them:

```
print(a.random())
print(b.random())
```

You will find they both print `0.22733602246716966`. Thus in the hypothetical case, the same affine transforms would be applied twice instead of having two independent affine transforms applied.

To fix this, I think it is sufficient to pass the `self.random_generator` instead of `seed`, e.g.:

```
    transform.set_random_state(self.random_generator)
```

This POC shows that it will cause the objects in the compose pipeline to share the same random state, which means the numbers they generate will be independent:

```
a = np.random.default_rng(12345)
b = np.random.default_rng(a)
print(a.random())
print(b.random())
```

Which produces:

```
0.22733602246716966
0.31675833970975287
```

ternaus: Good point, thanks. Will fix.

albumentations/core/transforms_interface.py
ternaus: Cleaned this up in a later Pull request

albumentations/core/composition.py
ternaus: Fixed in https://github.com/albumentations-team/albumentations/pull/2039

PR #2194, opened 2024-12-13
docs: fix inconsistency in YOLO normalization

mennohofste:
Earlier in the `normalization` arg for `Normalize` the docs state "The default mean and std are based on ImageNet. You can use mean and std values of (0.5, 0.5, 0.5) for inception normalization. And mean values of (0, 0, 0) and std values of (1, 1, 1) for YOLO."

This seems inconsistent with the note section. Furthermore, filling in a std value of 0 in the standard formula would lead to inifinite values, so I think the earlier mention description is correct.

## Summary by Sourcery

Documentation:
- Correct the documentation for YOLO normalization to use mean values of (0, 0, 0) and std values of (1, 1, 1) instead of the incorrect values previously stated.


ternaus:
@mennohofste Thanks, merged.

PR #2195, opened 2024-12-13
Fixed CenterCropPad incorrectly using fill value for the mask

iRyoka:
## Summary by Sourcery

Fix the issue with CenterCropPad incorrectly using the fill value for the mask and add a corresponding test to ensure correct behavior.

Bug Fixes:
- Fix CenterCropPad to correctly use the fill value for the mask during padding operations.

Tests:
- Add a test to verify that BaseCropAndPad uses the correct fill values for constant borders in images and masks.

### DIFF
--- a/albumentations/augmentations/crops/transforms.py
+++ b/albumentations/augmentations/crops/transforms.py
@@ -175,6 +175,26 @@ class BaseCropAndPad(BaseCrop):
             )
         return super().apply(img, crop_coords, **params)
 
+    def apply_to_mask(
+        self,
+        mask: np.ndarray,
+        crop_coords: Any,
+        **params: Any,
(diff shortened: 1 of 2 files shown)


iRyoka:
A minimal example to reprodice the addressed problem it:
```python
import albumentations as A
import numpy as np

c = A.CenterCrop(
    4, 
    3,
    pad_if_needed = True,
    fill = 1,
    fill_mask = 2
)

out = c(image = np.zeros((1,1,3)), mask = np.zeros((1,1)))

assert 2 in out['mask'], f"Wrong padding: {np.unique(out['mask'])}"
```

ternaus:
@iRyoka 

Thanks, good catch!

PR #2207, opened 2024-12-18
Adding A.AtLeastOneBBoxRandomCrop

guillaume-rochette-oxb:
Hi,

I'd like to add the transform `AtLeastOneBBoxRandomCrop` to the list of available transforms.

What does it do?
It crops an image at a given resolution, similarly to `RandomCrop`. However, unlike `BBoxSafeRandomCrop`, which ensures that **all** the bounding boxes are retained, `AtLeastOneBBoxRandomCrop` will ensure that **at least one** bounding box is retained.

Why?
1. Unlike `RandomCrop`, it ensures to keep at least one bounding box, so models, therefore models, such as [FasterRCNN](https://pytorch.org/vision/stable/models/faster_rcnn.html) or [MaskRCNN](https://pytorch.org/vision/stable/models/mask_rcnn.html), which require at least one box to compute the losses. This allow avoiding that "annoying moment", where the training would crash in the event where all bounding boxes would be dropped.
2. `BBoxSafeRandomCrop` requires an extra resizing transform in order to align every image to a common resolution, which is beneficial for accelerating training using options like [torch.backends.cudnn.benchmark](https://pytorch.org/docs/stable/backends.html#torch.backends.cudnn.benchmark).
3. A varying amount of boxes per sample is beneficial as it is a form of implicit regularization, in the sense that it affect the number of predicted region of interests, detections, etc.

Feel free to ask further questions about the motivations and/or modifications to in the PR.
Best regards,

## Summary by Sourcery

Add the AtLeastOneBBoxRandomCrop transform to ensure at least one bounding box is retained during cropping, enhancing model training stability and efficiency.

New Features:
- Introduce the AtLeastOneBBoxRandomCrop transform, which crops an image to a fixed resolution while ensuring that at least one bounding box is retained.

Tests:
- Add tests for the AtLeastOneBBoxRandomCrop transform to ensure its functionality and integration with existing augmentation workflows.


ternaus:
Looks good.

One small thing. When new transform is added, one need to run:

```bash
python -m tools.make_transforms_docs make
```

it will output the new table with transforms and targets. You will need to paste it to the readme, instead of the original one.

guillaume-rochette-oxb:
Hi,
Thanks :)
Sorry about the `README.md`, I did not know about that part, I've only read the `CONTRIBUTING.md`.

ternaus:
@guillaume-rochette-oxb Thank you for the PR.

I will update `Contributing.md`. Info about updating readme should be there.

PR #2211, opened 2024-12-20
Install `eval-type-backport` only on 3.10 and older

PerchunPak:
I assume you added this for pydantic, but that library is [basically useless on newer versions](https://github.com/alexmojaki/eval_type_backport/issues/24#issuecomment-2555757695). Also, `eval-type-backport` is broken on 3.12+ currently (see https://github.com/alexmojaki/eval_type_backport/issues/24)

## Summary by Sourcery

Bug Fixes:
- Fix incompatibility issues with `eval-type-backport` on Python 3.12 and later.


ternaus:
@PerchunPak Thanks for the Pull Request

PR #2226, opened 2024-12-31
Added ConstrainedCoarseDropout class

vedantdalimkar:
Added a new transform - Constrained coarse droput, which is a modification of the coarse dropout transform. 
In constrained coarse dropout, rectangular regions are only dropped out from the foreground objects belonging to the class specified by the user.

## Summary by Sourcery

Introduce the ConstrainedCoarseDropout transform. This transform drops out rectangular regions within user-specified foreground object classes, enhancing the CoarseDropout functionality.

New Features:
- Added a new transform called `ConstrainedCoarseDropout`. This transform allows for dropping out rectangular regions exclusively from foreground objects belonging to specific user-defined classes.

Tests:
- Added unit tests for the `ConstrainedCoarseDropout` transform.


vedantdalimkar:
@ternaus I have added some code for the test case which was failing. Are there any more issues to address with respect to the code?

PR #2288, opened 2025-01-22
fix distortion issue

RhysEvan:
I spent some more time looking into it, whilst the issue was already fixed I was determined to see for another option.
I cam to this result!

It still bugs as soon as the distortion value exceeds -0.6, in the positive direction it seems to be able to continue forever.
f.e.: distort_limit=(10, 10)
image
distort_limit=(-0.6, -0.6)
image

I must note that both these visual distortions for the checkerboard image actually do not make any sense... the image infinitely repeats itself in the first example (while the checkpoints don't). The resulting image should be more like the checkpoints in which there is simply black space in the image window and a "smaller" of your image with heavy distortions in the middle. This error is again of course caused by opencv but should be looked at for future improvements. Especially when aiming to ease the pipeline for synthetic fish eye distortion image creation.

Vise versa when you have a negative value suddenly the squares of the checkerboard image become so distorted that at times they become circles. These kinds of distortions do not comply with realistic heavy lens distortions. 

A funny side note is that the original code actually had a mistake in it, which caused all the corners to be shifted in comparison to their actual corner on the checkerboard, so even if we made it perfect it would have always looked of with the example script, see image below for clarification: zoomed in on the distortion plot with value -0.6, you can see in both frames that they are off

image

I hope that I adhered to all the requirements for being allowed to commit

## Summary by Sourcery

Bug Fixes:
- Fixed distortion issues that occurred when the distortion value exceeded -0.6.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
         slug: albumentations-team/albumentations
         files: ./coverage.xml
         flags: ubuntu-py3.12
-        fail_ci_if_error: true  # Optional: fails CI if upload fails
+        fail_ci_if_error: false  # Optional: fails CI if upload fails
         verbose: true  # Optional: helps with debugging
 
   check_code_formatting_types:
--- a/albumentations/augmentations/geometric/functional.py
+++ b/albumentations/augmentations/geometric/functional.py
@@ -15,6 +15,7 @@ from albucore import (
     preserve_channel_dim,
     vflip,
 )
+from scipy.interpolate import griddata
 from typing_extensions import NotRequired, TypedDict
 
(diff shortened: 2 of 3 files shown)


ternaus:
Generating inverse mapping this way is a more elegant solution.

Tests are failing. You may just not run that failing test on distortions.

RhysEvan:
When I ran pytest system after commenting out 2 lines, the tests gave no failure errors.
Did that fix everything or do I have to edit the error issue somewhere else specifically?

RhysEvan:
I created a requirement for the system not to seize when the Ubuntu CI upload fails by changing fail_ci_if_error to false.
I hope this makes it work; ubuntu tests are difficult to pre-emptively check as I only have Windows OS.

PR #2330, opened 2025-02-04
Fix rotate.py:Rotate.border_mode docstring to tell the correct default.

MalteEbner:
## Description

In https://github.com/albumentations-team/albumentations/pull/2245, the border_mode was changed to cv2.BORDER_CONSTANT. However, the docstring was not updated. This PR fixes it by updating the docstring as well.

## Tests

Not tested at all, as it only updates the docstring

## Summary by Sourcery

Documentation:
- Clarify the default border mode used in the `Rotate` transform.


ternaus:
Thanks. Trying my best to generate docstrings based on the code, but it the sitiation is far from perfect.

Thanks again.

Issue #2295, opened 2025-01-24
[Speed Up] GaussianNoise


ternaus:
Benchmark shows that `imgaug` has faster `GaussianNoise` implementation => need to learn from it and fix.

ternaus:
Fixed.

zetyquickly:
what about all other places that uses `random_generator.normal`? 

1. [generate_snow_textures](https://github.com/albumentations-team/albumentations/blob/1a4f551859610442c80e734df5179a3a8b6aabb8/albumentations/augmentations/functional.py#L620-L621)
2. [iso_noise](https://github.com/albumentations-team/albumentations/blob/1a4f551859610442c80e734df5179a3a8b6aabb8/albumentations/augmentations/functional.py#L1183)

ternaus:
@zetyquickly In theory in both of them should speed up as well.

PR #2336, opened 2025-02-09
Optimised the guassian sampling function.

vedantdalimkar:
Tries to address #2295 

The "sample_gaussian" was using a numpy function to sample gaussian noise. Changed it to the equivalent cv2 operation.

Also noticed that the "mean_range" and "std_range" parameters for GaussNoise is always a tuple of 2 equal values. "sample_gaussian" function randomly samples the mean and standard deviation from these 2 tuples. This is redudant and as sampling from an interval having one point always returns that point itself. This is also relatively expensive as compared to simply selecting that one point, as sampling uses a numpy generator object.

## Summary by Sourcery

Replace the NumPy-based Gaussian noise sampling with an equivalent OpenCV (cv2) implementation. Simplify the mean and standard deviation selection process by directly using the provided values when the ranges consist of identical elements, improving efficiency.

New Features:
- Add a seed parameter to control the random number generation in cv2 for Gaussian noise sampling.

Enhancements:
- Improve performance of Gaussian noise sampling by simplifying mean and standard deviation selection.

### DIFF
--- a/albumentations/augmentations/functional.py
+++ b/albumentations/augmentations/functional.py
@@ -2330,7 +2330,7 @@ def sample_gaussian(
         if params["std_range"][0] == params["std_range"][1]
         else random_generator.uniform(*params["std_range"])
     )
-    num_channels = size[2] if len(size) > 2 else 1
+    num_channels = size[2] if len(size) > MONO_CHANNEL_DIMENSIONS else 1
     mean_vector = mean * np.ones(shape=(num_channels,), dtype=np.float64)
     std_dev_vector = std * np.ones(shape=(num_channels,), dtype=np.float64)
     gaussian_sampled_arr = np.zeros(shape=size)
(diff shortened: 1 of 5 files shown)


albumentations/augmentations/functional.py
```diff
@@ -2158,6 +2158,7 @@ def generate_noise(
     max_value: float,
     approximation: float,
     random_generator: np.random.Generator,
+    seed: int | None = None,
```
ternaus, line 2161: We should not pass transforms seed to the function as it is fixed for all the life of the transform => we will generate exactly the same random noise

ternaus (changes requested):

albumentations/augmentations/functional.py
vedantdalimkar: I have made the changes as per your suggestion. Please find attached the results before and after the optimsation.
opt.txt contains results after optimisation, source.txt contains benchmark results.
[opt.txt](https://github.com/user-attachments/files/18737569/opt.txt)
[source.txt](https://github.com/user-attachments/files/18737570/source.txt)

ternaus (approved):

Issue #2293, opened 2025-01-24
[Speed up] GaussianBlur


ternaus:
Benchmark shows that `imgaug` has faster GaussianBlur implementation => need to learn from it and fix.

zetyquickly:
This one probably comes down to a comparison between [`cv2.filter2D`](https://github.com/albumentations-team/albumentations/blob/1a4f551859610442c80e734df5179a3a8b6aabb8/albumentations/augmentations/functional.py#L471) and [`cv2.GaussianBlur`](https://github.com/aleju/imgaug/blob/0101108d4fed06bc5056c4a03e2bcb0216dac326/imgaug/augmenters/blur.py#L279)

ternaus:
Could be. I use cv2.filder2D to apply normalized kernel to match Pillow behavbior.

vedantdalimkar:
@ternaus Just FYI, using cv2.GuassianBlur instead of seperable convolutions was giving similar results (almost equal) in terms of speed. I favored seperable convolutions over cv2.GuassianBlur because it was consistent with the design decision of creating kernels in a similar fashion to PIL. If you want any changes, please let me know.

PR #2347, opened 2025-02-18
Optimised gaussian blur transform

vedantdalimkar:
Addresses #2293 

Optimised gaussian blur at the algorithm level. Made use of the fact that gaussian kernels are separable, which means that convolution with 2D gaussian kernel can be simplified to 2 convolutions with 1D gaussian kernel which reduces the time complexity of the algorithm. 

1. Modified the 2D gaussian kernel creation function to a function that creates 1D gaussian kernel.
2. Replaced 2D convolution with separable 2D convolution
3. Added tests for the new 1D gaussian kernel creation function

Time benchmarks - 
 Old algorithm results -
 [source_gaussian_blur.txt](https://github.com/user-attachments/files/18847961/source_gaussian_blur.txt)
 
 Updated algorithm results -
[gaussian_blur.txt](https://github.com/user-attachments/files/18847965/gaussian_blur.txt)

### DIFF
--- a/albumentations/augmentations/blur/functional.py
+++ b/albumentations/augmentations/blur/functional.py
@@ -285,3 +285,48 @@ def create_gaussian_kernel(sigma: float, ksize: int = 0) -> np.ndarray:
 
     # Create 2D kernel
     return kernel_1d[:, np.newaxis] @ kernel_1d[np.newaxis, :]
+
+
+def create_gaussian_kernel_1d(sigma: float, ksize: int = 0) -> np.ndarray:
+    """Create a 1D Gaussian kernel following PIL's approach.
+
+    Args:
+        sigma: Standard deviation for Gaussian kernel.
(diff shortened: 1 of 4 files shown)


ternaus:
@vedantdalimkar Great job. Really nice optimizations. I did not even know that cv2 has separable convolutions.

vedantdalimkar:
@ternaus Just FYI, using cv2.GuassianBlur instead of seperable convolutions was giving similar results (almost equal) in terms of speed. I favored seperable convolutions over cv2.GuassianBlur because it was consistent with the design decision of creating kernels in a similar fashion to PIL. If you want any changes, please let me know.

ternaus:
I checked. GaussianBlur and Separable give similar performance, with Gaussian being a bit faster.

But here we want normalized kernel to match PIL, which prevents from using Gaussian Directly and using separable here was a material 2.5 speedup.

Thanks.

PR #2378, opened 2025-03-03
Fix ElasticTransform blur kernel and add missing args 

liamjwang:
Hello, a few suggested changes to ElasticTransform.

I think the `approximate` argument was reversed by mistake.

Also, currently the blur step is skipped if the kernel size is (0, 0). 
However, I think it was intended to let `cv2.GaussianBlur` automatically compute the kernel size.

With this change, the behavior matches the ElasticTransform output from older versions (e.g. 1.4.15).

Also added `border_mode`, `fill`, and `fill_mask` args.

## Summary by Sourcery

Fixes issues in the ElasticTransform augmentation, including a reversed `approximate` argument, skipping the blur step when the kernel size is (0, 0), and missing `border_mode`, `fill`, and `fill_mask` arguments.

Bug Fixes:
- Fixes a bug where the blur step was skipped in ElasticTransform when the kernel size was (0, 0).
- Fixes a bug where the `approximate` argument was reversed in ElasticTransform.

Enhancements:
- Adds `border_mode`, `fill`, and `fill_mask` arguments to ElasticTransform to provide more control over the transformation.

### DIFF
--- a/albumentations/augmentations/geometric/functional.py
+++ b/albumentations/augmentations/geometric/functional.py
@@ -1599,7 +1599,7 @@ def generate_displacement_fields(
         ).astype(np.float32)
 
     # # Apply Gaussian blur if needed using fast OpenCV operations
-     # When kernel_size is (0,0) cv2.GaussianBlur uses automatic kernel size. Kernel == (0,0) is NOT a noop.
+    # When kernel_size is (0,0) cv2.GaussianBlur uses automatic kernel size. Kernel == (0,0) is NOT a noop.
     # Reshape to 2D array (combining first dimension with height)
     shape = fields.shape
     fields = fields.reshape(-1, shape[-1])
--- a/albumentations/augmentations/geometric/functional.py
+++ b/albumentations/augmentations/geometric/functional.py
@@ -1599,22 +1599,22 @@ def generate_displacement_fields(
         ).astype(np.float32)
 
     # # Apply Gaussian blur if needed using fast OpenCV operations
-    if kernel_size != (0, 0):
-        # Reshape to 2D array (combining first dimension with height)
-        shape = fields.shape
-        fields = fields.reshape(-1, shape[-1])
+     # When kernel_size is (0,0) cv2.GaussianBlur uses automatic kernel size. Kernel == (0,0) is NOT a noop.
+    # Reshape to 2D array (combining first dimension with height)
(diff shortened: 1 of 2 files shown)


ternaus:
`border_mode`
`fill`
`fill_mask` 

Do not really have material effect in Elastic Transform

albumentations/augmentations/geometric/functional.py
```diff
@@ -1599,22 +1599,22 @@ def generate_displacement_fields(
...
+        kernel_size,
+        sigma,
+        dst=fields,
+        borderType=cv2.BORDER_REPLICATE,
+    )
 
-        # Restore original shape
-        fields = fields.reshape(shape)
+    # Restore original shape
+    fields = fields.reshape(shape)
```
ternaus, line 1617: I agree. Good catch.

albumentations/augmentations/geometric/transforms.py
```diff
@@ -143,11 +152,23 @@ def __init__(
...
+        fill: tuple[float, ...] | float = 0,
+        fill_mask: tuple[float, ...] | float = 0,
     ):
         super().__init__(p=p)
         self.interpolation = interpolation
         self.mask_interpolation = mask_interpolation
         self.keypoint_remapping_method = keypoint_remapping_method
+        self.border_mode = border_mode
+        self.fill = fill
+        self.fill_mask = fill_mask
```
ternaus, line 171: border mode, fill and fill mask do not have material effects in this transform, let's not add them.

ternaus (changes requested):

albumentations/augmentations/geometric/transforms.py
liamjwang: My use-case involves extreme deformations (alpha=7000, sigma=60).
In this case, significant portions around the edges of the image sample outside of the original image after stretching.
I have used the border_mode argument for choosing whether these parts are filled with a solid color or reflected, (which is a very meaningful visual difference), and I have found this to be an effective augmentation.
Without these arguments, I would not be able to use this transform.

ternaus: I see. Did not think about extreme deformations. Sounds good.

ternaus (approved):

liamjwang:
Thank you!

PR #2380, opened 2025-03-04
add codeflash GHA ⚡️

KRRT7:
## Summary by Sourcery

Adds a GitHub Actions workflow to run Codeflash on pull requests that modify files in the albumentations directory.
CI:
- Adds a new CI workflow to run Codeflash on pull requests.
- Configures the workflow to only run when code within the albumentations module is modified.
- Sets up Python and installs dependencies, including development dependencies.
- Runs Codeflash to optimize new Python code in the pull request.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -12,5 +12,3 @@ torch>=2.3.1
 torchvision>=0.18.1
 types-PyYAML
 types-setuptools
-codeflash
-ruff
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -12,3 +12,5 @@ torch>=2.3.1
 torchvision>=0.18.1
 types-PyYAML
 types-setuptools
+codeflash
+ruff
--- a/.github/workflows/codeflash.yml
+++ b/.github/workflows/codeflash.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@v5
       
-      - name: Install Dependecies
+      - name: Install Dependencies
         run: uv sync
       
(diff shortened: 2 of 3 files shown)


KRRT7:
Hi @plashchynski the GHA is good to go, you will have to create an API key @ https://app.codeflash.ai/app/apikeys and then set the key as an action secret.

requirements-dev.txt
```diff
@@ -12,3 +12,5 @@ torch>=2.3.1
 torchvision>=0.18.1
 types-PyYAML
 types-setuptools
+codeflash
+ruff
```
ternaus, line 16: It should not be in requirements-dev.txt

These files only needed in CI => should be installed in CI, not during local development.

ternaus (changes requested):

ternaus:
@KRRT7 Will this GHA propose changes or will modify files?

requirements-dev.txt
KRRT7: I'll modify to that end.

KRRT7:
> @KRRT7 Will this GHA propose changes or will modify files?

it will propose changes into the PRs themselves if any optimizations are found

requirements-dev.txt
misrasaurabh1: one more consideration is using codeflash locally. You can use codeflash to optimize any existing code through the CLI or trace and optimize any e2e workflow. It might be a good idea to have codeflash as a dev dependency if you want to use it locally, otherwise ci time installation makes sense.

ternaus (approved):

PR #2386, opened 2025-03-05
Deleted TemplateTransform

ternaus:
## Summary by Sourcery

Chores:
- Remove TemplateTransform class.


salvaba94:
Hi @ternaus , what is the substitute for TemplateTransform? I'm trying to use some of the MixUp kind but I've see both MixUp and TemplateTransform have been removed.

ternaus:
@salvaba94 Interesting. I did not think that anyone is using them.

Nothing existing comes to my mind, but you can implement a custom one with this functionality. A I understand you only need to add pixel values of target image and and additional one => straightforward to implement:

https://albumentations.ai/docs/4-advanced-guides/creating-custom-transforms/

and 

Domain adaptation transforms could be used as exampled on how to pass additional data to the pipeline: https://github.com/albumentations-team/AlbumentationsX/blob/main/albumentations/augmentations/mixing/domain_adaptation.py

PR #2471, opened 2025-03-13
Support empty label fields

sklum:
For my use case, it's valuable to be able to train on images without any labels of any kind as a part of a batch.  This change allows that.

## Summary by Sourcery

Enhancements:
- Handle empty label fields during data processing.

### DIFF
--- a/albumentations/core/utils.py
+++ b/albumentations/core/utils.py
@@ -209,7 +209,7 @@ class DataProcessor(ABC):
             for label_field in self.params.label_fields:
                 self._validate_label_field_length(data, data_name, label_field)
                 encoded_labels = self.label_manager.process_field(data_name, label_field, data[label_field])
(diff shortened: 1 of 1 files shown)


ternaus:
@sklum 
> For my use case, it's valuable to be able to train on images without any labels of any kind as a part of a batch. This change allows that.

What do you mean?

You can pass images, and no any other labels to nearly every transform.

sklum:
If e.g. the bounding boxes are an empty tensor (i.e. `bboxes = tensor([])`), the call to `hstack` will fail with the following error:

`ValueError: all the input arrays must have same number of dimensions, but the array at index 0 has 1 dimension(s) and the array at index 1 has 2 dimension(s)`

ternaus:
```python
train_transforms(image=image, bboxes=[], labels=[], ids=[])
```
works

PR #2534, opened 2025-05-25
Adding apply_to_images to Pad

### LINKED ISSUE #2463: [Feature request] Add apply_to_images to Pad

martinsbruveris:
In addition to adding `apply_to_images` function to `Pad`, this PR is also making a modification to unit tests: Currently unit tests are calling transforms with vector-valued `fill` parameters (possibly others too) even for grey-scale images, e.g., we are calling

```
img = np.zeros((128, 128))
transform = Pad(fill=(10, 10, 10))
result = transform(image=img)
```

If this is desired behaviour, I am happy to adapt my implementation of `Pad.apply_to_images` to use the first element of `fill` in this case. Right now I have modified the unit tests to only use scalar `fill` parameters for grey-scale images.

## Summary by Sourcery

Enable batch-oriented padding by adding a dedicated functional API and corresponding transform method, and update tests to enforce scalar fill values for grayscale images.

New Features:
- Introduce pad_images_with_params for batch padding support
- Add apply_to_images method to Pad transform to handle image batches

Tests:
- Normalize non-scalar fill parameters to scalars for grayscale images in unit tests

Resolves #2463

### DIFF
--- a/albumentations/augmentations/geometric/functional.py
+++ b/albumentations/augmentations/geometric/functional.py
@@ -1585,7 +1585,7 @@ def pad_images_with_params(
         kwargs = {"constant_values": constant_values}
     else:
         kwargs = {}
-    
+
     images = np.pad(images, pad_width=pad_width, mode=mode, **kwargs)
     if no_channel_dim:
         images = images[..., 0]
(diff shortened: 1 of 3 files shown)


ternaus:
@martinsbruveris your PR made `Pad` and `PadIfNeeded` transforms, being applied to videos 3.5 times faster. (+250%) 
gitThanks.

PR #2535, opened 2025-05-26
Vectorized BaseDropout

huydoanx172:
Vectorized BaseDropout for all the current dropout options except for inpainting, since cv2 inpainting doesn't support 3d images. Cv2 inpainting was implemented via a loop, had the same runtime as before. Achieved 3x-10x speed increase for other fill modes depending on size of batches.

## Summary by Sourcery

Enable volumetric (3D) and batch processing for BaseDropout by adding vectorized hole-filling functions for constant, random, and uniform fills, along with corresponding transform methods; retain per-slice looping for inpainting due to OpenCV limitations.

New Features:
- Add support for 3D volume and batch dropout with new cutout_on_volume and cutout_on_volumes functions

Enhancements:
- Vectorize constant and random fill modes for volumes and batches to improve performance

### DIFF
--- a/albumentations/augmentations/dropout/functional.py
+++ b/albumentations/augmentations/dropout/functional.py
@@ -157,6 +157,34 @@ def fill_holes_with_value(img: np.ndarray, holes: np.ndarray, fill_value: np.nda
     return img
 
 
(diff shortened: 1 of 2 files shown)


ternaus:
@huydoanx172 Thanks, merged.

